PR #14, opened 2021-04-04
refactor: make flint a submodule of a flint package

oscarbenjamin:
See https://github.com/fredrik-johansson/python-flint/issues/1#issuecomment-812901808

This moves the flint extension module to `flint._flint` inside a package called `flint`. This is needed to make it possible to build relocatable wheels. There needs to be a package directory to place the wheels inside.


fredrik-johansson:
Looks good. Thanks!

PR #19, opened 2021-09-18
Fix Doctest for arb220

jan-janssen:


oscarbenjamin:
Thanks for this. I just applied these same changes in #31 so this isn't needed any more.

PR #22, opened 2022-05-18
maint: update versions for built wheels (flint 2.9.0, arb 2.23.0)

oscarbenjamin:
https://groups.google.com/g/sage-devel/c/t6WnShEqoR4/m/-sUUQOb-AAAJ


oscarbenjamin:
The wheels build fine but can't be imported after install:
```python
    File "<string>", line 1, in <module>
    File "/tmp/tmp.jJDSzgvxEi/venv/lib/python3.7/site-packages/flint/__init__.py", line 1, in <module>
      from ._flint import *
  ImportError: /tmp/tmp.jJDSzgvxEi/venv/lib/python3.7/site-packages/flint/../python_flint.libs/libarb-43ad4021.so.2.13.0: undefined symbol: n_gcd_full
```
Looks like this was recently fixed in arb: https://github.com/fredrik-johansson/arb/pull/419

oscarbenjamin:
> Two changes that will be needed:
> 
> * Change layout of `fmpq_poly_struct` (`den` now comes last) in `_flind.pxd`
> * `arith_divisor_sigma` changed the order of the last two arguments

I've changed these. They aren't tested at all though so I don't know if they're actually working.

I've also updated various versions of thing including flint-2.9.0, arb-2.23.0 and updated various actions to be able to build wheels for CPython 3.11.

Those changes keep everything up to date. I've tested that the wheels work locally after downloading on Linux and I'll check OSX later.

Missing features are still Windows and Apple Silicon. I'll work on those soon but I don't have easy access to either platform.

fredrik-johansson:
Excellent, thanks! You want to merge this now or continue working in this PR?

oscarbenjamin:
Let's merge this for now and I'll keep working on it.

oscarbenjamin:
On the topic of Windows I was just looking at the flint CI and I saw that there is a mingw-gcc job but I don't quite understand how it works:
https://github.com/flintlib/flint2/blob/abe952e2397ea7b42b41d2eb1b74795e7921ac2c/.github/workflows/CI.yml#L234-L270
The other jobs all seem to use the `./.build_dependencies` script but in the mingw job it seems that somehow GMP and MPFR are already there. Do you have any idea how that works?

fredrik-johansson:
I'm not sure, but GMP and MPFR are dependencies of GCC, so both are presumably preinstalled on the system.

fredrik-johansson:
Maybe the Arb CI using Appveyor is helpful? https://github.com/fredrik-johansson/arb/blob/master/appveyor.yml

oscarbenjamin:
I wonder if to some extent our requirements here are a bit different from the CI systems because we are packaging up files for distribution rather than just to run in the CI system itself. I think that means that there needs to be precise control of which versions of all dependencies are being shipped and how those dependencies are built which is I guess not as much of a concern in CI.

PR #27, opened 2022-12-06
maint: update macos image for building wheels

oscarbenjamin:
The macOS-10.15 image is deprecated. It looks like numpy uses macos-12 to build their wheels so I went with that:
https://github.com/numpy/numpy/blob/297c66131f2a32ebb7dfb0aeb9d88d917a791430/.github/workflows/wheels.yml#L78

I'll download and test the wheels before merging.

### DIFF
--- a/.github/workflows/buildwheel.yml
+++ b/.github/workflows/buildwheel.yml
@@ -9,8 +9,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
I tested the wheel for Python 3.10 on OSX 12.0.1 and it seems to work fine. This also removes the last warning from GitHub Actions CI.

PR #28, opened 2022-12-06
fix: use long long on Windows

### LINKED ISSUE #10: Windows support

oscarbenjamin:
Fixes #10.

There might be more places that needs to be fixed to use `slong` rather than `long` but this is an intentionally minimal change that achieves the basic goal of getting python-flint to work on Windows. With this and #26 I was able to make a working Windows wheel that locally passed all tests.

### DIFF
--- a/.github/workflows/buildwheel.yml
+++ b/.github/workflows/buildwheel.yml
@@ -6,6 +6,7 @@ jobs:
   build_wheels:
     name: Build wheels for ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    continue-on-error: true
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


oscarbenjamin:
Okay the wheels are building in 32 bit linux and passing basic tests but it doesn't seem that there is a straight-forward way to run a CI job in a 32 bit OS to do a proper test.

oscarbenjamin:
Okay, I'll merge this now so we can focus on making the Windows wheels work.

PR #33, opened 2022-12-13
test: run make check for Arb

oscarbenjamin:
See https://github.com/fredrik-johansson/python-flint/issues/32#issuecomment-1348778081


oscarbenjamin:
So the Arb tests pass in CI on all platforms but the python-flint doctests still fail on Windows. I can reproduce the same locally on Windows. I think it's good to have these Arb tests. It would probably also be good to test gmp, mpfr and flint but for now this will do.

PR #38, opened 2022-12-17
doctesting & bugfixing

fredrik-johansson:
Nothing big, just getting the hang of this again.


oscarbenjamin:
Looks good to me.

Maybe there should be a CI job for building the docs.

PR #39, opened 2022-12-17
Add tests for modules wrapping flint

oscarbenjamin:
This is a work in progress.

Maybe the test script should be reorganised a bit as the current one-test-function-per-module structure will get a bit unwieldy if more tests are added.

### DIFF
--- a/src/flint/fmpq_series.pyx
+++ b/src/flint/fmpq_series.pyx
@@ -19,6 +19,13 @@ cdef fmpq_series_coerce_operands(x, y):
             return acb_series(x), acb_series(y)
     return NotImplemented, NotImplemented
 
+cdef any_as_fmpq_series(obj):
+    if typecheck(obj, fmpq_series):
+        return obj
+    if typecheck(obj, fmpz_series):
+        return fmpq_series(obj)
+    return NotImplemented
+
 cdef class fmpq_series(flint_series):
 
     cdef fmpq_poly_t val
@@ -41,7 +48,7 @@ cdef class fmpq_series(flint_series):
         if val is not None:
             if typecheck(val, fmpq_series):
                 fmpq_poly_set(self.val, (<fmpq_series>val).val)
-                self.prec = min((<fmpz_series>val).prec, getcap())
+                self.prec = min((<fmpq_series>val).prec, getcap())
             elif typecheck(val, fmpz_series):
                 fmpq_poly_set_fmpz_poly(self.val, (<fmpz_series>val).val)
                 self.prec = min((<fmpz_series>val).prec, getcap())
@@ -61,6 +68,23 @@ cdef class fmpq_series(flint_series):
             if fmpz_is_zero((<fmpz>den).val):
                 raise ZeroDivisionError("cannot create fmpq_series with zero denominator")
             fmpq_poly_scalar_div_fmpz(self.val, self.val, (<fmpz>den).val)
+
+    def __richcmp__(s, t, int op):
+        cdef bint r
+        if op != 2 and op != 3:
+            raise TypeError("series cannot be ordered")
+        s = any_as_fmpq_series(s)
+        if t is NotImplemented:
+            return s
+        t = any_as_fmpq_series(t)
+        if t is NotImplemented:
+            return t
+        r = fmpq_poly_equal((<fmpq_series>s).val, (<fmpq_series>t).val)
+        if r:
+            r = (<fmpq_series>s).prec == (<fmpq_series>t).prec
+        if op == 3:
+            r = not r
+        return r
 
     def __len__(self):
         return fmpq_poly_length(self.val)
@@ -197,7 +221,7 @@ cdef class fmpq_series(flint_series):
             u = fmpq_series.__new__(fmpq_series)
 
             if fmpq_poly_is_zero((<fmpq_series>s).val):
-                u.cap = cap
+                (<fmpq_series>u).prec = cap
                 return u
 
             sval = (<fmpq_series>s).valuation()
@@ -236,6 +260,7 @@ cdef class fmpq_series(flint_series):
         return fmpq_series._div_(s, t)
 
     # generic exponentiation (fallback code)
+    # XXX: use fmpq_poly_pow_trunc instead?
     def __pow__(s, ulong exp, mod):
         if mod is not None:
             raise NotImplementedError("modular exponentiation")
@@ -308,7 +333,7 @@ cdef class fmpq_series(flint_series):
     cdef bint one_constant_term(s):
         if fmpq_poly_is_zero((<fmpq_series>s).val):
             return False
-        if fmpz_is_one(&((<fmpq_series>s).val.coeffs[0])):
+        if fmpz_equal(&((<fmpq_series>s).val.coeffs[0]), (<fmpq_series>s).val.den):
             return True
         return False
 
--- a/src/flint/fmpq_series.pyx
+++ b/src/flint/fmpq_series.pyx
@@ -19,13 +19,6 @@ cdef fmpq_series_coerce_operands(x, y):
             return acb_series(x), acb_series(y)
     return NotImplemented, NotImplemented
 
-cdef any_as_fmpq_series(obj):
-    if typecheck(obj, fmpq_series):
-        return obj
-    if typecheck(obj, fmpz_series):
-        return fmpq_series(obj)
-    return NotImplemented
-
 cdef class fmpq_series(flint_series):
 
     cdef fmpq_poly_t val
@@ -69,21 +62,13 @@ cdef class fmpq_series(flint_series):
                 raise ZeroDivisionError("cannot create fmpq_series with zero denominator")
             fmpq_poly_scalar_div_fmpz(self.val, self.val, (<fmpz>den).val)
 
-    def __richcmp__(s, t, int op):
+    def _equal_repr(s, t):
         cdef bint r
-        if op != 2 and op != 3:
-            raise TypeError("series cannot be ordered")
-        s = any_as_fmpq_series(s)
-        if t is NotImplemented:
-            return s
-        t = any_as_fmpq_series(t)
-        if t is NotImplemented:
-            return t
+        if not typecheck(t, fmpq_series):
+            return False
         r = fmpq_poly_equal((<fmpq_series>s).val, (<fmpq_series>t).val)
         if r:
             r = (<fmpq_series>s).prec == (<fmpq_series>t).prec
-        if op == 3:
-            r = not r
         return r
 
     def __len__(self):
--- a/test/test.py
+++ b/test/test.py
@@ -1029,6 +1029,7 @@ def test_nmod_mat():
     assert (a * flint.fmpz(3)).entries() == [G(x,17) for x in [3,6,9,12,15,18]]
     assert (flint.fmpz(3) * a).entries() == [G(x,17) for x in [3,6,9,12,15,18]]
     assert M(2,2,[1,1,2,2],17).rank() == 1
+    assert M(Z(2,2,[1,2,3,4]),17) == M(2,2,[1,2,3,4],17)
     A = M(5,3,Z.randbits(5,3,5).entries(),17)
     B = M(3,7,Z.randtest(3,7,5).entries(),17)
     C = M(7,2,Z.randtest(7,2,5).entries(),17)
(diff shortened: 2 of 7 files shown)


test/test.py
```diff
@@ -63,6 +83,97 @@ def test_fmpz():
...
+        (lambda n: flint.fmpz.fib_ui(n),
+            [OE, 0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55]),
+        (lambda n: flint.fmpz(n).rising(0),
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]),
+        (lambda n: flint.fmpz(n).rising(1),
+            [-1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),
+        (lambda n: flint.fmpz(n).rising(2),
+            [0, 0, 2, 6, 12, 20, 30, 42, 56, 72, 90, 110]),
+        (lambda n: flint.fmpz.bin_uiui(n, 0),
+            [OE, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]),
```
oscarbenjamin, line 135: These function signatures are generally a bit inconsistent e.g.:
```python
fmpz(n).func()
fmpz.func(n)
fmpz(n).func(m)
fmpz.func(n, m)
```
Is that something that should be kept for backwards compatibility?

Perhaps if a consistent scheme was decided then aliases could be added with more consistent signatures e.g. `fmpz.rising_factorial(n,m)`.

Does it really make sense to have `_ui` in the name for a Python method?

test/test.py
```diff
@@ -63,6 +83,97 @@ def test_fmpz():
...
+        (1296814508839693536173209832765271992846610925502473758289451540212712414540699659186801, 1)]
+    T, F, VE, OE = True, False, ValueError, OverflowError
+    cases = [
+        # (f, [f(-1), f(0), f(1), f(2), ... f(10)]),
+        (lambda n: flint.fmpz(n).is_prime(),
+            [0, 0, 0, 1, 1, 0, 1, 0, 1, 0, 0, 0]),
+        (lambda n: flint.fmpz(n).is_probable_prime(),
+            [0, 0, 0, 1, 1, 0, 1, 0, 1, 0, 0, 0]),
+        (lambda n: flint.fmpz(n).is_perfect_power(),
+            [T, T, T, F, F, T, F, F, F, T, T, F]),
```
oscarbenjamin, line 117: I guess `is_prime` and `is_probable_prime` should be changed to return True/False rather than 0/1.

fredrik-johansson: Yes.

test/test.py
fredrik-johansson: This has to be decided on a case by case basis I think. Sometimes a function f(x,y) has a clear primary variable and then x.f(y) makes sense; binomials and rising factorials arguably fall into this category. In other cases a static method makes more sense.

We could also add a namespace of standalone functions (not requiring method calls), but that is a separate issue.

Agree there is no reason to have ui versions in the Python interface.

test/test.py
```diff
@@ -19,6 +19,13 @@ def test_fmpz():
     assert flint.fmpz() == flint.fmpz(0)
     L = [0, 1, 2, 3, 2**31-1, 2**31, 2**63-1, 2**63, 2**64-1, 2**64]
     L += [-x for x in L]
+    for i in L:
+        f = flint.fmpz(i)
+        assert int(f) == i
+        assert flint.fmpz(f) == f
+        assert flint.fmpz(str(i)) == f
+    raises(lambda: flint.fmpz("qwe"), ValueError)
+    raises(lambda: flint.fmpz([]), TypeError)
```
oscarbenjamin, line 28: I just realised I'm using this `raises` function incorrectly because it works different from the `pytest/sympy` one. Is there any reason not to use pytest in general here in `python_flint`?

fredrik-johansson: No, I wouldn't mind.

oscarbenjamin:
I hadn't realised that the Linux builds in CI were still using MPIR rather than GMP. They now fail because they're trying to download from mpir.org which has now expired.

fredrik-johansson:
The easiest solution might be to wrap ``arb_load_str`` and ``arb_dump_str``.

oscarbenjamin:
> The easiest solution might be to wrap `arb_load_str` and `arb_dump_str`.

It would probably be useful to add those if that's a format already used by arb.

To me the most natural representation in Python would be to convert an arb to/from a tuple of 4 int/fmpz e.g. `(m1,e1,m2,e2)` for `m1*2**e1 +- m2*2**e2`. If I understand correctly that's basically the representation apart from a separate flag to indicate special values. I guess if the special exponent values are defined using `COEFF_MIN` then they aren't really portable because that varies based on `FLINT_BITS` but there could be another way to encode those like saying that when `m1 = 0`, `e1=0,1,2,...` denotes the special values (rather than `COEFF_MIN+0,1,2,...`).

fredrik-johansson:
The ``dump``/``load`` interface is designed to be portable and future-compatible.

The representation used by ``arb_t`` in memory does depend on ``FLINT_BITS``, but ``arb_dump_str`` converts the special exponent values to something like 0 = ``0 0 0 0``, ``+inf`` = ``0 -1 0 0``, ...

oscarbenjamin:
> `arb_dump_str` converts the special exponent values to something like 0 = `0 0 0 0`, `+inf` = `0 -1 0 0`, ...

Yes, that's what I was thinking. It would be good to have a way to convert to and from this format as a tuple of integers rather than a string representation of those integers. I see what you mean though that it might be simplest to just wrap the text dump functions for now.

src/flint/fmpq_series.pyx
```diff
@@ -62,6 +69,23 @@ cdef class fmpq_series(flint_series):
...
+            raise TypeError("series cannot be ordered")
+        s = any_as_fmpq_series(s)
+        if t is NotImplemented:
+            return s
+        t = any_as_fmpq_series(t)
+        if t is NotImplemented:
+            return t
+        r = fmpq_poly_equal((<fmpq_series>s).val, (<fmpq_series>t).val)
+        if r:
+            r = (<fmpq_series>s).prec == (<fmpq_series>t).prec
```
oscarbenjamin, line 84: Is this the kind of equality comparison we want for series?

- A series cannot compare equal to anything that is not a series e.g. `fmpq_series([1]) != 1`
- An `fmpz_series` can compare equal to an `fmpq_series`: `fmpq_series([1]) == fmpz_series([1])`
- Two series with different prec compare unequal: `fmpz_series([1],5) != fmpz_series([1],6)`
- Invalid series compare equal: `fmpz_series([],-1) == fmpz_series([],-1)`

oscarbenjamin:
Okay I think this PR is at a reasonable milestone for review/merge.

Tests are added for almost full line coverage of all modules relating to flint (rather than arb). I've also fixed a few bugs and added `__eq__` and `__ne__` to `fmpz_series` and `fmpq_series`.

Later I think it would be good to reorganise these tests to make more of the testing code generic so that e.g. all poly types are tested by the same test code for consistency. There are currently some inconsistencies about exactly which type has which methods and what exceptions they raise etc so that could be fixed at the same time.

I haven't looked at the arb modules at all yet but again that can be something to follow up. The arb-related modules are only really tested by the doctests and in some cases they give fairly low coverage.

The coverage with both tests and doctests here is:
```
$ coverage report --sort=cover
Name                        Stmts   Miss  Cover
-----------------------------------------------
src/flint/arb_series.pyx      563    492    13%
src/flint/arb_poly.pyx        195    170    13%
src/flint/arf.pyx             109     85    22%
src/flint/acb_poly.pyx        233    179    23%
src/flint/dirichlet.pyx        56     27    52%
src/flint/arb_mat.pyx         323    127    61%
src/flint/acb_mat.pyx         370    114    69%
src/flint/arb.pyx             835    180    78%
src/flint/functions.pyx        61     12    80%
src/flint/acb.pyx             891    171    81%
test/dtest.py                   8      1    88%
src/flint/fmpz.pyx            340     34    90%
src/flint/fmpq.pyx            181     15    92%
src/flint/fmpz_poly.pyx       295     23    92%
src/flint/fmpz_mat.pyx        268     19    93%
src/flint/fmpq_poly.pyx       234     16    93%
src/flint/nmod.pyx             96      6    94%
src/flint/fmpq_mat.pyx        248     15    94%
src/flint/fmpz_series.pyx     186     11    94%
src/flint/fmpq_series.pyx     369     18    95%
src/flint/pyflint.pyx         132      6    95%
src/flint/nmod_mat.pyx        222     10    95%
src/flint/nmod_poly.pyx       199      5    97%
src/flint/__init__.py           1      0   100%
-----------------------------------------------
TOTAL                        6415   1736    73%
```

oscarbenjamin:
> Tests are added for almost full line coverage

The remaining parts that aren't covered are mostly:

1. Python 2 conditional blocks (I'll remove in a subsequent PR)
2. Lines with staticmethod, property or cpdef signatures. I'm not sure if these are being incorrectly reported as missed by coverage. I think that the C-code has some invisible branches that I don't see how to hit but it's difficult to decipher from looking at the generated code.
3. Impossible lines e.g. I think this one:
https://github.com/fredrik-johansson/python-flint/blob/e08d3dfc236f6b38fcc11bcde5c621993761a806/src/flint/fmpq_series.pyx#L203-L210
The code above already tested that s and t are not zero so valuation will be nonnegative and is the index of the first nonzero coefficient. Am I misunderstanding something or am I right that `fmpz_is_zero` will never return true here?

fredrik-johansson:
Equality of series is a bit tricky. It depends whether you want to compare series as symbolic expressions, as elements of R[x] / x^n, or as elements of R[[x]]. In the latter case, the O() term should be understood the same way as an error bound in interval arithmetic, and

``1 + x + O(x^2) == 1 + x + O(x^2)``

is actually undecidable (following the semantics of ``arb`` comparison, it should return ``False``).

BTW, I think ``fmpz_series([1,2,3],-1)`` should raise an exception. I don't see any good reason to allow invalid power series.

fredrik-johansson:
> The code above already tested that s and t are not zero so valuation will be nonnegative and is the index of the first nonzero coefficient. Am I misunderstanding something or am I right that fmpz_is_zero will never return true here?

Yeah, that looks bogus.

oscarbenjamin:
> BTW, I think `fmpz_series([1,2,3],-1)` should raise an exception. I don't see any good reason to allow invalid power series.

I don't necessarily disagree but I'm wondering why they were there before. Is it to allow something like a nan e.g.:
```python
In [50]: fmpq_series([1,1],1,3).derivative().derivative().derivative()
Out[50]: O(x^0)

In [51]: fmpq_series([1,1],1,3).derivative().derivative().derivative().derivative()
Out[51]: (invalid power series)
```

fredrik-johansson:
I don't remember why I ever put them in; probably because I was too lazy to implement all edge cases. Case in point: the derivative of ``O(x^0)`` should be ``O(x^0)``.

(I will add a C type for power series as part of the generics system eventually...)

oscarbenjamin:
> Equality of series is a bit tricky. It depends whether you want to compare series as symbolic expressions, as elements of R[x] / x^n, or as elements of R[[x]]. In the latter case, the O() term should be understood the same way as an error bound in interval arithmetic

I think that equality also needs to be understood differently from mathematical equality in the context of a programming language. For example in testing we really want `assert x == y` to mean that x and y are structurally the same although we can have some allowance for things like `fmpq(1) == fmpz(1) == 1`. My general interpretation of `==` is that it should be a relatively cheap operation that determines if two things are the same for most purposes.

The general question of mathematical equality or equivalence is something much more involved and that can be interpreted differently depending on the context. Of course it is possible that our notion of series equality can correspond to a particular notion of mathematical equality but in practical terms we want basic properties like `x == x` which (IIUC) the error bound notion would not allow.

fredrik-johansson:
Indeed:

    >>> arb("0.1") == arb("0.1")
    False

I think the interface should provide separate, explicit functions for testing equality of representations and testing mathematical equality, e.g. ``x.equal_repr(y)``, ``x.equal_value(y)``.

There are then pros and cons to having Python's ``==`` wrap either of these.

For mathematical equality, you could map the truth status to Python booleans in different ways.

Pessimistic logic (current ``arb`` behavior):

- True -> ``True``
- False -> ``False``
- Unknown -> ``False``

Optimistic logic:

- True -> ``True``
- False -> ``False``
- Unknown -> ``True``

Strict logic:

- True -> ``True``
- False -> ``False``
- Unknown -> some ``Exception``

Triple-valued logic:

- True -> ``True``
- False -> ``False``
- Unknown -> ``None`` or a new ``Unknown`` Python object

In the ``generic-rings`` wrapper (https://github.com/fredrik-johansson/generic-rings/blob/main/python/flint_ctypes.py), I'm experimenting with allowing users to select semantics at runtime:

        >>> a = (RR_arb(1) / 3) * 3
        >>> a
        [1.00000000000000 +/- 3.89e-16]
        >>> with strict_logic:
        ...     a == 1
        ...
          ...
        Undecidable: unable to decide x == y for x = [1.00000000000000 +/- 3.89e-16], y = 1.000000000000000 over Real numbers (arb, prec = 53)
        >>> with pessimistic_logic:
        ...     a == 1
        ...
        False
        >>> with optimistic_logic:
        ...     a == 1
        ...
        True

oscarbenjamin:
The consideration needed for `==` specifically in the context of Python is that it is used by all the core data structures. We can have different methods like `x.equals(y)` and `x.equals_modulo(y)` etc but the core parts of the language will only use `__eq__`. This is the case for sets, dicts, tuples, lists etc. It would be useful if the method used for these was different from the method used for the `==` operator even if they would usually coincide so that you wouldn't have things like this:
```python
>>> {1,1.0}
{1}
>>> {1.0,1}
{1.0}
```
With that being decided though the context in which we need to think about `==` specifically is (IMO) not direct usage by end users but rather usage by the core data structures of Python and other downstream library/application code. I envisage that the end user here is some way away e.g. python-flint is a component of a higher-level application and at that higher level `==` can be overloaded to do something different (e.g. the user is not necessarily even typing Python code or they do so through a wrapper interface like sage).

Having `==` behave strangely or even having its behaviour depend on global state is something that could be quite problematic for downstream code. The global state part couldn't work with anything hashable because it would invalidate any hash tables for example.

oscarbenjamin:
I wanted to run these tests on OSX arm64 after merging #42 but I don't seem to be able to get Cirrus CI to run on this PR.

oscarbenjamin:
> I think the interface should provide separate, explicit functions for testing equality of representations and testing mathematical equality, e.g. `x.equal_repr(y)`

Okay, I'll add this and use it for the series tests.

oscarbenjamin:
Okay, this is tested with OSX arm64 after rebasing (and the tests pass).

oscarbenjamin:
Current coverage report:
```
Name                        Stmts   Miss  Cover
-----------------------------------------------
src/flint/arb_series.pyx      563    492    13%
src/flint/arb_poly.pyx        195    170    13%
src/flint/arf.pyx             109     85    22%
src/flint/acb_poly.pyx        233    179    23%
src/flint/dirichlet.pyx        56     27    52%
src/flint/arb_mat.pyx         323    127    61%
src/flint/acb_mat.pyx         370    114    69%
src/flint/arb.pyx             835    180    78%
src/flint/functions.pyx        61     12    80%
src/flint/acb.pyx             891    171    81%
test/dtest.py                   8      1    88%
src/flint/fmpz.pyx            340     34    90%
src/flint/fmpq.pyx            181     15    92%
src/flint/fmpz_poly.pyx       295     23    92%
src/flint/fmpz_mat.pyx        268     19    93%
src/flint/fmpq_poly.pyx       234     16    93%
src/flint/nmod.pyx             96      6    94%
src/flint/fmpq_mat.pyx        248     15    94%
src/flint/fmpz_series.pyx     182     11    94%
src/flint/pyflint.pyx         132      6    95%
src/flint/fmpq_series.pyx     355     16    95%
src/flint/nmod_mat.pyx        222     10    95%
src/flint/nmod_poly.pyx       199      5    97%
src/flint/__init__.py           1      0   100%
-----------------------------------------------
TOTAL                        6397   1734    73%
```

oscarbenjamin:
Okay, let's merge this now to have better coverage testing in others PRs.

PR #41, opened 2022-12-21
maint: use msys2-ucrt instead of msys2-mingw64

oscarbenjamin:
I think that linking against ucrt is preferable rather than msvcrt although currently it doesn't seem to cause any problems.

There is some information about this here:
https://www.msys2.org/docs/environments/

My understanding is that python.org builds of CPython link against ucrt from Python 3.5 onwards:
https://matthew-brett.github.io/pydagogue/python_msvc.html

### DIFF
--- a/.github/workflows/buildwheel.yml
+++ b/.github/workflows/buildwheel.yml
@@ -21,7 +21,7 @@ jobs:
 
       - uses: msys2/setup-msys2@v2
         with:
-          msystem: mingw64
+          msystem: ucrt64
           # path-type inherit is used so that when cibuildwheel calls msys2 to
           # run bin/cibw_before_build_windows.sh the virtual environment
           # created by cibuildwheel will be available within msys2. The


oscarbenjamin:
This also needs to be changed:
https://github.com/fredrik-johansson/python-flint/blob/e08d3dfc236f6b38fcc11bcde5c621993761a806/bin/cibw_before_all_windows.sh#L17

oscarbenjamin:
This hasn't worked. The delvewheel output in the log has
```
2022-12-21T18:08:39.3257576Z External dependencies to copy into the wheel are
2022-12-21T18:08:39.3257972Z {   'libarb-2.dll',
2022-12-21T18:08:39.3258340Z     'libflint-17.dll',
2022-12-21T18:08:39.3258670Z     'libgcc_s_seh-1.dll',
2022-12-21T18:08:39.3259030Z     'libgmp-10.dll',
2022-12-21T18:08:39.3259388Z     'libmpfr-6.dll',
2022-12-21T18:08:39.3259757Z     'libwinpthread-1.dll'}
2022-12-21T18:08:39.3260178Z External dependencies not to copy into the wheel are
2022-12-21T18:08:39.3260656Z {'msvcrt.dll', 'kernel32.dll', 'user32.dll', 'python311.dll'}
```
So it's still linked against msvcrt.dll.

oscarbenjamin:
The problem is that `which gcc` in CI gives:
```
/c/ProgramData/Chocolatey/bin/gcc
```
So it's picking up the wrong gcc somehow.

oscarbenjamin:
Okay, this now fails in CI in the same way that it fails locally:
```
Error: Command python -c "import flint; print(str(flint.fmpz(2)))"
failed with code 3221225477. None
```
So the wheel gets built but crashes when used.

Basically with current master everything in the wheel is linked to msvcrt.dll. Although the wheel uses msvcrt.dll it does so consistently and that seems to work okay when the _flint.pyd is loaded into CPython which links with ucrt DLLs instead.

The changes here make it so that the dependencies are built linking to ucrt but the _flint.pyd itself still links to msvcrt. Having this inconsistency between _flint.pyd and the other DLLs in the shipped wheel makes it crash somehow. On the other hand it seems to work okay if everything in the wheel consistently uses msvcrt.dll even though that's inconsistent with what CPython itself does.

Having looked at other wheels I find that the wheels for numpy, scipy, gmpy2 etc also link to msvcrt.dll or vcruntime140.dll. Since scipy 1.9.0 the wheels are now ucrt only but this is a new thing that only happened when moving from setuptools/distutils to meson.

Also python-flint needs to move on from distutils and numpy.distutils which are both to be removed in Python 3.12. I propose to leave this PR as it is for now. I think it's fine for now to link against msvcrt.dll. The problem does need to be solved but so also does migrating to a new build system and it might be easier to solve the problem when that happens.

So there is a problem but and it needs to be solved soonish but I don't think it's a blocker. This PR can help as a starting point when it comes time to solve this but that can wait for now.

PR #42, opened 2023-01-07
maint: build wheels for macos arm64 in Cirrus CI

oscarbenjamin:
This adds Cirrus CI configuration that can build macOS arm64 wheels.

Cirrus CI needs to be enabled on the repo before this will work.


oscarbenjamin:
@fredrik-johansson can you enable Cirrus CI on this repo?

The link is here:
https://github.com/marketplace/cirrus-ci

Only the "public repositories" (free) level is needed. After it asks for a name and address and some access to your github account to see your repos. It's possible to enable it only on one repo though.

fredrik-johansson:
Cirrus CI seems to be installed now.

oscarbenjamin:
Thanks. Looks like it's working. It should take about 15 minutes to produce the wheels.

oscarbenjamin:
It's working. I've tested one of the wheels locally.

Next what is needed is to patch gap:
https://github.com/aleaxit/gmpy/issues/352#issuecomment-1360941902

I want to try to reproduce this issue first so that I can confirm the fix:
https://github.com/aleaxit/gmpy/issues/350

oscarbenjamin:
> I want to try to reproduce this issue

Actually I'm really not sure how I would reproduce that because obviously I've compiled python_flint rather than gmpy. The equivalent code for python_flint doesn't have any problem but that doesn't really prove anything. I think it's clear that GMP needs to be patched though (various other environments apply this patch for macos arm64).

oscarbenjamin:
I've added the patch for OSX arm64 and tested the wheel from Cirrus CI locally and it works.

oscarbenjamin:
Okay, I think this is good to go.

PR #45, opened 2023-08-08
Bump version to 0.4.0

oscarbenjamin:


oscarbenjamin:
I intend to merge this, push a 0.4.0 tag and then push the wheels and sdist to PyPI.

oscarbenjamin:
I've just pushed this to PyPI but when I downloaded the wheel on Linux it crashed:
```
$ python test/test.py 
test_pyflint...OK
Illegal instruction (core dumped)
```
On OSX all tests and doctests passed. I'm going to test Windows now...

oscarbenjamin:
> On OSX all tests and doctests passed. I'm going to test Windows now...

Windows is fine as well. So the problem is just Linux. (The one thing I didn't test!)

oscarbenjamin:
I've deleted the Linux wheels from PyPI. Now someone using `pip install python-flint` on Linux goes back to building from source as before.

Further investigation needed but it's late and I need to go to bed.

PR #46, opened 2023-08-09
Try manylinux2010

oscarbenjamin:
See https://github.com/fredrik-johansson/python-flint/issues/1#issuecomment-1671035655


oscarbenjamin:
The problem was fixed in #57

Issue #52, opened 2023-08-13
Python 3.12, pyproject.toml, distutils and a new build system


oscarbenjamin:
Currently python-flint builds using a setup.py and also uses distutils and numpy.distutils (see #4). As of Python 3.12 the stdlib distutils module is removed and also numpy is removing its own distutils module. This means that python-flint needs to move to a new build system somehow. Also setup.py should be removed and replaced with pyproject.toml and it probably makes sense to do that at the same time.

There are a few options for build system:

1. Try to use setuptools instead of distutils.
2. Switch to meson and meson-python.
3. Use cmake

Various Python projects seem to be choosing one of these three options. For pure Python projects there are other options like poetry, hatch, etc. Some of those can handle extension module building but I am not sure how well any of them works. The people who seem to know most about packaging extension modules seem to suggest either meson or cmake. In particular meson is being used by SciPy now and will be for NumPy soon (if not already).

To me switching to meson seems like the best long term option. I made a start on that here:
https://github.com/mesonbuild/meson-python/discussions/410
I discussed lots of things there like editable installs etc that make things a bit more complicated but the most important thing is just to have a build system that can produce wheels either for local install or upload to PyPI. Restricting to just producing wheels means that it would probably not be too hard to get something working with meson but the work I started here just needs to be carried through:
https://github.com/oscarbenjamin/mesontest

I don't know much about how using cmake would work here or what the potential upsides/downsides are.

Potentially the quickest fix for the removal of distutils is just to replace distutils with setuptools. It is probably not hard for someone to test that:

- Replace distutils with setuptools in setup.py
- Fix any immediate problems until you have a local build
- Push up to a PR in this repo and see if CI can build for all platforms and architectures.

One thing that might be a problem with setuptools is the use of MinGW compilers on Windows but this is tested in CI so we should know if it is working or not. My understanding is that numpy.distutils has special support for MinGW that is not in distutils or setuptools. If that means that the Windows build fails with setuptools then it probably means that python-flint just can't use setuptools.

Switching somehow to MSVC rather than MinGW would be good but I have already explored the options for doing that with GMP and I think we just have to give up on that. What *might* be possible is using MinGW to make a fat build of GMP and then using MSVC to compile everything else or perhaps just for the python-tools extension module. I don't know enough about MSVC to investigate that option so someone else would need to explore that. If we could get MSVC to pick up the MinGW compiled GMP then that might make it more possible to use setuptools or probably makes things easier in some sense for any of the build system choices.

[...]

PR #61, opened 2023-08-17
Introduce submodules into python-flint

GiacomoPope:
The hope for this pull request is a fairly large refactor of the code to remove all include from the `pyflint.pyx` file, and place everything into submodules.

I started by first factoring out `FlintContext` and some util functions from `pyflint.pyx`, and everything compiles, but when I try and use the module, I get import errors.

I can't seem to fix it, I don't know if this is a quirk of cython modules I'm not familiar with, or simply a dumb bug, but I'm hoping for some advice on how to properly import between submodules within the project. The module not found error doesn't make sense to me, as the functions are all found on compile time...

Sorry to have to ask for help so early!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 build/*
 dist/*
-src/flint/*.c
+src/flint/**/*.c
 src/flint/*.html
 doc/build/*
 fmake*
@@ -16,3 +16,4 @@ MANIFEST
 .coverage
 *.swp
 .python-version
+*.DS_Store
--- a/setup.py
+++ b/setup.py
@@ -63,15 +63,25 @@ if os.getenv('PYTHON_FLINT_COVERAGE'):
     compiler_directives['linetrace'] = True
 
 
-ext_modules = [
-    Extension(
-        "flint._flint", ["src/flint/pyflint.pyx"],
-        libraries=libraries,
-        library_dirs=default_lib_dirs,
-        include_dirs=default_include_dirs,
-        define_macros=define_macros,
-        )
+ext_files = [
+    ("flint._flint", ["src/flint/pyflint.pyx"]), # Main Module
+    # Submodules
+    ("flint.flint_base.flint_base", ["src/flint/flint_base/flint_base.pyx"]),
+    ("flint.flint_base.flint_context", ["src/flint/flint_base/flint_context.pyx"]),
+
 ]
+
+ext_options = {
+    "libraries" : libraries,
+    "library_dirs" : default_lib_dirs,
+    "include_dirs" : default_include_dirs,
+    "define_macros" : define_macros,
+}
+
+ext_modules = []
+for mod_name, src_files in ext_files:
+    ext = Extension(mod_name, src_files, **ext_options)
+    ext_modules.append(ext)
 
 for e in ext_modules:
     e.cython_directives = {"embedsignature": True}
@@ -92,3 +102,4 @@ setup(
     author_email='fredrik.johansson@gmail.com',
     license='MIT',
     classifiers=['Topic :: Scientific/Engineering :: Mathematics'])
+
--- a/src/flint/acb.pyx
+++ b/src/flint/acb.pyx
@@ -1,3 +1,6 @@
+from flint.flint_base.flint_base cimport flint_scalar
+from flint.flint_base.flint_context cimport getprec
+
 cdef int acb_set_python(acb_t x, obj, bint allow_conversion):
     cdef double re, im
 
--- a/src/flint/acb_mat.pyx
+++ b/src/flint/acb_mat.pyx
@@ -1,3 +1,7 @@
+from flint.flint_base.flint_context cimport getprec
+
+from flint.flint_base.flint_base cimport flint_mat
+
 cdef acb_mat_coerce_operands(x, y):
     if isinstance(y, (fmpz_mat, fmpq_mat, arb_mat)):
         return x, acb_mat(y)
--- a/src/flint/acb_poly.pyx
+++ b/src/flint/acb_poly.pyx
@@ -1,3 +1,8 @@
+from flint.flint_base.flint_context cimport getprec
+# TODO: waiting for fix on the roots method, currently 
+# globally defined.
+# from flint.flint_base.flint_base cimport flint_poly
+
 cdef acb_poly_coerce_operands(x, y):
     if isinstance(y, (int, long, float, complex, fmpz, fmpq, arb, acb, fmpz_poly, fmpq_poly, arb_poly)):
         return x, acb_poly(y)
--- a/src/flint/acb_series.pyx
+++ b/src/flint/acb_series.pyx
@@ -1,3 +1,6 @@
+from flint.flint_base.flint_context cimport getprec, getcap
+from flint.flint_base.flint_base cimport flint_series
+
 cdef acb_series_coerce_operands(x, y):
     if isinstance(y, (int, long, float, complex, fmpz, fmpz_poly, fmpz_series, fmpq, fmpq_poly, fmpq_series, arb, arb_poly, arb_series, acb, acb_poly)):
         return x, acb_series(y)
--- a/src/flint/arb.pyx
+++ b/src/flint/arb.pyx
@@ -1,3 +1,9 @@
+from cpython.version cimport PY_MAJOR_VERSION
+
+from flint.flint_base.flint_context cimport getprec
+from flint.flint_base.flint_base cimport flint_scalar
+from flint.utils.conversion cimport chars_from_str, str_from_chars
+
 cdef _str_trunc(s, trunc=0):
     if trunc > 0 and len(s) > 3 * trunc:
         left = right = trunc
--- a/src/flint/arb_mat.pyx
+++ b/src/flint/arb_mat.pyx
@@ -1,3 +1,6 @@
+from flint.flint_base.flint_context cimport getprec
+from flint.flint_base.flint_base cimport flint_mat
+
 cdef arb_mat_coerce_operands(x, y):
     if isinstance(y, (fmpz_mat, fmpq_mat)):
         return x, arb_mat(y)
--- a/src/flint/arb_poly.pyx
+++ b/src/flint/arb_poly.pyx
@@ -1,3 +1,8 @@
+from flint.flint_base.flint_context cimport getprec
+# TODO: waiting for fix on the roots method, currently 
+# globally defined.
+# from flint.flint_base.flint_base cimport flint_poly
+
 cdef arb_poly_coerce_operands(x, y):
     if isinstance(y, (int, long, float, fmpz, fmpq, arb, fmpz_poly, fmpq_poly)):
         return x, arb_poly(y)
--- a/src/flint/arb_series.pyx
+++ b/src/flint/arb_series.pyx
@@ -1,3 +1,6 @@
+from flint.flint_base.flint_context cimport getprec, getcap
+from flint.flint_base.flint_base cimport flint_series
+
 cdef arb_series_coerce_operands(x, y):
     if isinstance(y, (int, long, float, fmpz, fmpz_poly, fmpz_series, fmpq, fmpq_poly, fmpq_series, arb, arb_poly)):
         return x, arb_series(y)
--- a/src/flint/arf.pyx
+++ b/src/flint/arf.pyx
@@ -1,3 +1,6 @@
+from flint.flint_base.flint_context cimport getprec
+from flint.utils.conversion cimport prec_to_dps
+
 cdef class arf:
 
     cdef arf_t val
--- a/src/flint/dirichlet.pyx
+++ b/src/flint/dirichlet.pyx
@@ -1,3 +1,5 @@
+from flint.flint_base.flint_context cimport getprec
+
 cdef dict _dirichlet_group_cache = {}
 
 cdef class dirichlet_group(object):
--- a/src/flint/flint_base/flint_base.pxd
+++ b/src/flint/flint_base/flint_base.pxd
@@ -1,0 +1,19 @@
+cdef class flint_elem:
+    pass
+
+cdef class flint_scalar(flint_elem):
+    pass
+
+# TODO:
+# See .pyx file
+# cdef class flint_poly(flint_elem):
+#     pass
+
+cdef class flint_mpoly(flint_elem):
+    pass
+
+cdef class flint_mat(flint_elem):
+    pass
+
+cdef class flint_series(flint_elem):
+    pass
--- a/src/flint/flint_base/flint_base.pyx
+++ b/src/flint/flint_base/flint_base.pyx
@@ -1,0 +1,151 @@
+from flint.flint_base.flint_context cimport thectx
+
+cdef class flint_elem:
+    def __repr__(self):
+        if thectx.pretty:
+            return self.str()
+        else:
+            return self.repr()
+
+    def __str__(self):
+        return self.str()
+
+cdef class flint_scalar(flint_elem):
+    pass
+ 
+# TODO:
+# We cannot include this class until we can import
+# acb_poly, so for now we leave this class as a global
+# inside pyflint.pyx
+# 
+# cdef class flint_poly(flint_elem):
+#     """
+#     Base class for polynomials.
+#     """
+
+#     def __iter__(self):
+#         cdef long i, n
+#         n = self.length()
+#         for i in range(n):
+#             yield self[i]
+
+#     def coeffs(self):
+#         return list(self)
+
+#     def str(self, bint ascending=False):
+#         """
+#         Convert to a human-readable string (generic implementation for
+#         all polynomial types).
+
+#         If *ascending* is *True*, the monomials are output from low degree to
+#         high, otherwise from high to low.
+#         """
+#         coeffs = [str(c) for c in self]
+#         if not coeffs:
+#             return "0"
+#         s = []
+#         coeffs = enumerate(coeffs)
+#         if not ascending:
+#             coeffs = reversed(list(coeffs))
+#         for i, c in coeffs:
+#             if c == "0":
+#                 continue
+#             else:
+#                 if c.startswith("-") or (" " in c):
+#                     c = "(" + c + ")"
+#                 if i == 0:
+#                     s.append("%s" % c)
+#                 elif i == 1:
+#                     if c == "1":
+#                         s.append("x")
+#                     else:
+#                         s.append("%s*x" % c)
+#                 else:
+#                     if c == "1":
+#                         s.append("x^%s" % i)
+#                     else:
+#                         s.append("%s*x^%s" % (c, i))
+#         return " + ".join(s)
+
+#     def roots(self, **kwargs):
+#         """
+#         Isolates the complex roots of *self*. See :meth:`.acb_poly.roots`
+#         for details.
+#         """
+#         # TODO: 
+#         # To avoid circular imports, we import within the method
+#         from XXX.XXX.acb_poly import acb_poly
+#         return acb_poly(self).roots(**kwargs)
+
+cdef class flint_mpoly(flint_elem):
+    """
+    Base class for multivariate polynomials.
+    """
+
+cdef class flint_series(flint_elem):
+    """
+    Base class for power series.
+    """
+    def __iter__(self):
+        cdef long i, n
+        n = self.length()
+        for i in range(n):
+            yield self[i]
+
+    def coeffs(self):
+        return list(self)
+
+
+cdef class flint_mat(flint_elem):
+    """
+    Base class for matrices.
+    """
+
+    def repr(self):
+        if thectx.pretty:
+            return str(self)
+        # XXX
+        return "%s(%i, %i, [%s])" % (type(self).__name__,
+            self.nrows(), self.ncols(), (", ".join(map(str, self.entries()))))
+
+    def str(self, *args, **kwargs):
+        tab = self.table()
+        if len(tab) == 0 or len(tab[0]) == 0:
+            return "[]"
+        tab = [[r.str(*args, **kwargs) for r in row] for row in tab]
+        widths = []
+        for i in xrange(len(tab[0])):
+            w = max([len(row[i]) for row in tab])
+            widths.append(w)
+        for i in xrange(len(tab)):
+            tab[i] = [s.rjust(widths[j]) for j, s in enumerate(tab[i])]
+            tab[i] = "[" + (", ".join(tab[i])) + "]"
+        return "\n".join(tab)
+
+    def entries(self):
+        cdef long i, j, m, n
+        m = self.nrows()
+        n = self.ncols()
+        L = [None] * (m * n)
+        for i from 0 <= i < m:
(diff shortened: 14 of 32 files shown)


GiacomoPope:
This is the current error I get when running tests, but the compilation is OK

```
Jack: python-flint % source bin/activate                                
Jack: python-flint % python3 -m flint.test
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 112, in _get_module_details
  File "/Users/Jack/Documents/GitHub/python-flint/src/flint/__init__.py", line 1, in <module>
    from ._flint import *
  File "src/flint/pyflint.pyx", line 1, in init flint._flint
    """
ModuleNotFoundError: No module named 'flint.flint_base.flint_context
```

.gitignore
```diff
@@ -16,3 +16,4 @@ MANIFEST
 .coverage
 *.swp
 .python-version
+*.DS_Store
```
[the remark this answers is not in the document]
GiacomoPope: Oh thanks for spotting this. I'll see if the editor has an option to ensure an EOF

oscarbenjamin:
> ModuleNotFoundError: No module named 'flint.flint_base.flint_context

This error message indicates that at runtime there is expected to be a module of this name because of this line:
```python
from .flint_base.flint_context cimport FlintContext
```
I don't remember exactly how all of this works in Cython but the changes you have made here will not create that module because it either needs to be a .py file or an extension module but only one extension module is created:
https://github.com/flintlib/python-flint/blob/71bf11054576978a6bdb48cf3d054d3af301e765/setup.py#L66-L74

GiacomoPope:
Ahh ok, yeah so it's a problem with how `setup.py` is written and I need to properly learn the right way to do this.

A quick google linked to: https://stackoverflow.com/questions/44043620/how-to-prepare-cython-submodules which has an example: https://github.com/justou/cython_package_demo/blob/master/setup.py

```py
import os
from setuptools import setup, find_packages


def find_pyx(path='.'):
    pyx_files = []
    for root, dirs, filenames in os.walk(path):
        for fname in filenames:
            if fname.endswith('.pyx'):
                pyx_files.append(os.path.join(root, fname))
    return pyx_files


setup(
    name='dvedit',
    version='0.1',
    ext_modules=cythonize(find_pyx(), language_level=3),
    packages=find_packages(),
    )
```

But simply recursively searching like this seems like not the best practice, so maybe the best is simply to point to submodules as they are written.

oscarbenjamin:
From here:
https://cython.readthedocs.io/en/latest/src/userguide/sharing_declarations.html#the-cimport-statement

> It is important to understand that the [cimport](https://cython.readthedocs.io/en/latest/src/userguide/sharing_declarations.html#cimport) statement can only be used to import C data types, C functions and variables, and extension types. It cannot be used to import any Python objects, and (with one exception) it doesn’t imply any Python import at run time. If you want to refer to any Python names from a module that you have cimported, you will have to include a regular import statement for it as well.

> The exception is that when you use [cimport](https://cython.readthedocs.io/en/latest/src/userguide/sharing_declarations.html#cimport) to import an extension type, its type object is imported at run time and made available by the name under which you imported it. Using [cimport](https://cython.readthedocs.io/en/latest/src/userguide/sharing_declarations.html#cimport) to import extension types is covered in more detail below.

oscarbenjamin:
> But simply recursively searching like this seems like not the best practice

Better to list them explicitly I think. We will need to translate this to another build system later.

GiacomoPope:
The good news is simply modifying the setup as you would expect works, but I just move the error. From googling, this one seems to be a python3 vs. Python2 issue -- in fact, i see some `xrange` in certain places which is python2 syntax and generally this is EOL and shouldn't be supported IMO

Are you using python2 still? Confused about why my new module is making this get so upset... I'll debug, but probably tomorrow. My hope is once I get one submodule working, everything else will be the same. All this cython stuff is very new to me!



```
Jack: python-flint % python3 -m flint.test
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 112, in _get_module_details
  File "/Users/Jack/Documents/GitHub/python-flint/src/flint/__init__.py", line 1, in <module>
    from ._flint import *
  File "src/flint/pyflint.pyx", line 1, in init flint._flint
    """
ImportError: dynamic module does not define module export function (PyInit_flint_context)
```

oscarbenjamin:
> Are you using python2 still?

No.

Supported Python versions are 3.9, 3.10 and 3.11. A new build system is needed for 3.12 (#52).

Separately there is the concept of the Cython language level but that is set to 3:
https://github.com/flintlib/python-flint/blob/71bf11054576978a6bdb48cf3d054d3af301e765/setup.py#L54-L57
That should mean that the Cython code follows Python 3 although maybe `xrange` is still allowed there...

setup.py
```diff
@@ -35,60 +35,70 @@
...
-        arb = 'arb'
+        arb = "arb"
 
     libraries = [arb, "flint"]
-    (opt,) = get_config_vars('OPT')
-    os.environ['OPT'] = " ".join(flag for flag in opt.split() if flag != '-Wstrict-prototypes')
+    (opt,) = get_config_vars("OPT")
+    os.environ["OPT"] = " ".join(
+        flag for flag in opt.split() if flag != "-Wstrict-prototypes"
+    )
```
oscarbenjamin, line 47: Your editor is doing lots of strange things now.

GiacomoPope: Oh yeah, I have some automated linting with `black` configured from a previous project I was doing where I got lazy and found some tricks to automate formatting while collaborating with a bunch of people. I'll remove it for this, just so used to everything being modified I didn't notice.

GiacomoPope: Reverted the formatting back

GiacomoPope:
OK I think the issue is not anything very complicated, it's just more relative import issues.

```
[1/1] Cythonizing src/flint/flint_base/flint_context.pyx

Error compiling Cython file:
...
from ._flint cimport (
^

src/flint/flint_base/flint_context.pyx:1:0: 'flint/flint_base/_flint.pxd' not found
```

So when it tries to cythonize this as a submodule, it looks to:

```
from ._flint cimport
```

then reads this from the current directory rather than the directory relative to the top of the module, so maybe rather than `Extension` I need something like `SubExtension` or something, I'm just making stupid mistakes now, so I should do this after a good nights sleep

GiacomoPope:
Now I get a meaningful error:

```py
Error compiling Cython file:
...
    flint_set_num_threads
)
from flint.utils.conversion cimport prec_to_dps, dps_to_prec

cdef class FlintContext:
    cdef public bint pretty
                     ^

src/flint/flint_base/flint_context.pyx:11:21: C attributes cannot be added in implementation part of extension type defined in a pxd
```

Which makes sense, as I've written a cython class and I'm trying to allow it to be a submodule and it's getting cross at the c-types, so I need to go back to the cython man

oscarbenjamin:
> after a good nights sleep

Always worth having!

For what you are trying to do I would not be completely confident about exactly what code should be written or immediately how to diagnose any problem. In that situation I would make changes very carefully, fully testing things as I go, and attempting to confirm (or refute) my understanding at each stage so that I can learn as I go. It is easier to do this if you make *very* small changes one at a time.

GiacomoPope:
Yeah, that was kind of my goal with these initial commits. My hope was taking a single class from one file and sticking it in a submodule would be enough, but I'm already fighting these annoying bugs.

Understanding when an import is really importing from a `.so` and when it's importing from python seems to be the key, and from answers like this: https://stackoverflow.com/questions/44043620/how-to-prepare-cython-submodules it's hard to know if it's a case of cythonizing more files (by adding extensions) or by modifying the `__init__.py` files to link things together.

GiacomoPope:
Good news! Sleep was the right choice. I realised the problem was that the attributes initialised in `FlintContext` needed to be defined within the `pxd` rather than the `pyx` to ensure that the c-types were handled by C and not within python.

So in this example:

```cython
cdef class FlintContext:
    cdef public bint pretty
    cdef public long _prec
    cdef public long _dps
    cdef arf_rnd_t rnd
    cdef public bint unicode
    cdef public long _cap
```

```cython
cdef class FlintContext:
    def __init__(self):
        self.default()

    def default(self):
        self.pretty = True
        self.rnd = ARF_RND_DOWN
        self.prec = 53
        self.unicode = False
        self.threads = 1
        self.cap = 10
```

GiacomoPope:
I'm a bit stumped onto how to refactor with these global `ctx` and `thectx`. 

For example, say I make a submodule for `flint_elem`, this needs access to `ctx`

```py
cdef class FlintElem:
    def __repr__(self):
        if ctx.pretty:
            return self.str()
        else:
            return self.repr()

    def __str__(self):
        return self.str()
```

Let's say the above is in `/flint/flint_base/flint_elem.pyx` and I include this Extension so that I can do `from flint.flint_base.flint_elem import FlintElem`. This is what we would want to make on the child extensions, such as `FlintScalar(FlintElem)` and ultimately something like `nmod(FlintScalar), which we would want to make accessible within the module for people to use.

This means I would need within `pyflint.pyx` to have something like `from flint.nmod.nmod import nmod`, but if I'm importing `ctx` from `pyflint.pyx` to access the global for `FlintElem` I'll end up having circular imports.

I've tried searching around, but I can't see an example of how to handle these module globals. I wondered if I need to declare something within `flint/__init__.py` or something, but it's not obvious.

Advice appreciated

setup.py
```diff
@@ -70,7 +70,15 @@
...
-        )
+        ),
+    Extension(
+        "flint.flint_base.flint_context",
+        ["src/flint/flint_base/flint_context.pyx"],
+        libraries=libraries,
+        library_dirs=default_lib_dirs,
+        include_dirs=default_include_dirs,
+        define_macros=define_macros,
+    ),
```
oscarbenjamin, line 81: It would be better to factor this out like:
```
ext_files = [
    ('"flint._flint", ["src/flint/pyflint.pyx"]),
    ...
]

ext_options = {
    'libraries': libraries,
    ...
}

ext_modules = []
for mod_name, src_files in ext_files.items():
    ext = Extension(mod_name, src_files, **ext_options)
    ext_modules.append(ext)
```
Then it is easier to add to the `ext_files` list to make more extension modules and we don't need to duplicate the arguments in `ext_options`.

oscarbenjamin:
> but if I'm importing `ctx` from `pyflint.pyx` to access the global for `FlintElem` I'll end up having circular imports.

Is the solution not just to always import `ctx` from the module where it is defined?

setup.py
GiacomoPope: Agreed, addressed in commit https://github.com/flintlib/python-flint/pull/61/commits/f7c3bfc5f36d724a7c38b6a9a9ec6dfb0ff57bd9

oscarbenjamin:
One thing I am not sure about here is whether it is actually better to have many extension modules rather than just one. There are several separate things here:

- We want to split the code into multiple files (which is already the case).
- We want to be able to do incremental rebuilds and so we don't want to use `include`.
- Splitting everything into simple .pyx files without include seems to mean having separate modules at runtime

I'm not sure in Cython whether it is possible to solve the first two problems without actually having separate modules at runtime. At least currently we do not have separate modules at runtime but it depends on `include` and that is problematic because it prevents having incremental rebuilds, doesn't work with coverage etc.

What I'm also not sure about is whether it is a good or a bad thing to have multiple extension modules at runtime. Currently the `flint._flint` extension module is 13MB here on my Linux machine which seems surprisingly large to me.

Probably that 13MB is okay but if we split it into many smaller extension modules will the total size be approximately the same?

What effect does this have on file sizes on disk or on memory usage at runtime?

PyPI has file size limits for wheels etc so we need to be aware of those if we follow a path that might potentially increase the file size (plus obviously smaller file sizes are generally good). The thing I'm worried about is basically whether splitting the 13MB module into say 20 smaller modules means that the total memory usage is like 14MB or if it mean balloon to something like `20*13MB`.

Does having multiple extension modules mean that only some need to be loaded into memory at runtime? So if a user does not import `fq` then do they not pay any runtime memory cost for that extension module to exist on disk because it does not get loaded into memory?

GiacomoPope:
> Is the solution not just to always import ctx from the module where it is defined?

I thought this would be the case, but there's something extra happening, as at the moment the test is able to call `flint.ctx` when it want's to access this global, but if I do something like

```cython
from flint.flint_base.flint_context cimport FlintContext

cdef FlintContext thectx = FlintContext()
ctx = thectx
```
as a submodule and import `thectx, ctx` into `pyflint.pyx` this `flint.ctx` fails, and I don't really understand how the global in this file becomes an attribute of the module here but not when it's imported.

> Does having multiple extension modules mean that only some need to be loaded into memory at runtime? So if a user does not import fq then do they not pay any runtime memory cost for that extension module to exist on disk because it does not get loaded into memory?

I'm sorry, I'm not familiar enough with cython to say for sure, but my thought process here was that you would just take one big lump of C and split it around into smaller lumps of C which would at run time be imported as python stuff when needed. 

I've been trying to find some other nice cython libraries to take inspiration from, but I'm really only familiar with Sage and there's *a lot* going on there.

oscarbenjamin:
> I'm sorry, I'm not familiar enough with cython to say for sure

Neither am I but my point is that we need to measure the impact of the changes:

How big are the extension modules that are created with this PR on your machine?

How does that compare with the extension module that would be created without this PR?

Are the file sizes for the wheels generated in this PR different to the wheels with master?

The wheels themselves are accessible from as artifacts in GitHub actions connected to each CI run to be downloaded and inspected.

oscarbenjamin:
> we need to measure the impact of the changes:

Also what effect does this have on runtime memory usage? If you do e.g.
```
>>> from flint import *
```
and then check how much memory the process is using is it still the same?

oscarbenjamin:
> as a submodule and import `thectx, ctx` into `pyflint.pyx` this `flint.ctx` fails, and I don't really understand how the global in this file becomes an attribute of the module here but not when it's imported.

I'm not sure. Sometimes in cython you have to assign things for apparently no reason so maybe like:
```
from somewhere import x as y

x = y  # Apparently pointless assignment somehow changes things
```

GiacomoPope:
> we need to measure the impact of the changes

I 100% agree and we should catch this as early as possible so if this is a waste of time I waste less of it!!

> I'm not sure. Sometimes in cython you have to assign things for apparently no reason

So I have made `thectx` global and set `ctx = thectx` inside the module and now all the tests pass. Do you know if `ctx = thectx` is a pointer, or is it copying the Extension? If it's just a name then in other places of the code where `ctx` is being used, I assume it should actually be `thectx` (for example the snippet above for `flint_elem`.

The latest commit now has `thectx` removes and a few more functions split out of the mega-file.

> How big are the extension modules that are created with this PR on your machine?
> How does that compare with the extension module that would be created without this PR?
> Are the file sizes for the wheels generated in this PR different to the wheels with master?
> The wheels themselves are accessible from as artifacts in GitHub actions connected to each CI run to be downloaded and inspected.

I'm not currently in the best place to do these tests as I need to do some other work -- if you have the time I would appreciate you comparing my current branch with what you're familiar with on your machine. If you don't have time then I'll do more digging later today.

oscarbenjamin:
> If you don't have time then I'll do more digging later today.

I don't have time right now but there's no need to rush this.

Basically the change here is initially motivated for improving the development workflow but potentially has an impact on performance for end users. If this would worsen performance somehow for end users then it might be better just to accept the problems with the development workflow (or try to find some other solution to them).

On the other hand the change could potentially reduce memory usage at runtime because it might mean that a user can import only the pieces that they need (although we would probably need to stop exposing everything at top-level on the `flint` module for that to happen). If we can improve performance for end users *and* improve the development workflow then we should definitely do that.

oscarbenjamin:
There is one commit above for which the CI completed and checking the artifacts in CI the total size of all wheels (4 os/arch x 3 Python versions) plus sdist is 141MB. To the available precision that is *precisely* the same as the 141MB that I can see on the most recent master commit. It might be though that the pieces you are currently moving around are just too small (in memory/code terms) to make a noticeable difference.

GiacomoPope:
>  It might be though that the pieces you are currently moving around are just too small (in memory/code terms) to make a noticeable difference.

Yeah, at the moment really the only addition is this tiny `FlintContext` which just has a few attributes which it can get and set. 

I guess a "good" test would be taking one of the main flint inclusions and seeing what happens with that moved around. I think now I've done a few files this evening I could do a chunk more and I can see if the wheels change at all.

My hope in the long run is that this at the very worse keeps everything the same for the user while cleaning up the repo for further improvements. I would also hope that some clever refactoring would allow lower memory cost by allowing a user to do `from flint import fmpz_poly` instead of `from flint import *` and have everything work as they would expect.

oscarbenjamin:
> I would also hope that some clever refactoring would allow lower memory cost by allowing a user to do `from flint import fmpz_poly` instead of `from flint import *` and have everything work as they would expect.

A general problem in Python is that if you expose everything in `__init__.py` then it becomes impossible for anyone to import anything without loading *everything* into memory regardless of what is needed or used. The simple way to fix this is to use a submodule rather than `__init__.py` so it is like:
```python
from flint.everything import *  # conveniently import everything
from flint.things.fmpz import fmpz # import only specific things
```
The downside is that there is no way to change to that scheme after the fact without requiring all downstream code to change the way that it imports things. To some extent it could be possible to make compatibility wrappers though e.g. like:
```python
# __init__.py

def fmpz_poly(*args, **kwargs):
    """
    Wrapper for fmpz_poly

    Use
        from flint.things.fmpz_poly import fmpz_poly
    or
        from flint.everything import fmpz_poly
    to import the type directly.
    """
    # Maybe emit a DeprecationWarning here?
    from flint.things.fmpz_poly import fmpz_poly as _fmpz_poly
    return _fmpz_poly(*args, **kwargs)
```
That won't work with e.g. `isinstance(x, fmpz_poly)` but it could transparently work for most things.

GiacomoPope:
I don't want to overstep, but seeing as python-flint is still an early release version, I think if a change was made which reduced the memory footprint but kept the performance, potentially a minor-breaking change could be worth it? That's not up to me to decide though

Also, this was bothering me:

> Do you know if ctx = thectx is a pointer, or is it copying the Extension? 

So I quickly added a `ctx = thectx` and `test = thectx` and modifying `ctx` also changed `test`, so it's a pointer as expected. In my opinion, we should then have `ctx = thectx` in the `pyflint.pyx` to maintain how it is used in the past, but for all other instances (for example the Extensions for various types) we should instead directly import and reference `thectx`

For example:

```cython
cdef class FlintElem:
    def __repr__(self):
        if thectx.pretty:
            return self.str()
        else:
            return self.repr()

    def __str__(self):
        return self.str()
```

oscarbenjamin:
> potentially a minor-breaking change could be worth it?

Yes, potentially. We would need to refactor everything into submodules first though (this PR is a good start).

As it stands `libflint` will always be loaded into memory which is fairly large in itself but I imagine that the total size of the extension modules would grow a lot as more Flint types are added.

There might also be other ways to reduce the overall size of the Cython generated code (which could help with build time as well). I don't remember exactly but I think that if this is changed to True then it doubles the size of the C code and the time taken to compile it:
https://github.com/flintlib/python-flint/blob/71bf11054576978a6bdb48cf3d054d3af301e765/setup.py#L56
There might be other Cython settings that can be used to make more streamlined extension modules and potentially make things faster at runtime as well.

GiacomoPope:
Latest commit compiles with all tests passing.

`pyflint.pyx` now has no template classes, which have all been moved to `flint_base.flint_base` and for the flint types, the appropriate class has been cimported

GiacomoPope:
I guess the next thing is to take one of the proper flint classes and put it into its own submodule. This should give us a better idea about footprint etc. Happy that the general method of working here seems to be "working" though

oscarbenjamin:
> `pyflint.pyx` now has no template classes

Nice!

It looks like there is a problem building the sdist though:
```
  File "/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/Cython/Build/Dependencies.py", line 990, in cythonize
    module_list, module_metadata = create_extension_list(
                                   ^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/Cython/Build/Dependencies.py", line 825, in create_extension_list
    for file in nonempty(sorted(extended_iglob(filepattern)), "'%s' doesn't match any files" % filepattern):
  File "/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/Cython/Build/Dependencies.py", line 119, in nonempty
    raise ValueError(error_msg)
ValueError: 'src/flint/flint_base/flint_base.pyx' doesn't match any files
```

GiacomoPope:
The problem is very stupid: I didn't add the new .pyx on the latest commit.

oscarbenjamin:
> The problem is very stupid

We've all been there. Looks like it's working now!

I'll probably just merge this once CI passes. I think once I've merged the PR that will mean that CI runs automatically for your PRs in future (without me needing to "approve" each run).

> I guess the next thing is to take one of the proper flint classes and put it into its own submodule.

Yes, I agree. First though it would be good to spell out a plan in #15 for what the overall module layout should be so that we are clear what the expected end state looks like both in terms of what runtime modules there would be, and what the source layout would be.

GiacomoPope:
OK, let's make sure everything passes, if there are errors I can try and correct for them but I will leave further refactoring of the repo until we have a discussion in #15. I'm also going on vacation soon, so potentially depending on timings the work will be done somewhere other than this branch.

oscarbenjamin:
You need to mark the PR as "ready for review" before it can be merged.

I guess another thing to think about is adding something like an AUTHORS file so we can record attributions.

GiacomoPope:
Hmm ok. I actually don't really like this global_context file. It seems redundant. I want to have the class definition within flint_base.flint_context and then import thectx from here, but when I tried that thectx was `None` so I obviously was doing something stupid. 

I'll see if I can simplify the current context handling and that should fix the current issue with wheels too.

GiacomoPope:
OK, so I found a way to delete the strange global file and everything seems to be OK.

I have removed `flint_poly` from the refactor until we can import `acb_poly` to address issue #62, so the code is there, but commented out.

Happy for this to be reviewed for merging now, it seems to be working my end and I dont think I've made any breaking changes! (Famous last words??)

oscarbenjamin:
Okay, I think that this looks good.

Let's get it merged. Thanks!

oscarbenjamin:
Oh, there are merge conflicts.

I think it is because this predates #58.

GiacomoPope:
I can address those in a few hours. Also we have the option to depreciate roots() in this merge?

oscarbenjamin:
> Also we have the option to depreciate roots() in this merge?

Yes, that could be done as well.

oscarbenjamin:
I'm very confused by what happened here. I clicked the merge button but then GitHub told me there were merge conflicts. Now I see that it was merged and the changes from here are in master but this shows as "closed" rather than "merged".

GiacomoPope:
Yeah I have no idea. When I saw it was closed, I was going to ping you to ask you opened it again, then I noticed it had been merged... Maybe when I resolved the merge conflicts my end GitHub automatically finished the merge? I have no idea.

Issue #62, opened 2023-08-18
Template class calls `acb_poly` for method definition


GiacomoPope:
The `flint_poly` is a child class of `flint_elem` and is used to build other polynomial types on top of, however, `flint_poly` has a method `roots()` which converts the self to `acb` type, which I assume is not the intension.

Suggestion is to remove this method and include it to `acb` if needed.

https://github.com/flintlib/python-flint/blob/71bf11054576978a6bdb48cf3d054d3af301e765/src/flint/pyflint.pyx#L215-L220

oscarbenjamin:
Is this a problem because of refactoring (like a cyclic import)?

I guess `acb` is the only type that would always be able to represent the roots of any `fmpz_poly` or `fmpq_poly` although it probably shouldn't be used for `nmod_poly`.

Perhaps e.g. `fmpz_poly` should have some sort of roots method that only finds integer roots? It doesn't look like Flint has an immediate function for that.

Asking the user to convert to `acb_poly` explicitly does not seem unreasonable to me although it would be a breaking change.

fredrik-johansson:
By default, ``roots`` ought to return roots in the coefficient ring, e.g. integer roots for ``fmpz_poly`` and rational roots for ``fmpq_poly``.

This is supported explicitly by the generics system (though not all conversions have been implemented).

In ``flint_ctypes``, the ``roots`` method for ``gr_poly`` currently looks like this:

```
    def roots(self, domain=None):
        """
        Computes the roots in the coefficient ring of this polynomial,
        returning a tuple (``roots``, ``multiplicities``).
        If the ring is not algebraically closed, the sum of multiplicities
        can be smaller than the degree of the polynomial.
        If ``domain`` is given, returns roots in that ring instead.

            >>> (ZZx([3,2]) * ZZx([15,1])**2 * ZZx([-10,1])).roots()
            ([10, -15], [1, 2])
            >>> ZZx([1]).roots()
            ([], [])

        We consider roots of the zero polynomial to be ill-defined:

            >>> ZZx([]).roots()
            Traceback (most recent call last):
              ...
            ValueError

        We construct an integer polynomial with rational, real algebraic
        and complex algebraic roots and extract its roots over
        different domains:

            >>> f = ZZx([-2,0,1]) * ZZx([1, 0, 1]) * ZZx([3, 2])**2
            >>> f.roots()   # integer roots (there are none)
            ([], [])
            >>> f.roots(domain=QQ)    # rational roots
[...]

PR #63, opened 2023-08-18
Finish the `flint_base` submodule, depreciate old roots and factor out more utils

GiacomoPope:
Just a few additional things I wanted to do before finishing what I started in #61, this PR in part addresses Issue #62 by depreciating roots in the general case and asks the user to first convert to `acb_poly()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,13 @@ To do
 * Improved printing and string input/output
 * IPython hooks (TeX pretty-printing etc.)
 
+CHANGELOG
+-------------
+
+0.5.0
+
+- gh-63: The `roots` method of `arb_poly`, and `nmod_poly` is no longer supported. Use `acb_roots(p).roots()` to get the old behaviour of returning the roots as `acb`. Note that the `roots` method of `fmpz_poly` and `fmpq_poly` currently returns the complex roots of the polynomial.
+
 License
 ------------
 
--- a/src/flint/flint_base/flint_base.pyx
+++ b/src/flint/flint_base/flint_base.pyx
@@ -72,7 +72,7 @@ cdef class flint_poly(flint_elem):
 
         acb_poly(input_poly).roots()
         """
-        raise NotImplementedError('This method has been deprecated. Please instead use acb_poly(input_poly).roots()')
+        raise NotImplementedError('This method is no longer supported. To recover the complex roots first convert to acb_poly')
         
 
 
--- a/src/flint/flint_base/flint_base.pyx
+++ b/src/flint/flint_base/flint_base.pyx
@@ -1,3 +1,5 @@
+from warnings import warn
+
 from flint.flint_base.flint_context cimport thectx
 
 cdef class flint_elem:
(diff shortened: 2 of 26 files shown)


src/flint/flint_base/flint_base.pyx
```diff
@@ -10,78 +12,78 @@ cdef class flint_elem:
...
+                else:
+                    if c == "1":
+                        s.append("x^%s" % i)
+                    else:
+                        s.append("%s*x^%s" % (c, i))
+        return " + ".join(s)
+
+    def roots(self):
+        """
+        Depreciated function.
```
oscarbenjamin, line 71: This is a very common misspelling that I see a lot. I guess it makes more sense somehow when translated to English from some other languages.

The word is "deprecate" (to express disapproval of) rather than "depreciate" (to diminish in value over time).

src/flint/flint_base/flint_base.pyx
```diff
@@ -10,78 +12,78 @@ cdef class flint_elem:
...
+        
+        To recover roots of a polynomial, first convert to acb:
+
+        acb_poly(input_poly).roots()
+        """
+        warn('This method is deprecated. Please instead use acb_poly(input_poly).roots()', DeprecationWarning)
+        
+
```
oscarbenjamin, line 79: Warnings are useful only if we want to preserve existing behaviour while giving out the warning. This does not return the roots that would have previously been returned so existing behaviour is not preserved.

I think it is better just to raise an error here rather than give out a warning and implicity return None.

oscarbenjamin: Also if giving an error (or not preserving existing behaviour) then it does not really make sense to refer to this as "deprecated": rather it is "unsupported" or "disallowed" or something like that.

src/flint/flint_base/flint_base.pyx
GiacomoPope: Thanks for spotting this, spelling isn't my strong point. Will fix now.

src/flint/flint_base/flint_base.pyx
GiacomoPope: Yes, I agree, I'll swap this for
```py
NotImplementedError('This method has been deprecated. Please instead use acb_poly(input_poly).roots()')
```

oscarbenjamin: Also we should add a CHANGELOG section at the bottom of the README with something like:
```
CHANGELOG
---------

0.5.0

- gh-63: The `roots` method of `fmpz_poly`, `fmpq_poly` and `nmod_poly` (others?) is no longer supported. Use `acb_roots(p).roots()` to get the old behaviour of returning the roots as `acb`.

```

GiacomoPope: Maybe the simplest thing to do is simply say it's not supported and point to the `acb_poly()` then...

GiacomoPope: `fmpz_poly` still has roots, but they're complex roots, which is weird. `fmpq_poly` asks for roots of the numerator, which I believe is of type `fmpz_poly`, so the only ones currently broken I think are `arb_poly` and `nmod_poly`

```
>>> acb_poly([1,2]).roots()
[-0.500000000000000]
>>> arb_poly([1,2]).roots()
  File "src/flint/flint_base/flint_base.pyx", line 75, in flint.flint_base.flint_base.flint_poly.roots
    raise NotImplementedError('This method is no longer supported. To recover the complex roots first convert to acb_poly')
NotImplementedError: This method is no longer supported. To recover the complex roots first convert to acb_poly
>>> fmpz_poly([1,2]).roots()
[(-0.500000000000000, 1)]
>>> fmpq_poly([1,2]).roots()
[(-0.500000000000000, 1)]
>>> nmod_poly([1,2], 11).roots()
  File "src/flint/flint_base/flint_base.pyx", line 75, in flint.flint_base.flint_base.flint_poly.roots
    raise NotImplementedError('This method is no longer supported. To recover the complex roots first convert to acb_poly')
NotImplementedError: This method is no longer supported. To recover the complex roots first convert to acb_poly
```

README.md
```diff
@@ -122,6 +122,13 @@ To do
 * Improved printing and string input/output
 * IPython hooks (TeX pretty-printing etc.)
 
+CHANGELOG
+-------------
+
+0.5.0
+
+- gh-63: The `roots` method of `arb_poly`, and `nmod_poly` is no longer supported. Use `acb_roots(p).roots()` to get the old behaviour of returning the roots as `acb`. Note that the `roots` method of `fmpz_poly` and `fmpq_poly` currently returns the complex roots of the polynomial.
```
oscarbenjamin, line 130:
```suggestion
- gh-63: The `roots` method of `arb_poly`, and `nmod_poly` is no longer supported. Use `acb_roots(p).roots()` to get the old behaviour of returning the roots as `acb`. Note that the `roots` method of `fmpz_poly` and `fmpq_poly` previously returned the complex roots of the polynomial.
```

oscarbenjamin: I'm a bit confused here.

Why are the changes to `arb_poly` being described differently to `fmpz_poly`?

Is the roots method not being removed from all of them except `acb_poly`?

GiacomoPope: They still do. Although I've removed support from the very base class, the fmpz_poly has its own roots method which fmpq_poly inherits. One option is of course to delete the method here as well.

oscarbenjamin: Oh, I see.

Okay then this is fine for now.

GiacomoPope: In the future I would like integer roots from fmpz and rational roots from fmpq, but this can wait until we link things up with Flint3 and the genetics I hope. For now this removal allowed the most simple refactoring to start.

oscarbenjamin:
Looks good. Thanks!

PR #67, opened 2023-09-06
Refactor functions into submodule

GiacomoPope:
This PR aims to finish the work on refactoring to remove all include, fixing issue #15. 

The last file, `functions.pyx` has been renamed to `eval_helper.pyx` and is used in doctests for various flint types.


src/flint/types/acb.pyx
```diff
@@ -351,7 +351,7 @@ cdef class acb(flint_scalar):
         Complex sign function defined as a piecewise extension of
         the real sign function.
 
-            >>> from flint import showgood
+            >>> from flint.types.eval_helper import showgood
```
oscarbenjamin, line 354: The docstring should show how users are expected to import something.

Does `from flint import showgood` not work with these changes?

GiacomoPope: I could import showgood as all the other types are so it's available from flint itself but I left it this way as my understanding it was more for debugging? 

The change should be minor I hope, happy for you to make the call

oscarbenjamin: It is shown widely in the docs so it should be treated as an end user function. Previously the docs always showed it as being imported from `flint` so we should preserve that for now (and continue to show that in the docs).

I think that `from flint import showgood` should continue to work and should be what is shown in the docstrings.

GiacomoPope: Okay, I'm happy to make the changes to fix this (tomorrow, sleep now). Won't be more than a few extra lines of code :)

GiacomoPope:
Considering `showgood()` is a user facing function which we want to import, I have put this in 

- `src/flint/functions/showgood.pyx` 

and then I make this available in `__init__.py` on the module so 

```
from flint import showgood
```

works as expected. `flint/functions` can then be used for similar user-facing util functions.

oscarbenjamin:
This seems to duplicate the code from functions.pyx in two places.

Also an empty .pdx file has been added (I don't know what a .pdx file is).

GiacomoPope:
@oscarbenjamin I did something stupid when cleaning up and deleted the wrong branch (trying to do too many things) -- Do you have permission to reopen this? So sorry for the mess up.

oscarbenjamin:
This has a merge conflict now.

GiacomoPope:
Resolved, I can't keep up with all the versions!

src/flint/__init__.py
```diff
@@ -21,5 +21,5 @@
 from .types.acb_mat import *
 from .types.acb_series import *
 from .types.fmpz_mpoly import *
-
+from .functions.showgood import showgood
 __version__ = '0.4.4'
```
oscarbenjamin, line 25: I suggest putting two blank lines between these. The cause of the merge conflict is basically just not having a blank line here.

oscarbenjamin:
If this is adding a new package then it also needs to be add here:
https://github.com/flintlib/python-flint/blob/24267a5ebc16a176299d178641ba6f61313f4fa7/setup.py#L66-L73

oscarbenjamin:
Looks good. Thanks!

Issue #54, opened 2023-08-13
A lower-level interface to Flint


oscarbenjamin:
Currently the python-flint interface to Flint is sort of somewhere between a high-level interface and a lower-level interface. In #53 I mention some ideas related to fleshing out what a more complete higher-level interface might look like. It will take a long time though to design high-level interfaces to all parts of Flint though and those interfaces will always lag behind the features that exist in Flint itself.

Another idea would be to try to add a lower-level interface within python-flint that just directly wraps each individual C type and public C function. There could be Cython types for e.g. `ulong`, `slong`, `fmpz_t` etc and then Cython functions like `fmpz_add` that take these types. The code to use this sort of low-level interface would look exactly like the corresponding C code using Flint's C API with the possible exception that Python's reference counting could be used instead of needing explicit `*_free` calls.

The advantages of having a low-level interface like this are that:

- The Cython code could be generated in a semi-automatic way e.g. by scraping the function names from the Flint headers so that it always keeps up with the latest new functions in Flint.
- While the low-level interface might not be very convenient it would always be possible to call any Flint function from Python.
- It might be easier for python-flint contributors to prototype some new additions to python-flint in Python before subsequently transferring them to Cython.

fredrik-johansson:
> The Cython code could be generated in a semi-automatic way e.g. by scraping the function names from the Flint headers so that it always keeps up with the latest new functions in Flint.

Some others (eg Arblib.jl) have done this by scraping the documentation rather than the header files.

Mollywall:
Thanks for the info.

edgarcosta:
If we build a dictionary `name -> cfunction` something like
```
def call(fname, *args):
    return name_to_cfunction[fname](*args)
```
might work

oscarbenjamin:
I was imagining a script that would parse all of the docs and generate thousands of functions that are just wrappers around a call to each individual documented Flint function.

oscarbenjamin:
A first step for this is just to have all of out .pxd files generated fully automatically, proving that we are able to automatically parse all of the functions from the docs.

In gh-215 I have set this up for all of the `fmpz_*` types but so far that's only 6 modules out of about 70.

PR #77, opened 2023-09-09
Parserst

deinst:
This is code to generate most of a flintlib pxd file from an rst file.  It is ugly as sin, but it should work. (A less comprehensible version generated most of the existing pxd files)

### DIFF
--- a/bin/rst_to_pxd.py
+++ b/bin/rst_to_pxd.py
@@ -1,0 +1,165 @@
+from collections import defaultdict
+import re
+import sys
+import os
+import fnmatch
+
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
Nice. I added some command line options to set the paths (rather than hard-coding).

A future improvement would be to build on this script and get the .pxd files to the point where there is some script that generates all of them in one go.

An improvement beyond that would be to generate not just the .pxd files but a bunch of "level 0" .pyx files with callable Python functions that wraps every single one of these C functions (#54)

PR #78, opened 2023-09-09
Minor fixes for nmod

oscarbenjamin:
These are minimal fixes needed for SymPy to be able to use nmod.

### DIFF
--- a/src/flint/types/nmod.pyx
+++ b/src/flint/types/nmod.pyx
@@ -6,6 +6,7 @@ from flint.types.fmpz cimport fmpz
 from flint.types.fmpq cimport fmpq
 
 from flint.flintlib.fmpz cimport fmpz_t
+from flint.flintlib.nmod cimport nmod_pow_fmpz, nmod_inv
 from flint.flintlib.nmod_vec cimport *
 from flint.flintlib.fmpz cimport fmpz_fdiv_ui, fmpz_init, fmpz_clear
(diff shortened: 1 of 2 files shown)


src/flint/types/nmod.pyx
```diff
@@ -178,6 +182,8 @@ cdef class nmod(flint_scalar):
                 return NotImplemented
         if tval == 0:
             raise ZeroDivisionError("%s is not invertible mod %s" % (tval, mod.n))
+        if not s:
+            return s
         # XXX: check invertibility?
         x = nmod_div(sval, tval, mod)
```
oscarbenjamin, line 188: This change makes it so that `nmod(0, p) / nmod(a, p)` gives `nmod(0, p)`.

About invertibility: in the context of SymPy's use the modulus here will always be prime. Otherwise though if tval and mod are not coprime then this gives a core dump:
```python
In [3]: flint.nmod(1, 6) / flint.nmod(3, 6)
    Cannot invert modulo 3*0
Aborted (core dumped)
```
That is not very friendly in Python land but I am not sure if adding checks here for is worth it: you would need to compute `gcd(tval, mod)` and that calculation would always be unnecessary for prime `mod`. Ideally Flint should provide an efficient function that could return an error code rather than aborting.

fredrik-johansson: There is such a function: ``n_gcdinv``.

oscarbenjamin: Thanks. I've added this in #79

PR #81, opened 2023-09-12
Work around cython parameter passing bug.

deinst:
In the words of Mae West, "goodness had nothing to do with it".

In this case, she was wrong though, `__goodness` tickled a cython code generation bug if `__goodness` was compiled as a cdef function.

Made `__goodness` work as a cdef function by removing its default parameter values.

Boiling this down to a human comprehensible example suitable for a cython bug report is a work in progress.

It appears that cython has a bug in passing parameters with defaults to some recursive functions.  `__goodness` is one of them.  In this case if any of the parameters had defaults `parts` was passed a 0 (False).  Removing the defaults fixes this.

Also added the `showgood` submodule to the list of submodules to doctest.

### DIFF
--- a/src/flint/functions/showgood.pyx
+++ b/src/flint/functions/showgood.pyx
@@ -13,19 +13,18 @@ from flint.flintlib.acb cimport *
 
 ctx = thectx
 
-# xxx: this doesn't work when changed to a cdef function. why?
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
Nice.

Does this make showgood faster at all?

PR #83, opened 2023-09-14
Add fmpz mod

GiacomoPope:
This is a draft pull request adding the type `fmpz_mod`

Before continuing, I am a little confused about the `fmpz_mod_ctx` and it's behaving in a way which doesn't make much sense to me...

The context struct is of the form:

```cython
    ctypedef struct fmpz_mod_ctx_struct:
        fmpz_t n
        nmod_t mod
        ulong n_limbs[3]
        ulong ninv_limbs[3]
        fmpz_preinvn_struct * ninv_huge
    ctypedef fmpz_mod_ctx_struct fmpz_mod_ctx_t[1]
```

and as `n` is an `fmpz_t` type, I assumed this was the standard modulus. I think `mod` is then the modulus used for small input?

However, if I print out `ctx.n` and `ctx.mod` for various values I find:

```py
>>> from flint import *
>>> 
>>> fmpz_mod_ctx(100)
Stuff: ([100], {'n': 100, 'ninv': 5165088340638674452, 'norm': 57}, [0, 0, 0])
>>> fmpz_mod_ctx(10**10)
Stuff: ([10000000000], {'n': 10000000000, 'ninv': 13244520931996183421, 'norm': 30}, [0, 0, 0])
>>> fmpz_mod_ctx(10**100)
Stuff: ([4611721077084439544], {'n': 0, 'ninv': 0, 'norm': 0}, [0, 0, 0])
```

Is it obvious here what I'm doing wrong?? Looking at the data out (a list, a dict, and a list) it seems I don't understand the "python" version of the types in this struct at all :/

### DIFF
--- a/doc/source/fmpz_mod.rst
+++ b/doc/source/fmpz_mod.rst
@@ -1,5 +1,10 @@
-**fmpz** -- integers mod n
+**fmpz_mod** -- integers mod n
 ===============================================================================
+
+.. autoclass :: flint.fmpz_mod_ctx
+  :members:
+  :inherited-members:
+  :undoc-members:
 
 .. autoclass :: flint.fmpz_mod
   :members:
--- a/doc/source/fmpz_mod.rst
+++ b/doc/source/fmpz_mod.rst
@@ -1,7 +1,7 @@
-**nmod** -- integers mod wordsize n
+**fmpz** -- integers mod n
 ===============================================================================
 
-.. autoclass :: flint.nmod
+.. autoclass :: flint.fmpz_mod
   :members:
   :inherited-members:
   :undoc-members:
--- a/src/flint/test/test.py
+++ b/src/flint/test/test.py
@@ -1585,7 +1585,7 @@ def test_pickling():
         assert obj == obj2
 
 def test_fmpz_mod():
-    from flint import fmpz_mod_ctx
+    from flint import fmpz_mod_ctx, fmpz
 
     # Context tests
     F163 = fmpz_mod_ctx(163)
@@ -1634,8 +1634,19 @@ def test_fmpz_mod():
     # Addition
     assert F163(123) + F163(456) == F163(123 + 456)
     assert F163(123) + F163(456) == F163(456) + F163(123)
+    assert F163(123) + 456 == F163(123 + 456)
+    assert F163(123) + fmpz(456) == F163(456) + F163(123)
+
     test_inplace = F163(123) 
     test_inplace += F163(456) 
+    assert test_inplace == F163(123 + 456)
+
+    test_inplace = F163(123) 
+    test_inplace += 456
+    assert test_inplace == F163(123 + 456)
+
+    test_inplace = F163(123) 
+    test_inplace += fmpz(456)
     assert test_inplace == F163(123 + 456)
     
 
--- a/src/flint/test/test.py
+++ b/src/flint/test/test.py
@@ -1586,15 +1586,21 @@ def test_pickling():
 
 def test_fmpz_mod():
     from flint import fmpz_mod_ctx, fmpz
-
+    
     # Context tests
     F163 = fmpz_mod_ctx(163)
     assert F163.modulus() == 163
+
     big = 2**1024 - 1
     Fbig = fmpz_mod_ctx(2**1024 - 1)
     assert Fbig.modulus() == big
 
     # Rich comparisons 
+    assert raises(lambda: F163(123) > 0, TypeError)
+    assert raises(lambda: F163(123) >= 0, TypeError)
+    assert raises(lambda: F163(123) < 0, TypeError)
+    assert raises(lambda: F163(123) <= 0, TypeError)
+
     assert (F163(123) == 123) is True
     assert (F163(123) == 123 + 163) is True
     assert (F163(123) == 1) is False
@@ -1604,15 +1610,6 @@ def test_fmpz_mod():
     assert (F163(123) == F163(123 + 163)) is True
     assert (F163(123) == F163(1)) is False
     assert (F163(123) != F163(1)) is True
-
-    assert (Fbig(123) == 123) is True
-    assert (Fbig(123) == 123 + big) is True
-    assert (Fbig(123) == 1) is False
-    assert (Fbig(123) != 1) is True
-    assert (Fbig(123) == Fbig(123)) is True
-    assert (Fbig(123) == Fbig(123 + big)) is True
-    assert (Fbig(123) == Fbig(1)) is False
-    assert (Fbig(123) != Fbig(1)) is True
 
     # Is one, zero, canoncial
     assert (F163(0) == 0) is True
@@ -1625,6 +1622,7 @@ def test_fmpz_mod():
     assert F163(2).is_one() is False
     assert F163(123).is_canonical() is True
 
+
     # Arithmetic tests
 
     # Negation
@@ -1635,6 +1633,7 @@ def test_fmpz_mod():
     assert F163(123) + F163(456) == F163(123 + 456)
     assert F163(123) + F163(456) == F163(456) + F163(123)
     assert F163(123) + 456 == F163(123 + 456)
+    assert 456 + F163(123) == F163(123 + 456)
     assert F163(123) + fmpz(456) == F163(456) + F163(123)
     assert raises(lambda: F163(123) + Fbig(456), ValueError)
 
@@ -1649,6 +1648,102 @@ def test_fmpz_mod():
     test_inplace = F163(123) 
     test_inplace += fmpz(456)
     assert test_inplace == F163(123 + 456)
+
+    # Subtraction
+
+    assert F163(123) - F163(456) == F163(123 - 456)
+    assert F163(123) - 456 == F163(123 - 456)
+    assert F163(123) - 456 == F163(123) - F163(456)
+    assert F163(456) - 123 == F163(456) - F163(123)
+    assert 123 - F163(456) == F163(123) - F163(456)
+    # assert 456 - F163(123) == F163(456) - F163(123)
+    assert F163(123) - fmpz(456) == F163(123) - F163(456)
+    assert raises(lambda: F163(123) - Fbig(456), ValueError)
+
+    test_inplace = F163(123) 
+    test_inplace -= F163(456) 
+    assert test_inplace == F163(123 - 456)
+
+    test_inplace = F163(123) 
+    test_inplace -= 456
+    assert test_inplace == F163(123 - 456)
+
+    test_inplace = F163(123) 
+    test_inplace -= fmpz(456)
+    assert test_inplace == F163(123 - 456)
+
+    # Multiplication
+
+    assert F163(123) * F163(456) == (123 * 456) % 163
+    assert F163(123) * 456 == (123 * 456) % 163
+    assert 456 * F163(123) == (123 * 456) % 163
+
+    assert F163(1) * F163(123) == F163(1 * 123)
+    assert F163(2) * F163(123) == F163(2 * 123)
+    assert F163(3) * F163(123) == F163(3 * 123)
+    assert 1 * F163(123) == F163(1 * 123)
+    assert 2 * F163(123) == F163(2 * 123)
+    assert 3 * F163(123) == F163(3 * 123)
+    assert F163(123) * 1 == F163(1 * 123)
+    assert F163(123) * 2 == F163(2 * 123)
+    assert F163(123) * 3 == F163(3 * 123)
+    assert fmpz(1) * F163(123) == F163(1 * 123)
+    assert fmpz(2) * F163(123) == F163(2 * 123)
+    assert fmpz(3) * F163(123) == F163(3 * 123)
+
+    test_inplace = F163(123) 
+    test_inplace *= F163(456) 
+    assert test_inplace == F163(123 * 456)
+
+    test_inplace = F163(123) 
+    test_inplace *= 456
+    assert test_inplace == F163(123 * 456)
+
+    test_inplace = F163(123) 
+    test_inplace *= fmpz(456)
+    assert test_inplace == F163(123 * 456)
+
+    # Exponentiation 
+
+    assert F163(0)**0 == pow(0, 0, 163)
+    assert F163(0)**1 == pow(0, 1, 163)
+    assert F163(0)**2 == pow(0, 2, 163)
+    assert raises(lambda: F163(0)**(-1), ZeroDivisionError)
+
+    assert F163(123)**0 == pow(123, 0, 163)
+    assert F163(123)**1 == pow(123, 1, 163)
+    assert F163(123)**2 == pow(123, 2, 163)
+    assert F163(123)**3 == pow(123, 3, 163)
+    assert F163(123)**100 == pow(123, 100, 163)
+    assert F163(123)**(-1) == pow(123, -1, 163)
+    assert F163(123)**(-2) == pow(123, -2, 163)
+    assert F163(123)**(-3) == pow(123, -3, 163)
+    assert F163(123)**(-4) == pow(123, -4, 163)
+
+    # Inversion
+
+    assert raises(lambda: F163(0).__invert__(), ZeroDivisionError)
+    assert F163(123).__invert__() == pow(123, -1, 163)
+    assert F163(1).__invert__() == pow(1, -1, 163)
+    assert F163(2).__invert__() == pow(2, -1, 163)
+
+    assert F163(1).inverse(check=False) == pow(1, -1, 163)
+    assert F163(2).inverse(check=False) == pow(2, -1, 163)
+    assert F163(123).inverse(check=False) == pow(123, -1, 163)
+
+    # Division
+
+    assert raises(lambda: F163(1) / F163(0), ZeroDivisionError)
+    assert F163(123) / F163(456) == (123 * pow(456, -1, 163)) % 163
+    assert F163(123) / fmpz(456) == (123 * pow(456, -1, 163)) % 163
+    assert F163(123) / 456 == (123 * pow(456, -1, 163)) % 163
+
+    assert 1 / F163(2) ==  pow(2, -1, 163)
+    assert 1 / F163(123) ==  pow(123, -1, 163)
+    assert 1 / F163(456) ==  pow(456, -1, 163)
+
+    assert fmpz(456) / F163(123) == (456 * pow(123, -1, 163)) % 163
+    assert 456 / F163(123) == (456 * pow(123, -1, 163)) % 163
     
 
 all_tests = [
--- a/src/flint/test/test.py
+++ b/src/flint/test/test.py
@@ -1584,6 +1584,52 @@ def test_pickling():
         obj2 = pickle.loads(s)
         assert obj == obj2
 
+def test_fmpz_mod():
+    from flint import fmpz_mod_ctx
+
+    # Context tests
+    F163 = fmpz_mod_ctx(163)
+    assert F163.modulus() == 163
+    big = 2**1024 - 1
+    Fbig = fmpz_mod_ctx(2**1024 - 1)
+    assert Fbig.modulus() == big
+
+    # Rich comparisons 
+    assert raises(lambda: F163(123) > 0, TypeError)
+    assert raises(lambda: F163(123) >= 0, TypeError)
+    assert raises(lambda: F163(123) < 0, TypeError)
+    assert raises(lambda: F163(123) <= 0, TypeError)
+
+    assert F163(123) == 123
+    assert F163(123) == 123 + 163
+    assert F163(123) != 1
+    assert F163(123) == F163(123)
+    assert F163(123) == F163(123 + 163)
+    assert F163(123) != F163(1)
+
+    # Is one, zero, canoncial
+    assert F163(0) == 0
+    assert F163(0).is_zero()
+    assert not F163(0)
+    assert not F163(163)
+    assert F163(1).is_one()
+    assert F163(164).is_one()
+    assert not F163(2).is_one()
+    assert F163(123).is_canonical()
+
+    # Arithmetic tests
+
+    # Negation
+    assert -F163(123) == F163(-123) == (-123 % 163)
+    assert -F163(1) == F163(-1) == F163(162)
+
+    # Addition
+    assert F163(123) + F163(456) == F163(123 + 456)
+    assert F163(123) + F163(456) == F163(456) + F163(123)
+    test_inplace = F163(123) 
+    test_inplace += F163(456) 
+    assert test_inplace == F163(123 + 456)
+    
 
 all_tests = [
     test_pyflint,
@@ -1603,4 +1649,5 @@ all_tests = [
     test_nmod_poly,
     test_nmod_mat,
     test_arb,
+    test_fmpz_mod,
 ]
--- a/src/flint/test/test.py
+++ b/src/flint/test/test.py
@@ -1595,6 +1595,14 @@ def test_fmpz_mod():
     Fbig = fmpz_mod_ctx(2**1024 - 1)
     assert Fbig.modulus() == big
 
+    F163_copy = fmpz_mod_ctx(163)
+    assert F163_copy == F163
+    assert F163_copy != Fbig
+    assert str(F163) == "fmpz_mod_ctx(163)"
+    assert repr(F163) == "Context for fmpz_mod with modulus: 163"
+
+    # Type tests
+
     # Rich comparisons 
     assert raises(lambda: F163(123) > 0, TypeError)
     assert raises(lambda: F163(123) >= 0, TypeError)
@@ -1605,11 +1613,19 @@ def test_fmpz_mod():
     assert (F163(123) == 123 + 163) is True
     assert (F163(123) == 1) is False
     assert (F163(123) != 1) is True
-
     assert (F163(123) == F163(123)) is True
     assert (F163(123) == F163(123 + 163)) is True
     assert (F163(123) == F163(1)) is False
     assert (F163(123) != F163(1)) is True
+
+    assert (hash(F163(123)) == hash(123)) is True
+    assert (hash(F163(F163(123))) == hash(123)) is True
+    assert (hash(F163(123)) == hash(1)) is False
+    assert (hash(F163(123)) != hash(1)) is True
+    assert (hash(F163(123)) == hash(F163(123))) is True
+    assert (hash(F163(123)) == hash(F163(123 + 163))) is True
+    assert (hash(F163(123)) == hash(F163(1))) is False
+    assert (hash(F163(123)) != hash(F163(1))) is True
 
     # Is one, zero, canoncial
     assert (F163(0) == 0) is True
@@ -1620,8 +1636,14 @@ def test_fmpz_mod():
     assert F163(164).is_one() is True
     assert F163(1).is_one() is True
     assert F163(2).is_one() is False
-    assert F163(123).is_canonical() is True
 
+    # int, str, repr
+    assert str(F163(11)) == "11"
+    assert repr(F163(-1)) == "162"
+    assert repr(F163(11)) == "fmpz_mod(11, 163)"
+    assert repr(F163(-1)) == "fmpz_mod(162, 163)"
+
+    assert F163(5).__pos__() == F163(5)
 
     # Arithmetic tests
 
@@ -1631,6 +1653,8 @@ def test_fmpz_mod():
 
     # Addition
     assert F163(123) + F163(456) == F163(123 + 456)
+    assert raises(lambda: F163(123) + "AAA", TypeError)
+
     assert F163(123) + F163(456) == F163(456) + F163(123)
     assert F163(123) + 456 == F163(123 + 456)
     assert 456 + F163(123) == F163(123 + 456)
--- a/src/flint/test/test.py
+++ b/src/flint/test/test.py
@@ -1587,209 +1587,189 @@ def test_pickling():
 def test_fmpz_mod():
     from flint import fmpz_mod_ctx, fmpz, fmpz_mod
     
+    p_sml = 163
+    p_med = 2**127 - 1
+    p_big = 2**255 - 19
+
+    F_sml = fmpz_mod_ctx(p_sml)
+    F_med = fmpz_mod_ctx(p_med)
+    F_big = fmpz_mod_ctx(p_big)
+
     # Context tests
-    F163 = fmpz_mod_ctx(163)
-    assert raises(lambda: fmpz_mod_ctx("AAA"), NotImplementedError)
+    assert raises(lambda: fmpz_mod_ctx("AAA"), TypeError)
     assert raises(lambda: fmpz_mod_ctx(-1), ValueError)
-    assert F163.modulus() == 163
+    assert F_sml.modulus() == p_sml
+    assert F_med.modulus() == p_med
+    assert F_big.modulus() == p_big
 
-    big = 2**1024 - 1
-    Fbig = fmpz_mod_ctx(2**1024 - 1)
-    assert Fbig.modulus() == big
+    F_big_copy = fmpz_mod_ctx(p_big)
+    assert F_big_copy == F_big
+    assert F_big != F_sml
+    assert hash(F_big_copy) == hash(F_big)
+    assert hash(F_big) != hash(F_sml)
+    assert F_big_copy != F_sml
+    assert F_big_copy != "A"
 
-    F163_copy = fmpz_mod_ctx(163)
-    assert F163_copy == F163
-    assert Fbig != F163
-    assert hash(F163_copy) == hash(F163)
-    assert hash(Fbig) != hash(F163), f"{hash(Fbig)}, {hash(F163)}"
-    assert F163_copy != Fbig
-    assert F163_copy != "A"
-
-    assert str(F163) == "fmpz_mod_ctx(163)"
-    assert repr(F163) == "Context for fmpz_mod with modulus: 163"
+    assert repr(F_sml) == "fmpz_mod_ctx(163)"
+    assert str(F_sml) == "Context for fmpz_mod with modulus: 163"
 
     # Type tests
+    assert raises(lambda: fmpz_mod(1, "AAA"), TypeError)
 
-    assert raises(lambda: fmpz_mod(1, "AAA"), ValueError)
+    # Test for small, medium and large char.
+    for F_test in [F_sml, F_med, F_big]:
+        test_mod = int(F_test.modulus())
+        test_x = (-123) % test_mod # canonical value
+        test_y = ((-456) % test_mod)**2 # non-canoncial value
 
+        F_test_copy = fmpz_mod_ctx(test_mod)
+        F_other = fmpz_mod_ctx(11)
 
-    # Rich comparisons 
-    assert raises(lambda: F163(123) > 0, TypeError)
-    assert raises(lambda: F163(123) >= 0, TypeError)
-    assert raises(lambda: F163(123) < 0, TypeError)
-    assert raises(lambda: F163(123) <= 0, TypeError)
+        assert raises(lambda: F_test(test_x) > 0, TypeError)
+        assert raises(lambda: F_test(test_x) >= 0, TypeError)
+        assert raises(lambda: F_test(test_x) < 0, TypeError)
+        assert raises(lambda: F_test(test_x) <= 0, TypeError)
 
-    assert (F163(123) == 123) is True
-    assert (F163(123) == 123 + 163) is True
-    assert (F163(123) == 1) is False
-    assert (F163(123) != 1) is True
-    assert (F163(123) == F163(123)) is True
-    assert (F163(123) == F163(123 + 163)) is True
-    assert (F163(123) == F163(1)) is False
-    assert (F163(123) != F163(1)) is True
+        assert (test_x == F_test(test_x)) is True, f"{test_x}, {F_test(test_x)}"
+        assert (124 != F_test(test_x)) is True
+        assert (F_test(test_x) == test_x) is True
+        assert (F_test(test_x) == test_x + test_mod) is True
+        assert (F_test(test_x) == 1) is False
+        assert (F_test(test_x) != 1) is True
+        assert (F_test(test_x) == F_test(test_x)) is True
+        assert (F_test(test_x) == F_test(test_x + test_mod)) is True
+        assert (F_test(test_x) == F_test(1)) is False
+        assert (F_test(test_x) != F_test(1)) is True
 
-    assert (hash(F163(123)) == hash(123)) is True
-    assert (hash(F163(F163(123))) == hash(123)) is True
-    assert (hash(F163(123)) == hash(1)) is False
-    assert (hash(F163(123)) != hash(1)) is True
-    assert (hash(F163(123)) == hash(F163(123))) is True
-    assert (hash(F163(123)) == hash(F163(123 + 163))) is True
-    assert (hash(F163(123)) == hash(F163(1))) is False
-    assert (hash(F163(123)) != hash(F163(1))) is True
+        assert (hash(F_test(test_x)) == hash(test_x)) is True
+        assert (hash(F_test(F_test(test_x))) == hash(test_x)) is True
+        assert (hash(F_test(test_x)) == hash(1)) is False
+        assert (hash(F_test(test_x)) != hash(1)) is True
+        assert (hash(F_test(test_x)) == hash(F_test(test_x))) is True
+        assert (hash(F_test(test_x)) == hash(F_test(test_x + test_mod))) is True
+        assert (hash(F_test(test_x)) == hash(F_test(1))) is False
+        assert (hash(F_test(test_x)) != hash(F_test(1))) is True
 
-    # Is one, zero, canoncial
-    assert (F163(0) == 0) is True
-    assert F163(0).is_zero() is True
-    assert not F163(0)
-    assert not F163(163)
-    assert F163(1).is_one() is True
-    assert F163(164).is_one() is True
-    assert F163(1).is_one() is True
-    assert F163(2).is_one() is False
+        # Is one, zero
+        assert (F_test(0) == 0) is True
+        assert F_test(0).is_zero() is True
+        assert not F_test(0)
+        assert not F_test(test_mod)
+        assert F_test(1).is_one() is True
+        assert F_test(test_mod + 1).is_one() is True
+        assert F_test(1).is_one() is True
+        assert F_test(2).is_one() is False
 
-    # int, str, repr
-    assert str(F163(11)) == "11"
-    assert str(F163(-1)) == "162"
-    assert repr(F163(11)) == "fmpz_mod(11, 163)"
-    assert repr(F163(-1)) == "fmpz_mod(162, 163)"
+        # int, str, repr
+        assert str(F_test(11)) == "11"
+        assert str(F_test(-1)) == str(test_mod - 1)
+        assert repr(F_test(11)) == f"fmpz_mod(11, {test_mod})"
+        assert repr(F_test(-1)) == f"fmpz_mod({test_mod - 1}, {test_mod})"
 
-    assert F163(5).__pos__() == F163(5)
+        assert +F_test(5) == F_test(5)
 
-    # Arithmetic tests
+        # Arithmetic tests
 
-    # Negation
-    assert -F163(123) == F163(-123) == (-123 % 163)
-    assert -F163(1) == F163(-1) == F163(162)
+        # Negation
+        assert -F_test(test_x) == F_test(-test_x) == (-test_x % test_mod)
+        assert -F_test(1) == F_test(-1) == F_test(test_mod - 1)
 
-    # Addition
-    assert F163(123) + F163(456) == F163(123 + 456)
-    assert F163(123) + F163_copy(456) == F163(123 + 456)
-    assert F163(123) + F163(456) == F163_copy(123 + 456)
-    assert raises(lambda: F163(123) + "AAA", TypeError)
+        # Addition
+        assert F_test(test_x) + F_test(test_y) == F_test(test_x + test_y)
+        assert F_test(test_x) + F_test_copy(test_y) == F_test(test_x + test_y)
+        assert F_test(test_x) + F_test(test_y) == F_test_copy(test_x + test_y)
+        assert raises(lambda: F_test(test_x) + "AAA", TypeError)
 
-    assert F163(123) + F163(456) == F163(456) + F163(123)
-    assert F163(123) + 456 == F163(123 + 456)
-    assert 456 + F163(123) == F163(123 + 456)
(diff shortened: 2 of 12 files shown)


oscarbenjamin:
> ```python
> >>> fmpz_mod_ctx(100)
> Stuff: ([100], {'n': 100, 'ninv': 5165088340638674452, 'norm': 57}, [0, 0, 0])
> >>> fmpz_mod_ctx(10**10)
> Stuff: ([10000000000], {'n': 10000000000, 'ninv': 13244520931996183421, 'norm': 30}, [0, 0, 0])
> >>> fmpz_mod_ctx(10**100)
> Stuff: ([4611721077084439544], {'n': 0, 'ninv': 0, 'norm': 0}, [0, 0, 0])
> ```

Here three things are being printed.

The first is `self.val.n` where `self` is the cython `fmpz_mod_ctx` and `self.val` is the C `fmpz_mod_ctx_t` and then `self.val.n` is the `fmpz_t` for `n` in the context struct. An `fmpz_t` is an array of 1 `fmpz_struct`. In the Cython code we pretend that `fmpz_struct` is an `slong`.

In other words `self.val.n` should be (from Cython's perspective) an array of 1 `slong` and so it displays as e.g. `[100]`. An `fmpz` larger than `2**62-1` will be stored as a pointer rather than an inline slong which I think is why the 3rd example shows `[4611721077084439544]`.

The second thing printed is `self.val.mod` and `mod` is an `nmod_t` which is a struct having 3 values so Cython displays it as a dict showing its members.

The third thing being printed is `n_limbs` which is an array of 3 ulong so `[0, 0, 0]` makes sense.

The first two cases that you have picked are less than 64 bits (`FLINT_BITS`) so we go here:
https://github.com/flintlib/flint2/blob/283e514d39f8c8668123ebb981599cc94500e5c4/src/fmpz_mod/ctx_init.c#L37-L43

The third case is larger then 128 bits so we go to the general case:
https://github.com/flintlib/flint2/blob/283e514d39f8c8668123ebb981599cc94500e5c4/src/fmpz_mod/ctx_init.c#L26-L34

If you choose a modulus in between 64 and 128 bits then you should see `n_limbs` being used.

GiacomoPope:
Thanks so much Oscar, so the "generic" thing I misunderstood was that to get nice printing for `fmpz_t` I need to use `fmpz_get_intlong()` to access the right data. Importing it from `fmpz` directly upset cython so for now I included it again into `fmpz_mod` and now the context seems to be working as intended and also prints sensible data. 

I have weird things happening with the initialisation of the actual values but I'll figure this out I hope.

oscarbenjamin:
If `fmpz_mod` stores a reference to the modulus is it safe to have `set_modulus` modify the context?

Maybe the contexts should be immutable.

oscarbenjamin:
Maybe also there should be a way to create and `fnpz_mod` from the context like `ctx(3)` or `ctx.new(3)`

GiacomoPope:
OK very basic initialisation works, feels like that's most of the battle!

```py
>>> from flint import *
>>> mod_ctx = fmpz_mod_ctx(163)
>>> 
>>> fmpz_mod(-1, mod_ctx)
fmpz_mod(162, 163)
>>> fmpz_mod(165, mod_ctx)
fmpz_mod(2, 163)
```

GiacomoPope:
> If fmpz_mod stores a reference to the modulus is it safe to have set_modulus modify the context?
> Maybe the contexts should be immutable.

Hmm good point, this is worth thinking about. I only included it as I saw the function. I should also do the `__dealloc__`

GiacomoPope:
Also, rather than ctx.new(val) we could have it callable so we can do ctx(val), not sure what you think fits better for the library?

Then we'd basically have ZmodN = fmpz_mod_ctx(N) and we can make elements from ZmodN(val)

oscarbenjamin:
> Also, rather than ctx.new(val) we could have it callable so we can do ctx(val), not sure what you think fits better for the library?

I think either is fine. You could use this like:
```python
F11 = fmpz_mod_ctx(11)

a = F11(2)
b = F11(7)
c = a / b
```

GiacomoPope:
I agree with the above, I made something which allows the following:

```py
>>> from flint import *
>>> ZmodN = fmpz_mod_ctx(163)
>>> x = ZmodN(123)
>>> type(x)
<class 'flint.types.fmpz_mod.fmpz_mod'>
>>> x
fmpz_mod(123, 163)
>>> 
>>> y = fmpz(456)
>>> ZmodN(y)
fmpz_mod(130, 163)
>>> 
>>> z = fmpz_mod(789, ZmodN)
fmpz_mod(137, 163)
```

oscarbenjamin:
Currently we have
```python
In [3]: print(str(flint.nmod(2, 3)))
2

In [4]: print(repr(flint.nmod(2, 3)))
2
```
I'm not sure that I like that this does not display the modulus. It is at least inconsistent with what you show above though:
```python
>>> x
fmpz_mod(123, 163)
```
This is how the SymPy equivalent looks:
```python
In [5]: from sympy import GF

In [6]: F3 = GF(3)

In [7]: print(str(F3(2)))
2 mod 3

In [8]: print(repr(F3(2)))
SymmetricModularIntegerMod3(2)
```
I like the `2 mod 3` but I don't know if it is really necessary to show that (I am less keen in `SymmetricModularIntegerMod3(2)` but perhaps with a shorter name it could be okay).

We also have:
```python
In [9]: flint.nmod_poly([1, 2, 3], 7)
Out[9]: 3*x^2 + 2*x + 1

In [10]: flint.nmod_mat([[1, 2], [3, 4]], 7)
Out[10]: 
[1, 2]
[3, 4]
```
So the modulus is not shown there. This is how those look with SymPy:
```python
In [12]: Matrix([[1, 2], [3, 4]]).to_DM(domain=F3).to_dense()
Out[12]: DomainMatrix([[1 mod 3, 2 mod 3], [0 mod 3, 1 mod 3]], (2, 2), GF(3))

In [13]: Poly([1, 2, 3, 4], x, domain=F3)
Out[13]: Poly(x**3 - x**2 + 1, x, modulus=3)
```
I'm not sure what is best here. My natural inclination is that the modulus should not be invisible because nonzero characteristic is very different from zero characteristic.

We also have:
```python
In [15]: print(str(gmpy2.mpz(3)))
3

In [16]: print(repr(gmpy2.mpz(3)))
mpz(3)

In [17]: print(str(flint.fmpz(3)))
3

In [18]: print(repr(flint.fmpz(3)))
3
```

GiacomoPope:
I agree that `nmod` and `fmpz_mod` should match.

For another point of reference, in SageMath we have:

```py
sage: F = GF(163)
sage: x = F.random_element()
sage: x
116
sage: str(x)
'116'
sage: 
sage: K = Zmod(163)
sage: x = K.random_element()
sage: x
136
sage: str(x)
'136'
```

GiacomoPope:
Addition and negation and comparison are done, but not in a clean way. More to do, but will leave it for tomorrow. I have some Dad chores to do while everyone is sleeping!!

```py
>>> from flint import *
>>> 
>>> F163 = fmpz_mod_ctx(163)
>>> assert F163.modulus() == 163
>>> 
>>> assert F163(162) == F163(-1) == -F163(1)
>>> assert F163(123) + F163(456) == F163(123 + 456) 
>>> 
>>> assert F163(162) != F163(1)
```

src/flint/test/test.py
```diff
@@ -1584,6 +1584,52 @@ def test_pickling():
...
+    # Rich comparisons 
+    assert raises(lambda: F163(123) > 0, TypeError)
+    assert raises(lambda: F163(123) >= 0, TypeError)
+    assert raises(lambda: F163(123) < 0, TypeError)
+    assert raises(lambda: F163(123) <= 0, TypeError)
+
+    assert F163(123) == 123
+    assert F163(123) == 123 + 163
+    assert F163(123) != 1
+    assert F163(123) == F163(123)
```
oscarbenjamin, line 1606: I would generally write these tests like:
```python
assert (F163(123) == 123) is True
```
The reason is that most objects in Python are "truthy" so in this case
```python
assert F163(0) == 0
```
the `F163(0).__eq__` method could return almost anything and the test would still pass.

This consideration only applies to testing the behaviour of `__eq__` and related methods though. It is important to be very careful and explicit in the `__eq__` tests because if `__eq__` doesn't work then we can't trust any of the other tests!

GiacomoPope:
Allowed the addition of:

```py
    assert F163(123) + F163(456) == F163(123 + 456)
    assert F163(123) + F163(456) == F163(456) + F163(123)
    assert F163(123) + 456 == F163(123 + 456)
    assert F163(123) + fmpz(456) == F163(456) + F163(123)
```

So anything which can be cast to `fmpz` can now be added to `fmpz_mod` as well as `fmpz_mod` providing that the moduli match

GiacomoPope:
First draft of the class is now finished, I have basic addition, subtraction, multiplication, division, inversion and exponentiation.

GiacomoPope:
I'm not totally happy with the code quality yet, I think there's some parts where I'm writing too much redundant code because I don't know cython best practices. I would be really interested in feedback on how to make this as good as possible before starting on the next type :)

src/flint/test/test.py
```diff
@@ -1584,6 +1584,167 @@ def test_pickling():
...
+    # Is one, zero, canoncial
+    assert (F163(0) == 0) is True
+    assert F163(0).is_zero() is True
+    assert not F163(0)
+    assert not F163(163)
+    assert F163(1).is_one() is True
+    assert F163(164).is_one() is True
+    assert F163(1).is_one() is True
+    assert F163(2).is_one() is False
+    assert F163(123).is_canonical() is True
```
oscarbenjamin, line 1623: Can `is_canonical` ever be False?

This seems like the kind of thing that makes sense in C but where the type is actually `fmpz` but not in Python where we can wrap everything up in an `fmpz_mod` and ensure that `is_canonical` is always true.

GiacomoPope: Yeah, I have a TODO about this. I think the proper use case could be in `__mul__`, as we can only multiply `a*b` when both `a,b` are canonical. 

For the case when `a` is an `fmpz_mod` but `b` is any integer, we use `fmpz_mod_mul` but we must first reduce `b  mod n`, I suppose one could only do this reduction if `fmpz_mod_is_canonical` returns false, but just always reducing seemed fine?

I think the method on the class should be removed.

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,428 @@
...
+    def __iadd__(self, other):
+        # Add two fmpz_mod if moduli match
+        if typecheck(other, fmpz_mod):
+            if not self.ctx.val.n == (<fmpz_mod>other).ctx.val.n:
+                raise ValueError("moduli must match")
+
+            fmpz_mod_add(
+                self.val, self.val, (<fmpz_mod>other).val, self.ctx.val
+            )
+            return self
```
oscarbenjamin, line 207: Generally in Python something like this would not have an in-place mutating method e.g. `fmpz` does not, `int` does not etc. Mutable container types like a matrix would but I would not have defined any in-place methods for `fmpz_mod`.

Does it really make sense to add this (and other mutating methods)?

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,428 @@
...
+        return "fmpz_mod({}, {})".format(
+            fmpz_get_intlong(self.val),
+            self.ctx.modulus()
+        )
+
+    # TODO: seems ugly...
+    def __hash__(self):
+        return  hash(
+            (int(self), int(self.ctx.modulus()))
+        )
```
oscarbenjamin, line 144: I think this seems fine but note that `__hash__` should only be defined for immutable types (i.e. no `__iadd__` etc).

src/flint/types/fmpz_mod.pyx
GiacomoPope: If `fmpz` doesn't then I suppose we can remove these. The only reason I added them is it should be slightly faster as you don't have to init the result? Not sure if that level of benchmarking is interesting?

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,428 @@
...
+    def __floordiv__(self, other):
+        raise NotImplemented
+
+    def inverse(self, check=True):
+        cdef fmpz_mod res
+        res = fmpz_mod.__new__(fmpz_mod)
+        res.ctx = self.ctx
+
+        # Warning! This will crash if there's no solution
+        if check is False:
```
oscarbenjamin, line 391: Better to have this warning in a docstring rather than just a code comment.

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,428 @@
...
+
+            fmpz_mod_add(
+                res.val, self.val, (<fmpz_mod>other).val, self.ctx.val
+            )
+            return res
+
+        # Attempt to add an fmpz to an fmpz_mod
+        other = any_as_fmpz(other)
+        if other is NotImplemented:
+            raise NotImplementedError
```
oscarbenjamin, line 187: Methods like `__add__` should return `NotImplemented` for unrecognised types rather than raising `NotImplementedError`. The interpreter will translated `NotImplemented` into `TypeError` but only after both operands are tried i.e. if `a.__add__(b)` returns `NotImplemented` and `b.__radd__(a)` returns `NotImplemented` then `a + b` will give a `TypeError`. Raising `NotImplementedError` here means that `b.__radd__(a)` is not called and so not other type can overload `a + b` with `fmpz_mod`.

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,428 @@
...
+
+        # There's no `fmpz_mod_mul_fmpz`
+        # So we need to make sure that other is canonical
+        # TODO: should we check with `fmpz_mod_is_canoncial`
+        # or just always reduce?
+        fmpz_mod_set_fmpz(
+            (<fmpz>other).val, (<fmpz>other).val, self.ctx.val
+        )
```
oscarbenjamin, line 295: I think that there should be a function like `any_as_fmpz_mod` or perhaps a `cdef` method like `self.any_as_fmpz_mod(other)`.

I wouldn't worry too much about the performance of `a * b` for mixed types. If someone wants maximum speed they should be ensuring that `a` and `b` are already the same type, same modulus etc.

GiacomoPope: OK, I'm glad you said this as it was one of my questions. If I implement this then the arithmetic simplifies nicely as I'll do the conversion to `fmpz_mod` and then have the simplier cases.

GiacomoPope: In fact, the "Easiest" method is simply to call `other = self.ctx(other)` which does all the type conversions. 

A faster(?) method would be a `cdef` method which does things with the `cython` rather than `python` types I suppose.

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,428 @@
...
+
+        fmpz_mod_neg(
+            res.val, self.val, 
+            (<fmpz_mod_ctx_t>self.ctx.val))
+        return res
+
+    def __add__(self, other):
+        cdef fmpz_mod res
+        res = fmpz_mod.__new__(fmpz_mod)
+        res.ctx = self.ctx
```
oscarbenjamin, line 172: There are multiple exits from this function that do not return `res`. I think that `__new__` should not be called until we are sure that `res` will be returned.

The call to `__new__` is equivalent to calling `fmpz_init`

oscarbenjamin: > The call to `__new__` is equivalent to calling `fmpz_init`

Actually this is not true but it *should* be. We need a `__cinit__` method like for `fmpz`.

oscarbenjamin: > The call to `__new__` is equivalent to calling `fmpz_init`

The point here is that any call to `fmpz_init` should eventually be paired with a call to `fmpz_clear`. That can happen here if we are not going to return `res`. Otherwise it should happen in dealloc.

src/flint/test/test.py
oscarbenjamin: I think that `F163(anything)` should either raise an error or return a canonical `fmpz_mod` in `F163`. In other wose the `is_canonical` method for `fmpz_mod` should always return True (and so the method is not needed and should be removed to avoid confusion).

src/flint/types/fmpz_mod.pyx
oscarbenjamin: I would be interested to hear benchmarks but regardless the divergence in behaviour from all other numeric types in Python would just confuse people and lead to bugs. Understanding exactly how `+=` works with mutable types is something that confuses a lot of people especially when passing values to functions:
```python
def f(a):
    a += 1
    return a**2

b = 2
c = f(b)
# Most people would be suprised if this did not print 2:
print(b)
```

src/flint/types/fmpz_mod.pyx
GiacomoPope: So I understand, we want to have both a 

```cython
def __cinit__(self):
    fmpz_init(self.val)
    
def __dealloc__(self):
    fmpz_clear(self.val)
```

Included in the `fmpz` class, something similar for `fmpz_mod` and aslo I need to ensure that I am only calling `__new__` in places when I know `res` is returned?

oscarbenjamin: > I need them on the `fmpz_mod()` class?

Yes, and also `fmpz_mod_ctx`.

src/flint/types/fmpz_mod.pyx
oscarbenjamin: The hash function here should be consistent with the behaviour of `==` so any time `a == b` is True then we should also have that `hash(a) == hash(b)`.

You allow `a == b` to succeed between `fmpz_mod` and `int` which then means that it should have the same hash as an `int` regardless of modulus.

The hash algorithm used by CPython is:
```
In [24]: e = 10**1000

In [25]: e % (2**61 - 1) == hash(e)
Out[25]: True
```
That is also the same in PyPy. I'm not sure if that should be considered a language feature or not.

The most portable option here is to call `hash(int(self))` but I expect that it can be a lot faster to compute self mod `2**61 - 1` and convert that to an `int`.

oscarbenjamin: Some tests should be added to the test suite to cover cases like this to be clear that the python-flint hash matches the int hash.

src/flint/types/fmpz_mod.pyx
GiacomoPope: Very naive benchmark, but can do something more scientific later

```py
>>> from flint import *
>>> import timeit
>>>
>>> Fp = fmpz_mod_ctx(10934940634374701148446891134773061114063995215204380092061981208900044999798983348516563438926479516948194331363194666667838046483659456628774161285399003)
>>>  
>>> a = Fp(1312529757321787310305008287695652105910635919800619521256615619505416908854385534618644512159270554422685672404970180005984257654287906025857278585166015)
>>> b = Fp(2743993601401652712330999846100462585643293332719558784670795396207932456640420188396741854989884607067889476240198970284190649813933943542680798609562697)
>>> 
>>> timeit.timeit("c = a + b; c += b", globals={"a":a, "b":b}, number=1_000_000)
0.2847036040038802
>>> timeit.timeit("c = a + b; c = c + b", globals={"a":a, "b":b}, number=1_000_000)
0.4899688330187928
```

src/flint/types/fmpz_mod.pyx
GiacomoPope: There is no `fmpz_mod_init` :/

https://flintlib.org/doc/fmpz_mod.html

oscarbenjamin: > There is no `fmpz_mod_init`

That's because it is `fmpz_init`.

oscarbenjamin:
Note, also there is `bin/coverage.sh` which will produce a test coverage report. you have to delete all of the .c files first because the module needs to be rebuilt with Cython tracing enabled:
```
$ find src -name '*.c' | xargs rm
$ bin/coverage.sh
$ xdg-open htmlcov/index.html
```
Currently `fmpz_mod.pyx` has 81% coverage but if you look at the html view then you can see which lines are not covered.

src/flint/types/fmpz_mod.pyx
oscarbenjamin: I think that a lower-level API could allow mutation but we should not allow it for the `fmpz_mod` type that is added here.

GiacomoPope:
I'm getting more weird:
```
dlopen(/Users/Jack/Documents/GitHub/python-flint/src/flint/types/fmpz.cpython-311-darwin.so, 0x0002): symbol not found in flat namespace '__fmpz_promote_set_si'
```

When running the `coverage` so I'll try and just eyeball what's missing and debug this boring problem later.

oscarbenjamin:
> Oh, I think this is because I deleted the `.c` but not the `.so`.

Maybe. I think if you delete the .c it should rebuild the .so though. At least that is what happens for me.

We need the equivalent of `make clean` really.

oscarbenjamin:
These are the lines reported as missing under coverage:
```
$ coverage report -m src/flint/types/fmpz_mod.pyx
Name                           Stmts   Miss  Cover   Missing
src/flint/types/fmpz_mod.pyx     181     35    81%   29-32, 46, 50, 66-72, 89-91, 129, 135-143, 157, 187, 202, 212, 237, 253, 262, 277, 287, 310, 320, 346, 356, 366, 383, 419
TOTAL                            181     35    81%
```
A lot of them are just error cases so there should be tests that various exceptions are raised/

GiacomoPope:
I'm getting all sorts of weird bugs since I tried running the coverage, so I've pushed what I have so far and will debug later. Got a few things to do. It's weird how fragile the local build is for me.

oscarbenjamin:
> It's weird how fragile the local build is for me.

This is all related to needing a new build system. Ideally the build system would do out of tree builds like meson does. Also it needs the equivalent of `make clean` and should be made to have a proper understanding of what the dependencies are between all of the files.

GiacomoPope:
```
test_fmpz_mod...Traceback (most recent call last):
  File "/private/var/folders/76/zy5ktkns50v6gt5g8r0sf6sc0000gn/T/cibw-run-l2v24_kg/cp39-macosx_arm64/venv-test/lib/python3.9/site-packages/flint/test/__main__.py", line 36, in run_tests
    test()
  File "/private/var/folders/76/zy5ktkns50v6gt5g8r0sf6sc0000gn/T/cibw-run-l2v24_kg/cp39-macosx_arm64/venv-test/lib/python3.9/site-packages/flint/test/test.py", line 1599, in test_fmpz_mod
    assert F163_copy == F163
```

Looks like I did something bad with my equality for contexts too (which I guess I would have spotted if I could run tests locally lol)

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,337 @@
...
+        if mod < 1:
+            raise ValueError("Modulus is expected to be positive")
+
+        # Init the context
+        fmpz_mod_ctx_init(self.val, (<fmpz>mod).val)
+
+    def __eq__(self, other):
+        if typecheck(other, fmpz_mod_ctx):
+            return self.val.n == (<fmpz_mod_ctx>other).val.n
+        return False
```
oscarbenjamin, line 66: Maybe it should be `self.modulus() == other.modulus()`.

Here you are directly accessing `self.val.n` which is an `fmpz_t`. Those cannot be compared directly at the C level (they might be pointers): you would need to use `fmpz_equal` to compare them in C.

GiacomoPope: I need to be careful actually that the same modulus but different context object doesn't cause issues.

oscarbenjamin: There could be a global context cache using `WeakValueDictionary`:
https://docs.python.org/3/library/weakref.html#weakref.WeakValueDictionary

```python
from weakref import WeakValueDictionary
cache = WeakValueDictionary()

def get_context(n):
    ctx = cache.get(n)
    if ctx is not None:
        return ctx
    ctx = cache.setdefault(n, fmpz_mod_ctx(n))
    return ctx
```
https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#making-extension-types-weak-referenceable

Then there would only ever be one context object for a given modulus in memory at a time.

The advantage of using `WeakValueDictionary` rather than an ordinary `dict` is that the context will be freed when it is not needed any more.

GiacomoPope:
Total mystery but I can't remove the 

```
ImportError: dlopen(/Users/Jack/Documents/GitHub/python-flint/src/flint/types/fmpz.cpython-311-darwin.so, 0x0002): symbol not found in flat namespace '__fmpz_promote_set_si'
```

Error. I've deleted everything and rebuilt and also restarted the computer. Bit of a mystery :/. I'll look for similar errors elsewhere.

oscarbenjamin:
I have a SymPy PR here: https://github.com/sympy/sympy/pull/25663 that adds support for SymPy to use `nmod` as its `GF(p)` domain. I have just tested what happens if I switch that to use `fmpz_mod` with this PR and it throws up some test failures in the SymPy test suite. With these changes all the SymPy and python-flint tests passed:
```diff
diff --git a/src/flint/test/test.py b/src/flint/test/test.py
index ffb0a0a..5ca65d2 100644
--- a/src/flint/test/test.py
+++ b/src/flint/test/test.py
@@ -1639,7 +1639,7 @@ def test_fmpz_mod():
 
     # int, str, repr
     assert str(F163(11)) == "11"
-    assert repr(F163(-1)) == "162"
+    assert str(F163(-1)) == "162"
     assert repr(F163(11)) == "fmpz_mod(11, 163)"
     assert repr(F163(-1)) == "fmpz_mod(162, 163)"
 
diff --git a/src/flint/types/fmpz_mod.pyx b/src/flint/types/fmpz_mod.pyx
index 56c9379..86776af 100644
--- a/src/flint/types/fmpz_mod.pyx
+++ b/src/flint/types/fmpz_mod.pyx
@@ -61,9 +61,12 @@ cdef class fmpz_mod_ctx:
         fmpz_mod_ctx_init(self.val, (<fmpz>mod).val)
 
     def __eq__(self, other):
-        if typecheck(other, fmpz_mod_ctx):
-            return self.val.n == (<fmpz_mod_ctx>other).val.n
-        return False
+        if not typecheck(other, fmpz_mod_ctx):
+            return NotImplemented
+        return self.modulus() == other.modulus()
+
+    def __hash__(self):
+        return hash(self.modulus())
     
     def modulus(self):
         """
```
The context object needs to be hashable and its `__eq__` needs to be well-behaved.

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,340 @@
...
+from flint.types.fmpz cimport (
+    fmpz,
+    any_as_fmpz,
+)
+
+cimport libc.stdlib
+
+# TODO: import this from types.fmpz somehow, it's not good
+# to have the function repeated here.
+cdef fmpz_get_intlong(fmpz_t x):
```
oscarbenjamin, line 26: You need to add this function in `fmpz.pxd` to be able to `cimport` it here.

oscarbenjamin:
```diff
diff --git a/src/flint/types/fmpz.pxd b/src/flint/types/fmpz.pxd
index 4317c89..074aa41 100644
--- a/src/flint/types/fmpz.pxd
+++ b/src/flint/types/fmpz.pxd
@@ -10,6 +10,7 @@ from flint.flintlib.fmpz cimport fmpz_t, fmpz_set_str, fmpz_set_si
 from cpython.version cimport PY_MAJOR_VERSION
 
 cdef int fmpz_set_any_ref(fmpz_t x, obj)
+cdef fmpz_get_intlong(fmpz_t x)
 
 cdef inline int fmpz_set_pylong(fmpz_t x, obj):
     cdef int overflow
diff --git a/src/flint/types/fmpz_mod.pyx b/src/flint/types/fmpz_mod.pyx
index 3c92408..092a1de 100644
--- a/src/flint/types/fmpz_mod.pyx
+++ b/src/flint/types/fmpz_mod.pyx
@@ -21,20 +21,7 @@ from flint.types.fmpz cimport (
 
 cimport libc.stdlib
 
-# TODO: import this from types.fmpz somehow, it's not good
-# to have the function repeated here.
-cdef fmpz_get_intlong(fmpz_t x):
-    """
-    Convert fmpz_t to a Python int or long.
-    """
-    cdef char * s
-    if COEFF_IS_MPZ(x[0]):
-        s = fmpz_get_str(NULL, 16, x)
-        v = int(str_from_chars(s), 16)
-        libc.stdlib.free(s)
-        return v
-    else:
-        return <slong>x[0]
+from flint.types.fmpz cimport fmpz_get_intlong
 
 cdef class fmpz_mod_ctx:
     """
```

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,340 @@
...
+        libc.stdlib.free(s)
+        return v
+    else:
+        return <slong>x[0]
+
+cdef class fmpz_mod_ctx:
+    """
+    """
+    def __cinit__(self, mod):
+        self.__init__(mod)
```
oscarbenjamin, line 43: `__cinit__` should not call `__init__`. The `__cinit__` function is called as part of `__new__` to create the object. The `__init__` method will be called later.

Here is where `fmpz_mod_ctx_init` should be called.

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,340 @@
...
+        fmpz_mod_ctx_clear(self.val)
+
+    def __init__(self, mod):
+        """
+        """
+        # Ensure modulus is fmpz type
+        if not typecheck(mod, fmpz):
+            mod = any_as_fmpz(mod)
+            if mod is NotImplemented:
+                raise NotImplementedError("TODO")
```
oscarbenjamin, line 55: This should be `TypeError` and a better error message.

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,340 @@
...
+            return fmpz_equal(self.val.n, (<fmpz_mod_ctx>other).val.n)
+        return False
+
+    def __hash__(self):
+        return hash(self.modulus())
+
+    def __repr__(self):
+        return "Context for fmpz_mod with modulus: {}".format(
+            self.modulus()
+        )
```
oscarbenjamin, line 84:
```
return f"Context for fmpz_mod with modulus: {self.modulus()}"
```

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,340 @@
...
+
+    def __str__(self):
+        return "fmpz_mod_ctx({})".format(
+            self.modulus()
+        )
```
oscarbenjamin, line 89: To me these two methods `__str__` and `__repr__` seem the wrong way round. Usually `repr` shows the code to create the object.

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,340 @@
...
+            else:
+                return not res
+        # TODO: is this the best method for comparison?
+        # Seems like I'm doing too many type conversions?
+        elif typecheck(s, fmpz_mod) and typecheck(t, int):
+            res = int(s) == t % (<fmpz_mod>s).ctx.modulus()
+            if op == 2:
+                return res
+            else:
+                return not res
```
oscarbenjamin, line 157:
```
else:
    return `NotImplemented`.
```

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,340 @@
...
+            res = ((<fmpz_mod>s).val[0] == (<fmpz_mod>t).val[0]) and \
+                  ((<fmpz_mod>s).ctx == (<fmpz_mod>t).ctx)
+            if op == 2:
+                return res
+            else:
+                return not res
+        # TODO: is this the best method for comparison?
+        # Seems like I'm doing too many type conversions?
+        elif typecheck(s, fmpz_mod) and typecheck(t, int):
+            res = int(s) == t % (<fmpz_mod>s).ctx.modulus()
```
oscarbenjamin, line 153: I would just do
```
if typecheck(t, int):
    t = fmpz_mod(t, self.ctx)
```
above.

src/flint/test/test.py
```diff
@@ -1584,6 +1584,212 @@ def test_pickling():
...
+    assert hash(Fbig) != hash(F163), f"{hash(Fbig)}, {hash(F163)}"
+    assert F163_copy != Fbig
+    assert F163_copy != "A"
+
+    assert str(F163) == "fmpz_mod_ctx(163)"
+    assert repr(F163) == "Context for fmpz_mod with modulus: 163"
+
+    # Type tests
+
+    assert raises(lambda: fmpz_mod(1, "AAA"), ValueError)
```
oscarbenjamin, line 1613: Should be TypeError

src/flint/test/test.py
```diff
@@ -1584,6 +1584,212 @@ def test_pickling():
         obj2 = pickle.loads(s)
         assert obj == obj2
 
+def test_fmpz_mod():
+    from flint import fmpz_mod_ctx, fmpz, fmpz_mod
+    
+    # Context tests
+    F163 = fmpz_mod_ctx(163)
+    assert raises(lambda: fmpz_mod_ctx("AAA"), NotImplementedError)
```
oscarbenjamin, line 1592: Should be TypeError

src/flint/test/test.py
```diff
@@ -1584,6 +1584,212 @@ def test_pickling():
         obj2 = pickle.loads(s)
         assert obj == obj2
 
+def test_fmpz_mod():
+    from flint import fmpz_mod_ctx, fmpz, fmpz_mod
+    
+    # Context tests
+    F163 = fmpz_mod_ctx(163)
```
oscarbenjamin, line 1591: There should be some tests with different moduli. Especially we know that flint itself distinguishes the smaller than `2**64` and the larger than `2**128` cases for different implementations and we need to test that those are working correctly.

The whole point of `fmpz_mod` is to go at least larger than `2**64` so we definitely need tests for that.

src/flint/test/test.py
```diff
@@ -1584,6 +1584,212 @@ def test_pickling():
...
+    assert (hash(F163(123)) == hash(123)) is True
+    assert (hash(F163(F163(123))) == hash(123)) is True
+    assert (hash(F163(123)) == hash(1)) is False
+    assert (hash(F163(123)) != hash(1)) is True
+    assert (hash(F163(123)) == hash(F163(123))) is True
+    assert (hash(F163(123)) == hash(F163(123 + 163))) is True
+    assert (hash(F163(123)) == hash(F163(1))) is False
+    assert (hash(F163(123)) != hash(F163(1))) is True
+
+    # Is one, zero, canoncial
```
oscarbenjamin, line 1640: canonical is gone

src/flint/test/test.py
```diff
@@ -1584,6 +1584,212 @@ def test_pickling():
...
+    assert F163(1).is_one() is True
+    assert F163(2).is_one() is False
+
+    # int, str, repr
+    assert str(F163(11)) == "11"
+    assert str(F163(-1)) == "162"
+    assert repr(F163(11)) == "fmpz_mod(11, 163)"
+    assert repr(F163(-1)) == "fmpz_mod(162, 163)"
+
+    assert F163(5).__pos__() == F163(5)
```
oscarbenjamin, line 1656: You would test this with `+F163(5)`.

src/flint/test/test.py
```diff
@@ -1584,6 +1584,212 @@ def test_pickling():
...
+    test_inplace += F163(456) 
+    assert test_inplace == F163(123 + 456)
+
+    test_inplace = F163(123) 
+    test_inplace += 456
+    assert test_inplace == F163(123 + 456)
+
+    test_inplace = F163(123) 
+    test_inplace += fmpz(456)
+    assert test_inplace == F163(123 + 456)
```
oscarbenjamin, line 1686: The in-place methods are removed so the references to `inplace` here are confusing. We don't need to separately test `+=` from `+` if we are not implementing the `__iadd__` etc methods. If we did want to test anything with `+=` then it would be to assert that the original object was not mutated which is not what this tests.

src/flint/test/test.py
```diff
@@ -1584,6 +1584,212 @@ def test_pickling():
...
+    assert F163(0)**0 == pow(0, 0, 163)
+    assert F163(0)**1 == pow(0, 1, 163)
+    assert F163(0)**2 == pow(0, 2, 163)
+    assert raises(lambda: F163(0)**(-1), ZeroDivisionError)
+    assert raises(lambda: F163(0)**("AA"), NotImplementedError)
+
+
+    assert F163(123)**0 == pow(123, 0, 163)
+    assert F163(123)**1 == pow(123, 1, 163)
+    assert F163(123)**2 == pow(123, 2, 163)
```
oscarbenjamin, line 1757: Can an `fmpz` be passed as the exponent?

src/flint/test/test.py
```diff
@@ -1584,6 +1584,212 @@ def test_pickling():
...
+    assert F163(123)**(-2) == pow(123, -2, 163)
+    assert F163(123)**(-3) == pow(123, -3, 163)
+    assert F163(123)**(-4) == pow(123, -4, 163)
+
+    # Inversion
+
+    assert raises(lambda: F163(0).__invert__(), ZeroDivisionError)
+    assert F163(123).__invert__() == pow(123, -1, 163)
+    assert F163(1).__invert__() == pow(1, -1, 163)
+    assert F163(2).__invert__() == pow(2, -1, 163)
```
oscarbenjamin, line 1770: This would be like `assert ~F163(123) == pow(123, -1, 163)`.

I think that this use of `~` is confusing but that is what the `__invert__` method is for: we should test the operator rather than the method.

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,340 @@
...
+    def __str__(self):
+        return "fmpz_mod_ctx({})".format(
+            self.modulus()
+        )
+
+    def __call__(self, val):
+        return fmpz_mod(val, self)
+
+# TODO: this seems stupid 
+def any_as_fmpz_mod(obj, ctx):
```
oscarbenjamin, line 95: This function is always called in a context where there is a `self` so I would make it a `cdef` method on the `fmpz_mod` class.

GiacomoPope:
At 99% coverage now, the only line is the 

```
else:
    return NotImplemented
```

Added back into the `__richcmp__`

Please have a look at my `__cinit__`, as I'm not sure what I have is best practices, otherwise all other points above should have been addressed.

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,311 @@
...
+
+        # Ensure modulus is positive
+        if mod < 1:
+            raise ValueError("Modulus is expected to be positive")
+
+        # Init the context
+        fmpz_mod_ctx_init(self.val, (<fmpz>mod).val)
+
+    def __dealloc__(self):
+        fmpz_mod_ctx_clear(self.val)
```
oscarbenjamin, line 40: In the same way that everything in Flint has an `_init` method and a `_clear` method every Cython type that has some C-level data needs to have `__cinit__` and `__dealloc__`. The only thing that should really be done here is just that `__cinit__` should call the Flint `_init` function to ensure that the C level data attached to the Cython class instance is in a consistent state ready for when `__dealloc__` is called and will call the `_clear` function.

Note that `__dealloc__` will still be called if this function exits with an exception so this function must not exit unless `fmpz_mod_ctx_clear` can be called safely right after. It should be considered an error to call the `_clear` function if the `_init` function was not called so any exit from `__cinit__` needs to have called `fmpz_mod_ctx_init`.

Probably what should be done here is to use `fmpz_mod_ctx_init` but set the modulus to 1. Then I guess the `__init__` method can call `fmpz_mod_ctx_set_modulus` to change it to the appropriate value. Or perhaps what is needed is just to set the modulus to one before doing any type checks and then change it if an exception is not going to be raised.

oscarbenjamin: https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#initialisation-methods-cinit-and-init

GiacomoPope: For now I will init the modulus to one in `cinit` and then move everything back to `__init__`.

oscarbenjamin:
I think this looks good.

What is needed now is documentation. We need an `fmpz_mod.rst` and some docstrings to fill it out:
https://fredrikj.net/python-flint/#reference

oscarbenjamin:
Looks like `nmod` could do with a few more docstrings...

GiacomoPope:
Do you recommend one of the classes for docstring examples? Otherwise I'll just do what seems sensible.

oscarbenjamin:
A reference is needed here:

https://github.com/flintlib/python-flint/blob/11bb708430a04715a5b231841e333cfc98a8db2c/doc/source/index.rst?plain=1#L41-L50

oscarbenjamin:
> There aren't many non-inbuilt functions so the docstrings may be a bit sparse for this type.

Agreed but the class level docstring should show the basic idea of how to construct an `fmpz_mod` at least and explain what it is. An `fmpz_mod` is not a complicated thing but the simple definition should at least be explained because the name `fmpz_mod` is not exactly intuitive.

doc/source/fmpz_mod.rst
```diff
@@ -0,0 +1,8 @@
+**fmpz** -- integers mod n
+===============================================================================
+
+.. autoclass :: flint.fmpz_mod
```
oscarbenjamin, line 4: The `fmpz_mod_ctx` class also needs to be documented here.

doc/source/fmpz_mod.rst
```diff
@@ -0,0 +1,8 @@
+**fmpz** -- integers mod n
+===============================================================================
```
oscarbenjamin, line 2: This should say `fmpz_mod`.

GiacomoPope: Thank you!!

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,357 @@
...
+        return f"fmpz_mod_ctx({self.modulus()})"
+
+    def __call__(self, val):
+        return fmpz_mod(val, self)
+
+cdef class fmpz_mod(flint_scalar):
+    """
+    The *fmpz_mod* type represents integer modulo an 
+    arbitrary-size modulus. For wordsize modulus, see
+    *nmod*.
```
oscarbenjamin, line 86: A Sphinx reference like
```
:class:`flint.types.nmod.nmod`
```
creates a clickable link in the html docs which is better.

GiacomoPope: Ahh thanks, I haven't ever used sphinx. I'll add this now.

oscarbenjamin: Actually this is a shorter version:
```
:class:`~.nmod`
```

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,357 @@
...
+
+    def __rtruediv__(s, t):
+        return fmpz_mod._div_(t, s)
+
+    def __floordiv__(self, other):
+        return NotImplemented
+
+    def inverse(self, check=True):
+        """
+        Computes a^-1 mod N
```
oscarbenjamin, line 305: It should be possible to use LaTeX here but possibly the LaTeX Sphinx extension is needed but not enabled...

GiacomoPope: Yeah I'm trying to do that but the naive 
```
math:`a^{-1} \mod N`
```
isn't working at least...

GiacomoPope: Oh, I got it.

GiacomoPope:
Added the latex to the doctoring which works (at least locally) and in the process spotted that `dirichlet` wasn't being imported to flint, so I fixed this! I guess `dirichlet_group` and `dirichlet_char` have no coverage in the tests? Maybe this should be an issue?

oscarbenjamin:
> I guess `dirichlet_group` and `dirichlet_char` have no coverage in the tests? Maybe this should be an issue?

I guess it should be. I'm not sure what the situation was before. Possibly the recent mega-refactoring affected this...

In any case if they are there then they should be made available and documented.

GiacomoPope:
>In any case if they are there then they should be made available and documented.

The functions are there and are fairly well docstring'd so, they're being tested in other ways. My commit above makes them available from `flint` again, so it's fixed the import issue and docstring / testing can always be improved so I guess there's no reason to make a specific issue right now.

GiacomoPope:
I'm off for the evening now. Thanks so much for your help with this class. I'm hoping there's a magnitude less discussion for my next type as I now have a much better idea of what I should be doing and how :)

❤️

src/flint/types/fmpz_mod.pyx
```diff
@@ -0,0 +1,364 @@
...
+
+        # For all other cases, the easiest is to first convert to
+        # fmpz type and set this way
+        if not typecheck(val, fmpz):
+            val = any_as_fmpz(val)
+            if val is NotImplemented:
+                raise NotImplementedError
+        fmpz_mod_set_fmpz(self.val, (<fmpz>val).val, self.ctx.val)
+
+    def any_as_fmpz_mod(self, obj):
```
oscarbenjamin, line 128: Can this be `cdef`?

oscarbenjamin: Or otherwise a leading underscore like `_any_as...`

oscarbenjamin: currently it shows in the docs but it should not be there.

oscarbenjamin:
Unless you have anything to add I think this looks good now. More improvements can be made to the documentation but I think that is an already general problem rather than something that must be fixed in this PR.

src/flint/types/fmpz_mod.pyx
GiacomoPope: Made it a cdef

GiacomoPope:
I think this is good to go now, at least, it doesn't feel much worse than the existing types? 

I want to eventually add the discrete log functions, but there's no rush for this and I'm more interested in getting `fmpz_mod_poly` working, as I need this for `fq` (and then hopefully with `fq`, I get `fq_poly` and finally I'll have enough to use `python-flint` for some of my isogeny things!! haha

oscarbenjamin:
One thing that would be good to think about going forwards is how to share testing code a bit e.g. most of the tests for `fmpz_mod` here could have been to some extent shared with `nmod`. As we add more types the benefits of sharing test code will grow and the current structure that is like `test_nmod`, `test_fmpz_mod` etc will become less and less efficient.

We should make new test functions that are like `test_add` or something that test that all types can add. Then it is easy to just include a new type in the test matrix rather than needing to rewrite all tests for each type.

oscarbenjamin:
> I think this is good to go now

I think so too. It will be in once the tests have passed.

GiacomoPope:
Agree about the testing. I also think it would be good to have randomised testing in certain places (addition, subtraction etc are all good candidates for this)

PR #80, opened 2023-09-11
Bump to Flint version 3

GiacomoPope:
This PR follows #43 but from the latest commits to master avoiding conflicts.

PR #69, opened 2023-09-06
Bump version to 0.4.3

oscarbenjamin:
See #66.

The wheels are missing for OSX arm64. Putting out a new release to fix it...

PR #84, opened 2023-09-16
Bump version to 0.5.0

GiacomoPope:
Version bump to show new type added. Removed a TODO from the README and another message in the change log to discuss the integer roots from #72


GiacomoPope:
Potentially we want to fix #74 before updating the version, but I don't have a good idea how to fix this myself.

README.md
```diff
@@ -128,6 +128,7 @@ CHANGELOG
 0.5.0
 
 - gh-63: The `roots` method of `arb_poly`, and `nmod_poly` is no longer supported. Use `acb_roots(p).roots()` to get the old behaviour of returning the roots as `acb`. Note that the `roots` method of `fmpz_poly` and `fmpq_poly` currently returns the complex roots of the polynomial.
+- gh-72: The roots method of `arb_poly` is not supported. Use either the `complex_roots` method or `acb_roots(p).roots()` to get the old behaviour of returning the complex roots. The `roots` method on `fmpz_poly` and `fmpq_poly` now return integer and rational roots respectively. To access complex roots on these types, use the `complex_roots` method. For `acb_poly`, both `roots` and `complex_roots` behave the same.
```
oscarbenjamin, line 131: We should also add notes for other PRs here.

- #83
- #80
- #78 #79

Big refactor across several versions:

- #75 #77 #67 #61

For 0.4.4 there should be

- #71 (adds Cython files again in sdist

For 0.4.3

- #69 (adds OSX arm64 wheels)

We should get better at adding release notes here. It is important for people to see these so that they can see that development is happening. Also I think that each note should list the names of the authors. It is informative for anyone who wants to understand the project to see that there are multiple people working on things.

oscarbenjamin: #63 was for 0.4.3 not 0.5.0

For 0.4.2

- #57 (Adds manylinux wheels)

For 0.4.1

- #48 Removes Linux wheels, updates instructions for building from source.

For 0.4.0

- #45 Adds wheels for Windows, OSX and manylinux but the Linux wheels are broken.

oscarbenjamin: For 0.4.3 #74 should be noted.

oscarbenjamin:
The Windows tests failed:
```
test_fmpz_mod...Flint exception (General error):
    mkstemp failed
```
That is because of #80. Note that my original PR for #80 noted this: https://github.com/flintlib/python-flint/pull/43#issuecomment-1675292534 although somehow the tests passed.

It is good that we spotted this before release: we can't use flint 3-alpha1 because of the Windows Flint bug that was fixed in https://github.com/flintlib/flint2/pull/1416

oscarbenjamin:
So we either revert #80 or wait for a new (pre-)release of Flint 3 before making a new release of python-flint.

GiacomoPope:
Do we have a timeline on the next pre-release? I guess that determines the best thing to do. I will find time today to address the comments above and do a more detailed change log for all these PR.

For "authors" how would you like to do it? I'm happy with whatever

oscarbenjamin:
> For "authors" how would you like to do it? I'm happy with whatever

I don't know really. Maybe just linking to the PRs is fine.

GiacomoPope:
For progress, I want to try and do fmpz_mod_poly, maybe if I finish this and the 3.0 bug remains we can revert #80 and if it's fixed by then it's all OK? I doubt the small addition for fmpz_mod will really help anyone right now.

oscarbenjamin:
We can consider reverting #80 at some point if it takes a while for a new release of Flint. for now I don't think it is urgent.

Especially if we expect that many new features might be added in a relatively short time there is no rush to get a release out for each one.

oscarbenjamin:
Rerun CI after gh-94 was merged

oscarbenjamin:
Some new Windows problem:
```
test_fmpz_mod...OK
test_fmpz_mod_dlog...
Error: Process completed with exit code 1.
```
Further investigation needed...

oscarbenjamin:
Twice now this pull request has picked up Windows problems that do not seem to show up elsewhere in CI. The changes here should not affect anything significant so I don't understand why that happens.

GiacomoPope:
Is it always the discrete log tests which crash? The only thought is there’s a bug either in Flint3 which means windows can’t do something in this instance or there’s a memory issue in the python-flint which is causing issues. It’s mysterious… 

I’ll have some time next week I hope to look into it.

oscarbenjamin:
I can reproduce the problem on Windows locally with master so it is not this PR that causes the problem but there is a bug that causes a crash on Windows.

It is always the discrete log test that crashes and specifically it crashes here:
https://github.com/flintlib/python-flint/blob/e6c007879d7f7a76fa4a0ae42d6d9f6254b5439e/src/flint/test/test.py#L1831-L1837
The crash happens in `x = g.discrete_log(a) ` for the 1st iteration of the inner loop on the 2nd or 3rd iteration of the outer loop.

I think that this is most likely a memory bug in python-flint. The memory management for `fmpz_mod` and `fmpz_mod_ctx` is suspicious:
https://github.com/flintlib/python-flint/blob/e6c007879d7f7a76fa4a0ae42d6d9f6254b5439e/src/flint/types/fmpz_mod.pxd#L8-L19
I think this is something to do with either L or x_g which will be initialised on a 1st call to `g.discrete_log(a)` for a given `g`.

oscarbenjamin:
Here when an fmpz_mod is created x_g is set to NULL and when it is deallocated a test `if self.x_g` is used:
https://github.com/flintlib/python-flint/blob/e6c007879d7f7a76fa4a0ae42d6d9f6254b5439e/src/flint/types/fmpz_mod.pyx#L236-L244
I think firstly that the test should be `if self.x_g is NULL` although it is not clear if that is the problem.

A call to discrete_log will initialise this:
https://github.com/flintlib/python-flint/blob/e6c007879d7f7a76fa4a0ae42d6d9f6254b5439e/src/flint/types/fmpz_mod.pyx#L360-L376
Again I think `is NULL` should be tested.

We also have:
https://github.com/flintlib/python-flint/blob/e6c007879d7f7a76fa4a0ae42d6d9f6254b5439e/src/flint/types/fmpz_mod.pyx#L140-L151

I think firstly that it would be better to avoid the use of malloc/free wherever possible in favour of using CPython-style memory management.

oscarbenjamin:
This all seems to be working with flint 3.0.0-rc1 now after gh-95.

I've tested locally on Windows that everything seems to be working.

oscarbenjamin:
Rerun after #98 with Flint 3.0.0 final.

oscarbenjamin:
All tests have passed. Previously this PR strangely seemed to flag up errors but I don't see any now.

I think that `fmpz_mod_poly` should be added to the changelog but otherwise this is good for a new release of python-flint.

GiacomoPope:
> I think that fmpz_mod_poly should be added to the changelog but otherwise this is good for a new release of python-flint.

I went to do this just now, but looks like you beat me to it!

oscarbenjamin:
Okay I think this is ready for a 0.5.0 release.

GiacomoPope:
Yeah I agree.

oscarbenjamin:
Okay, let's do it!

oscarbenjamin:
@fredrik-johansson the docs for python-flint have not been updated sine 0.3.0 (5 years ago).

As of 0.5.0 there are some significant new sections in the docs for `fmpz_mod` and `fmpz_mod_poly` as well as changes to the build instructions (although those might actually need updating again).

Ideally docs updates would be handled automatically in CI along with pushing to PyPI but for now at least I am pushing manually to PyPI.

The docs are hosted on your website fredrikj.net though so I don't know if you would prefer to push the updates there yourself manually. Alternatively we could designate a repo somewhere for hosting the built docs that could be pulled automatically by fredrikj.net or something.

fredrik-johansson:
Maybe host the docs on readthedocs or somesuch? We could do something automatic on my server but I'm not too eager to set that up right now.

oscarbenjamin:
I'll have a look at setting up read the docs.

I have *finally* managed to upload the release files to PyPI:

https://pypi.org/project/python-flint/#files

It took 3 hours with a lot restarting twine to get those files up there. Maybe there is a better way to do the uploading.

I need to go do other things now but I'll put out announcements later.

PR #85, opened 2023-09-18
Clean up and include discrete logarithm functions into fmpz_mod

GiacomoPope:
Also some other clean up which should have been done before but I didn't see it.

### DIFF
--- a/src/flint/types/fmpz_mod.pxd
+++ b/src/flint/types/fmpz_mod.pxd
@@ -5,10 +5,9 @@ from flint.flintlib.fmpz_mod cimport fmpz_mod_ctx_t
 
 cdef class fmpz_mod_ctx:
     cdef fmpz_mod_ctx_t val
-
+    cdef any_as_fmpz_mod(self, obj)
+    
 cdef class fmpz_mod(flint_scalar):
     cdef fmpz_mod_ctx ctx
     cdef fmpz_t val
-
-    cdef any_as_fmpz_mod(self, obj)
 
--- a/src/flint/types/fmpz_mod.pxd
+++ b/src/flint/types/fmpz_mod.pxd
@@ -8,13 +8,11 @@ from flint.flintlib.fmpz_mod cimport (
 
 cdef class fmpz_mod_ctx:
     cdef fmpz_mod_ctx_t val
-    cdef fmpz_mod_discrete_log_pohlig_hellman_t L
-    cdef bint _dlog_precomputed
+    cdef fmpz_mod_discrete_log_pohlig_hellman_t *L
     cdef any_as_fmpz_mod(self, obj)
     cdef _precompute_dlog_prime(self)
     
 cdef class fmpz_mod(flint_scalar):
     cdef fmpz_mod_ctx ctx
     cdef fmpz_t val
-    cdef bint base_dlog_precomputed
-    cdef fmpz_t x_g
+    cdef fmpz_t *x_g
--- a/src/flint/types/fmpz_mod.pxd
+++ b/src/flint/types/fmpz_mod.pxd
@@ -1,11 +1,17 @@
 from flint.flint_base.flint_base cimport flint_scalar
 from flint.flintlib.fmpz cimport fmpz_t
-from flint.flintlib.fmpz_mod cimport fmpz_mod_ctx_t
+from flint.flintlib.fmpz_mod cimport (
+    fmpz_mod_ctx_t,
+    fmpz_mod_discrete_log_pohlig_hellman_t
+)
 
 
 cdef class fmpz_mod_ctx:
     cdef fmpz_mod_ctx_t val
+    cdef fmpz_mod_discrete_log_pohlig_hellman_t L
+    cdef bint _dlog_precomputed
     cdef any_as_fmpz_mod(self, obj)
+    cdef _precompute_dlog_prime(self)
     
 cdef class fmpz_mod(flint_scalar):
     cdef fmpz_mod_ctx ctx
--- a/src/flint/types/fmpz_mod.pxd
+++ b/src/flint/types/fmpz_mod.pxd
@@ -16,4 +16,5 @@ cdef class fmpz_mod_ctx:
 cdef class fmpz_mod(flint_scalar):
     cdef fmpz_mod_ctx ctx
     cdef fmpz_t val
-
+    cdef bint base_dlog_precomputed
+    cdef fmpz_t x_g
--- a/src/flint/types/fmpz_mod.pyx
+++ b/src/flint/types/fmpz_mod.pyx
@@ -4,8 +4,16 @@ from flint.flintlib.fmpz cimport (
     fmpz_set,
     fmpz_init,
     fmpz_clear,
-    fmpz_equal
+    fmpz_equal,
+    fmpz_is_probabprime,
+    fmpz_sub,
+    fmpz_mul,
+    fmpz_invmod,
+    fmpz_divexact,
+    fmpz_gcd
 )
+from flint.flintlib.fmpz cimport fmpz_mod as fmpz_type_mod
+
 from flint.flintlib.fmpz_mod cimport *
 
 from flint.utils.typecheck cimport typecheck
@@ -157,6 +165,112 @@ cdef class fmpz_mod(flint_scalar):
         cdef bint res
         res = fmpz_mod_is_one(self.val, self.ctx.val)
         return res == 1
+
+    def inverse(self, check=True):
+        r"""
+        Computes :math:`a^{-1} \pmod N`
+
+        When check=False, the solutions is assumed to exist and Flint will abort on
+        failure. 
+
+            >>> mod_ctx = fmpz_mod_ctx(163)
+            >>> mod_ctx(2).inverse()
+            fmpz_mod(82, 163)
+            >>> mod_ctx(2).inverse(check=False)
+            fmpz_mod(82, 163)
+        """
+        cdef fmpz_mod res
+        res = fmpz_mod.__new__(fmpz_mod)
+        res.ctx = self.ctx
+
+        if check is False:
+            fmpz_mod_inv(res.val, self.val, self.ctx.val)
+            return res
+
+        cdef bint r
+        cdef fmpz one = fmpz.__new__(fmpz)
+        fmpz_one(one.val)
+
+        r = fmpz_mod_divides(
+            res.val, one.val, self.val, self.ctx.val
+        )
+        if r == 0:
+            raise ZeroDivisionError(f"{self} is not invertible modulo {self.ctx.modulus()}")
+
+        return res
+
+    def discrete_log(self, a, check=False):
+        """
+        Solve the discrete logarithm problem, using `self = g` as a base.
+        Assumes a solution, :math:`a = g^x \pmod p` exists.
+        
+        NOTE: Requires that the context modulus is prime.
+
+        TODO: This could instead be initalised as a class from a 
+        given base and the precomputations could be stored to allow 
+        faster computations for many discrete logs with the same base. 
+
+            >>> p = 163
+            >>> F = fmpz_mod_ctx(p)
+            >>> g = F(2)
+            >>> x = 123
+            >>> g.discrete_log(g**123)
+            123
+        """
+        cdef fmpz_mod_discrete_log_pohlig_hellman_t L
+        cdef bint is_prime
+
+        # Ensure that the modulus is prime
+        if check:
+            is_prime = fmpz_is_probabprime(self.ctx.val.n)
+            if not is_prime:
+                raise ValueError("modulus must be prime")
+
+        # Then check the type of the input
+        if typecheck(a, fmpz_mod):
+            if self.ctx != (<fmpz_mod>a).ctx:
+                raise ValueError("moduli must match")
+        else:
+            a = self.any_as_fmpz_mod(a)
+            if a is NotImplemented:
+                raise TypeError
+
+        # Initalise the dlog data, all discrete logs are solved with an internally
+        # chosen base `y`
+        fmpz_mod_discrete_log_pohlig_hellman_init(L)
+        fmpz_mod_discrete_log_pohlig_hellman_precompute_prime(L, self.ctx.val.n)
+
+        # Solve the discrete log for the chosen base and target
+        # g = y^x_g and  a = y^x_a
+        # We want to find x such that a = g^x =>
+        # (y^x_a) = (y^x_g)^x => x = (x_a / x_g) mod (p-1) 
+        cdef fmpz_t x_a
+        cdef fmpz_t x_g
+        fmpz_mod_discrete_log_pohlig_hellman_run(x_g, L, self.val)
+        fmpz_mod_discrete_log_pohlig_hellman_run(x_a, L, (<fmpz_mod>a).val)
+
+        # If x_a, x_g share a common factor divide it out?
+        # TODO: this step seems annoying, but it solved the 
+        # problem when I had even x_a, x_g so there was no
+        # inverse mod (p-1)
+        cdef fmpz_t g
+        fmpz_gcd(g, x_a, x_g)
+        fmpz_divexact(x_g, x_g, g)
+        fmpz_divexact(x_a, x_a, g)
+
+        # Finally, compute output exponent
+        cdef fmpz x
+        x = fmpz.__new__(fmpz)
+
+        # Compute (x_a / x_g) mod (p-1) 
+        fmpz_invmod(x.val, x_g, L.pm1)
+        fmpz_mul(x.val, x.val, x_a)
+        fmpz_type_mod(x.val, x.val, L.pm1)
+
+        # Clear the dlog struct
+        fmpz_mod_discrete_log_pohlig_hellman_clear(L)
+        
+        return x
 
     def __richcmp__(self, other, int op):
         cdef bint res
@@ -315,39 +429,6 @@ cdef class fmpz_mod(flint_scalar):
 
     def __floordiv__(self, other):
         return NotImplemented
-
-    def inverse(self, check=True):
-        r"""
-        Computes :math:`a^{-1} \pmod N`
-
-        When check=False, the solutions is assumed to exist and Flint will abort on
-        failure. 
-
-            >>> mod_ctx = fmpz_mod_ctx(163)
-            >>> mod_ctx(2).inverse()
-            fmpz_mod(82, 163)
-            >>> mod_ctx(2).inverse(check=False)
-            fmpz_mod(82, 163)
-        """
-        cdef fmpz_mod res
-        res = fmpz_mod.__new__(fmpz_mod)
-        res.ctx = self.ctx
-
-        if check is False:
-            fmpz_mod_inv(res.val, self.val, self.ctx.val)
-            return res
-
-        cdef bint r
-        cdef fmpz one = fmpz.__new__(fmpz)
-        fmpz_one(one.val)
-
-        r = fmpz_mod_divides(
-            res.val, one.val, self.val, self.ctx.val
-        )
-        if r == 0:
-            raise ZeroDivisionError(f"{self} is not invertible modulo {self.ctx.modulus()}")
-
-        return res
 
     def __invert__(self):
         return self.inverse()
--- a/src/flint/types/fmpz_mod.pyx
+++ b/src/flint/types/fmpz_mod.pyx
@@ -158,9 +158,11 @@ cdef class fmpz_mod(flint_scalar):
 
     def __cinit__(self):
         fmpz_init(self.val)
+        fmpz_init(self.x_g)
 
     def __dealloc__(self):
         fmpz_clear(self.val)
+        fmpz_clear(self.x_g)
 
     def __init__(self, val, ctx):
         if not typecheck(ctx, fmpz_mod_ctx):
@@ -183,6 +185,9 @@ cdef class fmpz_mod(flint_scalar):
             if val is NotImplemented:
                 raise NotImplementedError
         fmpz_mod_set_fmpz(self.val, (<fmpz>val).val, self.ctx.val)
+
+        # Bool to say whether x_g has been computed before
+        self.base_dlog_precomputed = 0
 
     def is_zero(self):
         """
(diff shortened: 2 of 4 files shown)


GiacomoPope:
This is significantly slower than sage right now...

```py
>>> from flint import fmpz_mod_ctx
>>> import timeit
>>>
>>> F = fmpz_mod_ctx(15126189511369606177)
>>> g = F(5)
>>> x = 8693773790212585448
>>> a = g**x
>>> x == g.discrete_log(a)
True
>>>
>>> timeit.timeit("g.discrete_log(a)", globals={"a":a,"g":g}, number=1000)
8.544251491999603
```

```py
>>> from sage.all import GF, discrete_log
>>> import timeit
>>> 
>>> F = GF(15126189511369606177)
>>> g = F(5)
>>> x = 8693773790212585448
>>> a = g**x
>>> x == discrete_log(a, g)
True
>>>
>>> timeit.timeit("discrete_log(a, g)", globals={"a":a,"g":g,"discrete_log":discrete_log}, number=1000)
3.307495746001223
```

GiacomoPope:
Looking at the arithmetic, it seems sage has something for `a*a` which makes it much faster?

```py
>>> from flint import *
>>> import timeit
>>>
>>> p = 2**255 - 19
>>> F = fmpz_mod_ctx(p)
>>> a = 56453160318328368479028265579673628992621083854069162893669294104711133348344
>>> b = 22540897019420235855175446449184775485001034014122189096905866994063997055779
>>> a = F(a)
>>> b = F(b)
>>> 
>>> timeit.timeit("a*a", globals={"a":a, "b":b}, number=1_000_000)
0.36299371299901395
>>> timeit.timeit("a*b", globals={"a":a, "b":b}, number=1_000_000)
0.3671105850007734
```

```py
>>> from sage.all import *
>>> import timeit
>>>
>>> p = 2**255 - 19
>>> K = Zmod(p)
>>> a = 56453160318328368479028265579673628992621083854069162893669294104711133348344
>>> b = 22540897019420235855175446449184775485001034014122189096905866994063997055779
>>> a = F(a)
>>> b = F(b)
>>>
>>> timeit.timeit("a*a", globals={"a":a, "b":b}, number=1_000_000)
0.27032256000165944
>>> timeit.timeit("a*b", globals={"a":a, "b":b}, number=1_000_000)
0.8113395829968795
```

Which is interesting

GiacomoPope:
Rewriting `any_as_fmpz_mod` helped a lot!

```py
Python 3.11.4 (main, Jul 25 2023, 17:07:07) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
>>> from flint import fmpz_mod_ctx as GF
>>> import timeit
>>> 
>>> p = 2**255 - 19
>>> F = GF(p)
>>> a = 56453160318328368479028265579673628992621083854069162893669294104711133348344
>>> b = 22540897019420235855175446449184775485001034014122189096905866994063997055779
>>> a = F(a)
>>> b = F(b)
>>> 
>>> timeit.timeit("a*a", globals={"a":a, "b":b}, number=1_000_000)
0.22983049499816843
>>> timeit.timeit("a*b", globals={"a":a, "b":b}, number=1_000_000)
0.2311925540016091
```

src/flint/types/fmpz_mod.pyx
```diff
@@ -62,6 +70,30 @@ cdef class fmpz_mod_ctx:
...
         return n
 
+    cdef any_as_fmpz_mod(self, obj):
+        # If `obj` is an `fmpz_mod`, just check moduli
+        # match
+        # TODO: we could allow conversion from one modulus to another?
+        if typecheck(obj, fmpz_mod):
+            if self != (<fmpz_mod>obj).ctx:
+                raise ValueError("moduli must match")
+            return obj
```
oscarbenjamin, line 80: I'm not sure exactly how the `!=` is implemented at the C level here but if there was something to ensure that all contexts for a given modulus are unique in memory then this could be like a pointer comparison.

src/flint/types/fmpz_mod.pyx
```diff
@@ -158,13 +184,123 @@ cdef class fmpz_mod(flint_scalar):
...
+    def discrete_log(self, a, check=False):
+        """
+        Solve the discrete logarithm problem, using `self = g` as a base.
+        Assumes a solution, :math:`a = g^x \pmod p` exists.
+        
+        NOTE: Requires that the context modulus is prime.
+
+        TODO: This could instead be initalised as a class from a 
+        given base and the precomputations could be stored to allow 
+        faster computations for many discrete logs with the same base.
```
oscarbenjamin, line 229: We already have a `.ctx` context object. Maybe this should be attached to the context object?

oscarbenjamin: It could be initialised at first use but stored on the context object after.

I don't know how much memory this uses compared to the size of the context object itself but if the base size of the struct is not large it could be stored inline as part of the ctx object or otherwise the ctx object could just have a pointer.

GiacomoPope: We need to have the first discrete log on some base g, which is an fmpz_mod object, but certain parts of L only depend on the modulus and so yes, this could be on the context for sure.

oscarbenjamin: Oh, I see what you mean: there are two internal log calculations but if we stored the log of `g` we could reduce that to one.

So what you could add would be an object like
```python
log_g = fmpz_mod_log(g, ctx)
log_g(12)
```
I don't know how these things are used in practice and I don't have much sense for what the overhead would be of just always calling the internal log calculation twice (do these functions take a long time?).

GiacomoPope: It's also doing way too much work, as it solves the discrete log for primative roots, but if I had some element of order 2, it would be much much faster to do Pohlig-Hellman in this subgroup.

TL;DR I think we shouldn't worry about optimising this in `python-flint` as it's not something state of the art in Flint itself. E.g. Sage calls a sub-exponential algorithm via Pari for Z/NZ, which is (partially) where all the time saving is here.

GiacomoPope:
Not sure what's causing macOS arm64 to crash, I can't reproduce it locally

oscarbenjamin:
> Not sure what's causing macOS arm64 to crash, I can't reproduce it locally

Do you have an arm64 CPU?

GiacomoPope:
> > Not sure what's causing macOS arm64 to crash, I can't reproduce it locally
> 
> Do you have an arm64 CPU?

No, sorry. What I mean is I don't think it's the python layer but maybe a bug around the discrete log formula running on Arm64, so I can't easily dig into it.

GiacomoPope:
Oh great... it worked this time. That's even more annoying lmao.

oscarbenjamin:
> Oh great... it worked this time. That's even more annoying lmao.

That's what "undefined behaviour" means: you can't even rely on it to crash properly!

src/flint/types/fmpz_mod.pyx
```diff
@@ -158,13 +184,145 @@ cdef class fmpz_mod(flint_scalar):
...
+        print("\t[DEBUG]: Precomputed prime for L")
+
+        # Solve the discrete log for the chosen base and target
+        # g = y^x_g and  a = y^x_a
+        # We want to find x such that a = g^x =>
+        # (y^x_a) = (y^x_g)^x => x = (x_a / x_g) mod (p-1) 
+        cdef fmpz_t x_a
+        cdef fmpz_t x_g
+        # TODO: should this value be stored for efficiency?
+        fmpz_mod_discrete_log_pohlig_hellman_run(x_g, L, self.val)
```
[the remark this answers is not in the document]
GiacomoPope: Hmmm I would have said yes, so I'm confused about why this is working at all :/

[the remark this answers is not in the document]
oscarbenjamin: It all depends what the uninitialised memory looks like (hence undefined behaviour). Probably the main thing is whether flint thinks that the `fmpz_t` is a small (inline) fmpz or a big (pointer) fmpz. If it is a pointer then I guess that it will try to free the memory or something. If it is an inline fmpz then probably its value is just overwritten.

GiacomoPope:
Addressing the above, let's see if it works again :)

GiacomoPope:
I've been experimenting with trying to implement the cache, but I haven't got something that works with cython.

I came up with a naive idea:

```py
cache = {}

class C:
    def __new__(cls, n):
        instance = cache.get(n)
        if instance is not None:
            return instance
        instance = super().__new__(cls)
        cache[n] = instance
        return instance

    def __init__(self, n):
        self.n = n

assert C(1) is C(1)
```

and talking with some people on the CryptoHack Discord, `@hellman` suggested a solution with metaclass:

```py
class CachedMeta(type):
    def __call__(cls, a):
        if a not in cls._cache:
            obj = cls._cache[a] = cls.__new__(cls, a)
            obj.__init__(a)
        return cls._cache[a]

class C(metaclass=CachedMeta):
    _cache = {}
    def __init__(self, a):
        print("init", self, a)
```

Which has a benefit of only calling `__init__` once. But, I can't seem to use `metaclass` in cython, and `__new__` is also not supported for cython extensions. One option I guess would be writing a python `CachedClass` class which we then use as `fmpz_mod_ctx(CachedClass)`, but I also haven't quite got that working myself.

Sagemath has `CachedRepresentation` and `UniqueRepresentation` which are a bit bloated for what we need, but might have the right idea.

GiacomoPope:
Potentially the right thing to do is to use something very similar to the following: 

https://github.com/sagemath/sage/blob/develop/src/sage/misc/fast_methods.pyx

oscarbenjamin:
I would definitely not use metaclasses.

The simple way to do this is just to have a function rather than messing around with different methods.

The naive approach looks okay as well though. You can remove the `__init__` method and do `obj.n = n` in `__new__`.

GiacomoPope:
The issue I had with the naive method was that cython compiler shouted at me for __new__ and told me to use __cinit__, but it didn't feel like I could do the above inside the __cinit__

oscarbenjamin:
> and `__new__` is also not supported for cython extensions.

Oh, right I missed that.

This same issue arises for mpoly contexts. I don't remember how it is handled in that PR.

oscarbenjamin:
Also a `@classmethod` can be used.

GiacomoPope:
A classmethod was my first idea but I wanted to be able to make the class using fmpz_mod_ctx(n)

I guess the easiest method is to do all of the cache with the function so the user never directly calls the class but this seems ugly. 

Will look at the mpoly PR later :)

GiacomoPope:
Another thing to think about is how nmod works differently (just two arguments to nmod) where as fmpz_mod asks for an integer and a context. 

The natural solution to me is to make a pseudo nmod_ctx which is callable in the same way fmpz_mod_ctx is handled so we can have nmod_ctx(x) behave as nmod(x, n)

oscarbenjamin:
> The natural solution to me is to make a pseudo nmod_ctx which is callable in the same way fmpz_mod_ctx is handled so we can have nmod_ctx(x) behave as nmod(x, n)

I see what you mean but I wonder if this is all just something to do at a higher level.

It would probably quite easy to wrap up a lot Flint's functionality just in a relatively thin wrapper like the current interface where everything has the same name as the corresponding Flint types and functions.

It would be nice to have a higher level but we can also define that on top like:
```python
class ZZ_n:
    """Integers mod n."""
    def __init__(self, n):
        self.n = n
        if n < threshold:
            self.dtype = lambda x: nmod(x, n)
        else:
            self.dtype = fmpz_mod_ctx(n)

    def __repr__(self):
        return f'ZZ_n({self.n})'

    def __call__(self, val):
        # This is the only performance sensitive part.
        # Could optimise in Cython
        return self.dtype(val)
```

GiacomoPope:
For the caching of contexts, the PR for fmpz_mpoly (#59) also has some TODO and my feeling is this PR could be merged and at a later date, potentially with input from deinst, we can find a standard method for context caching which we apply for all typoe contexts

oscarbenjamin:
The one thing that I would probably change here is to make the `fmpz_mod_discrete_log_pohlig_hellman` struct an attribute of the `fmpz_mod_ctx` object so that `fmpz_mod_discrete_log_pohlig_hellman_precompute_prime` only needs to be called once per context and not every time `discrete_log` is called. I presume there is a reason that this is a separate function with a struct to be initialised rather than just an internal part of the `_run` function.

There could be a cdef method like `ctx.get_discrete_log_pohlig_hellman()` that returns a pointer to this struct. The `__cinit__` could set this pointer to NULL but the first time the method is called it could use `malloc` (or perhaps `flint_malloc`?).

Alternatively the struct could be inline in the `fmpz_mod_ctx` Cython struct. Looks like the struct only adds 40 bytes over the context itself (although presumably more memory elsewhere after `precompute_prime` is called):
https://github.com/flintlib/flint2/blob/283e514d39f8c8668123ebb981599cc94500e5c4/src/fmpz_mod.h#L153-L161

GiacomoPope:
We could even have as the equality check 

```py
def __eq__(self, other)__:
    return self is other or fmpz_equal(self.val.n, other.val.n)
```

Which could still be hashed by the modulus?

GiacomoPope:
I agree about the point about the dlog context. I'll try and rework this to something which feels sensible.

GiacomoPope:
So (rough) benchmark for various moduli:

```
Jack: python-flint % python3 benchmark.py
Modulus       Addition Time (ns)    Subtraction Time (ns)    Multiplication Time (ns)    Division Time (ns)
----------  --------------------  -----------------------  --------------------------  --------------------
163                          105                      121                         103                   186
2^127 - 1                    110                      111                         110                   843
2^255 - 19                   130                      136                         220                   997
```

If we patch context equality to check equality in memory before anything else:

```py
    def __eq__(self, other):
        if self is other:
            return True
        if typecheck(other, fmpz_mod_ctx):
            return fmpz_equal(self.val.n, (<fmpz_mod_ctx>other).val.n)
        return False
```

```
Modulus       Addition Time (ns)    Subtraction Time (ns)    Multiplication Time (ns)    Division Time (ns)
----------  --------------------  -----------------------  --------------------------  --------------------
163                           85                      112                          86                   181
2^127 - 1                    103                      104                          95                   879
2^255 - 19                   133                      146                         256                  1094
```

GiacomoPope:
Moving precomputation from the `discrete_log` call, to the `ctx` class made my test dlog go from `570us` to `500us`, which seems worth while.

src/flint/types/fmpz_mod.pxd
```diff
@@ -1,14 +1,19 @@
...
+)
 
 
 cdef class fmpz_mod_ctx:
     cdef fmpz_mod_ctx_t val
-
+    cdef fmpz_mod_discrete_log_pohlig_hellman_t L
+    cdef bint _dlog_precomputed
+    cdef any_as_fmpz_mod(self, obj)
+    cdef _precompute_dlog_prime(self)
```
oscarbenjamin, line 14: In memory terms the first element of L is an `fmpz_mod_ctx_t` which would end up being a copy of `val`. If we include the struct `L` inline here then we don't need `val`.

I think it is better though to just have a pointer like
```
cdef fmpz_mod_discrete_log_pohlig_hellman_t *L
```
If the pointer is set to NULL in `__cinit__` then there is no need for a separate `_dlog_computed` variable. The `_precompute_dlog_prime` method can just check for NULL and then call malloc etc.

GiacomoPope:
For the following prime in sagemath we find:

```py
sage: p = 2265864780164119884430588711799732781030118760200689436801123468132597600920497412606647
sage: p.nbits()
291
sage: factor(p-1)
2 * 1607 * 8807 * 12037 * 24533 * 25309 * 34871 * 36697 * 38459 * 38653 * 49531 * 53441 * 67867 * 77479 * 79283 * 81919 * 82493 * 86719 * 88873 * 97967
sage: 
sage: F = GF(p)
sage: g = F(2)
sage: a = F.random_element()
sage: %timeit discrete_log(g, a)
8.82 ms ± 390 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

For the same parameters with flint, we find:

```py
from flint import fmpz_mod_ctx
import time

p = 2265864780164119884430588711799732781030118760200689436801123468132597600920497412606647
F = fmpz_mod_ctx(p)
g = F(2)
x = 1912688597311598499324749152080691796129556087988551556197726085819727650974177072903102
a = g**x

def time_dlog(g, a):
    t0 = time.process_time_ns()
    x = g.discrete_log(a)
    time_ns = time.process_time_ns() - t0
    assert g**x == a
    return f"{time_ns / 1_000_000} ms"

time_dlog(g, a)
time_dlog(g, a)
```
```
'61.17 ms'
'4.864 ms'
```

So for the first call it's much slower, but for repeated calls we actually get something faster, despite using an algorithm with worse complexity.

I have also tried adding `x_g` to the `fmpz_mod` class for precomputing this first dlog, this seems to save about 50% again.

oscarbenjamin:
> So for the first call it's much slower, but for repeated calls we actually get something faster, despite using an algorithm with worse complexity.

You can't really compare `timeit` timings with times measured in any other way. For one thing `timeit` will probably ignore the time of the first run. In general though it will always show a lower time than any other way of measuring the time. You can use `timeit` outside of sage though for comparisons: it is an ipython feature.

GiacomoPope:
Oh!! Very cool, TIL.

So using `ipython` I get:

```py
In [1]: from flint import fmpz_mod_ctx
   ...: import time
   ...: 
   ...: p = 2265864780164119884430588711799732781030118760200689436801123468132597600920497412606647
   ...: F = fmpz_mod_ctx(p)
   ...: g = F(2)
   ...: x = 1912688597311598499324749152080691796129556087988551556197726085819727650974177072903102
   ...: a = g**x

In [2]: %timeit g.discrete_log(a)
3.11 ms ± 41.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

and with the new push which also stores `x_g` we get:

```py
In [2]: from flint import fmpz_mod_ctx
   ...: import time
   ...: 
   ...: p = 2265864780164119884430588711799732781030118760200689436801123468132597600920497412606647
   ...: F = fmpz_mod_ctx(p)
   ...: g = F(2)
   ...: x = 1912688597311598499324749152080691796129556087988551556197726085819727650974177072903102
   ...: a = g**x

In [3]: 

In [3]: %timeit g.discrete_log(a)
1.41 ms ± 11.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

As for "when is this useful", much of the time people only want to solve a single discrete log, so this caching is not so helpful, but there are other cases where a single base might need to be solved for several discrete logs and when (p+1) is not so smooth, the saving by storing `x_g` will be even more useful

src/flint/types/fmpz_mod.pxd
GiacomoPope: I have not addressed this yet. So you think I should set the pointer to `NULL` for `__cinit__` and then when I do the normal discrete log stuff then I want to work with `&self.ctx.L`? 

I guess I can also do something similar for `x_g` with a pointer?

oscarbenjamin: I think it's like `f(L[0])`. You can't use `f(*L)` because that means something else in Python.

GiacomoPope: Ahh got it! The real issue was my `malloc` which was poorly written before.

```py
self.L = <fmpz_mod_discrete_log_pohlig_hellman_t *>libc.stdlib.malloc(1 * cython.sizeof(fmpz_mod_discrete_log_pohlig_hellman_struct))
```

Seems to work

GiacomoPope: (There was no `flint_malloc` which I could see, so I did it this way, but we could maybe write a small function to do this?)

GiacomoPope:
Just to show the limitation of the discrete log being exponential in space/time:

```py
In [1]: from flint import fmpz_mod_ctx
   ...: 
   ...: p = 2079343919
   ...: F = fmpz_mod_ctx(p)
   ...: g = F(7)
   ...: a = F(1507522516)
   ...: x = 191604335
   ...: assert a == g**x
   ...: assert x == g.discrete_log(a)
   ...: %timeit g.discrete_log(a)

395 ms ± 11.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```py
sage: q = 2079343919
sage: factor(q-1)
2 * 1039671959
sage: F = GF(q)
sage: g = F(7)
sage: a = F.random_element()
sage: a
1507522516
sage: pari.znlog(a, g)
191604335
sage: %timeit pari.znlog(a, g)
3.74 ms ± 75.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

GiacomoPope:
I don't understand the latest error (or how I could have introduced it, at least)

```

Run pip install --find-links wheelhouse python_flint
Looking in links: wheelhouse
Collecting python_flint
  Obtaining dependency information for python_flint from https://files.pythonhosted.org/packages/c2/58/9246c9daf2d6746bca3b3d57675d8ae44ef4eabcab143257688a2b0e6c6e/python_flint-0.4.4-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata
  Downloading python_flint-0.4.4-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (4.7 kB)
Downloading python_flint-0.4.4-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (35.1 MB)
   ━━━━━━━━                                 7.3/35.1 MB 3.6 MB/s eta 0:00:08
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    unknown package:
        Expected sha256 9dfbc3639fbf695b15e5435ceafcafc1bff46ce9be1446e69e5093da292042a9
             Got        d2f667587a046cfabf46d8bd1032910f8be1ac2fdc764af7d7f9c9ed53339e27


Notice:  A new release of pip is available: 23.0.1 -> 23.2.1
Notice:  To update, run: pip install --upgrade pip
Error: Process completed with exit code 1.
```

oscarbenjamin:
Looks like a network error. I've restarted it.

GiacomoPope:
Happy with this now, but also happy to continue work if you have more ideas. I could probably re-read it all and try and speed things up a little more, but now the ctx comparison is faster there's not a whole bunch of obvious slow down imo

oscarbenjamin:
Okay, looks good to me!

PR #88, opened 2023-09-23
Added fmpq_set_any_ref

deinst:
For implementing `fmpq_mpoly` a `fmpq_set_any_ref` function would be convenient and useful.  Also it makes `fmpq` work similiarly to `fmpz`, `arb`, and `acb`.

There is some ugliness (I inject a c typedef into the code) to get around the fact that what flint calls a `fmpq`, python flint calls a `fmpq_struct`.

### DIFF
--- a/src/flint/types/fmpq.pxd
+++ b/src/flint/types/fmpq.pxd
@@ -1,9 +1,7 @@
 from flint.flint_base.flint_base cimport flint_scalar
-
 from flint.flintlib.fmpq cimport fmpq_t
 
+cdef int fmpq_set_any_ref(fmpq_t x, obj)
 cdef any_as_fmpq(obj)
 cdef class fmpq(flint_scalar):
     cdef fmpq_t val
-
-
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
Looks good.

oscarbenjamin:
> what flint calls a `fmpq`, python flint calls a `fmpq_struct`

Is the problem that in C the struct is `fmpq` and `fmpq_t` is the `fmpq[1]` but then python-flint wants to use `fmpq` as the name for its Python type?

So it would be ambiguous in the Cython code whether `fmpq` means the C struct or the Cython class?

I wonder if there is any better way around that in general...

Does the same also apply to fmpz?

deinst:
It seems to work, and avoids name collisions the way it is now written, but cython is a mystery to me.

It does not apply to fmpz, because fmpz is not a struct, or can at least be treated like something typedefed as a `slong`.

As I understand it the python object's name is mangled in the c code so that the collision only happens at the cython leve so there should be no collision.

PR #87, opened 2023-09-20
Begin adding fmpz_mod_poly

GiacomoPope:
Opening a PR just to have the ability for group discussion while working on this new type. Only work so far is an import of the level0 files, and a simple print test to ensure everything compiles happily.

PR #90, opened 2023-09-29
Add generic tests for polynomial types

oscarbenjamin:
Adds a test_polys function that runs the same same generic tests over `fmpz_poly`, `fmpq_poly` and `nmod_poly` to share tests and ensure that the interfaces of the three types are consistent where appropriate. Some methods that were present on some types but not on others were added.

### DIFF
--- a/src/flint/test/test.py
+++ b/src/flint/test/test.py
@@ -742,9 +742,9 @@ def test_fmpq():
     assert raises(lambda: Q("1.0"), ValueError)
     assert raises(lambda: Q("1.5"), ValueError)
     assert raises(lambda: Q("1/2/3"), ValueError)
-    assert raises(lambda: Q([]), ValueError)
-    assert raises(lambda: Q(1, []), ValueError)
-    assert raises(lambda: Q([], 1), ValueError)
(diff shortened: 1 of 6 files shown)


src/flint/test/test.py
```diff
@@ -1815,6 +1814,204 @@ def test_fmpz_mod_dlog():
...
+        (lambda *a: flint.nmod_poly(*a, 17), lambda x: flint.nmod(x, 17), True),
+    ]
+
+
+def test_polys():
+    for P, S, is_field in _all_polys():
+
+        assert P([S(1)]) == P([1]) == P(P([1])) == P(1)
+
+        assert raises(lambda: P([None]), TypeError)
```
oscarbenjamin, line 1831: @GiacomoPope if we write the tests more like this then we don't need to duplicate the test code for different types and it also means that we can test that the different types have consistent method names and signatures.

GiacomoPope: Yeah I agree with this. I think once #87 is merged I will take the time to address the tests before working on `fq` types.

Issue #91, opened 2023-10-01
Help understanding whether to implement `fq`, `fq_default` or both


GiacomoPope:
Next on my list after fmpz_mod_poly is to try and implement finite fields. I see in the 3.0 documentation there are two related types with the *_default options having less methods (I think)

For python-flint should I only focus on fq? What's the use of the default type?

oscarbenjamin:
The doc for `fq_default_ctx_init_type` shows what this is for I think:

> As per the previous function except that if type == 1 an fq_zech context is created, if type == 2 an fq_nmod and if type == 3 an fq. If type == 0 the functionality is as per the previous function.

For `fq_default_ctx_init` the doc says:

> Initialises the context for prime and extension degree, with name var for the generator. By default, it will try use a Conway polynomial; if one is not available, a random irreducible polynomial will be used.

So the purpose is to provide a generic interface over the three representations `fq`, `fq_nmod` and `fq_zech`. The doc for `fq_default_ctx_init` does not explicitly say this but the intention is that it will make a good choice for which of the three representations will be used. The code is here:
https://github.com/flintlib/flint2/blob/72ff1553903198ad34830db12d1a8462c52a7e33/src/fq_default/ctx_init_modulus.c#L14

So if you want to construct a context for a finite field of order $q = p^k$ then by default it will use:

- `fq_zech` if `k > 1` and `q <= 65536`
- `fq_nmod` if `k > 1` and `p` is word-sized.
- `fq` if `k > 1` and `p` is not word-sized.
- `nmod` if `k == 1` and `p` is word-sized.
- `fmpz_mod` if `k == 1` and `p` is not word-sized.

Then all other `fq_default` functions are just wrappers that call through to the appropriate routine for the actual type:
https://github.com/flintlib/flint2/blob/72ff1553903198ad34830db12d1a8462c52a7e33/src/fq_default.h#L348-L372

Probably for python-flint it makes most sense to wrap `fq_default` rather than any of the others. The overhead of the `fq_default` wrapper is unlikely to be significant compared to the overhead of the Cython wrapper and the Python interpreter.

Any functions that are provided by specific types like `fq` or `fq_nmod` could also be obtained in python-flint by performing some conversions.

[...]

PR #97, opened 2023-10-19
Include the `fq_default` type

GiacomoPope:
I did this work a few weeks ago, then dropped it because I got married and now I'm trying to implement other work for a paper so my brain is VERY outside of python-flint world but thought I'd show at least roughly what's happening for this type and what we'll need. 

Main thing is that we have these "union structs" for the `fq_default` type which aren't implemented in cython, so we need to do nesting instead.


oscarbenjamin:
Note also previous discussion about this in gh-91.

oscarbenjamin:
The way that you have defined the `fq_default*` structs in Cython here seems reasonable to me. I'm actually unsure if it is necessary to define anything about these structs at all if only `fq_default_*` functions are being used. Maybe these structs can just be treated as opaque at the Cython level. If there is any need to convert them then maybe functions like `fq_default_get_fmpz_mod_poly` are sufficient.

I think it is only necessary to define the struct members if we want to use direct member access but if the `fq_default*` functions are used instead then it is probably fine (possibly safer?) to just leave the struct definitions empty.

GiacomoPope:
After I've wrapped up a few other projects, I'll come back to this then and at least get the `fq_default` doing some standard stuff. Then we can maybe start looking at adding the generic types in

oscarbenjamin:
It's always worth remembering that it is most important just to get the basic features in. Trying to add all the possible functions that finite fields can support can take a lot longer than just making it possible to construct the objects in the first place and do some basic arithmetic.

Also once a type is added it becomes a lot easier for other people in future to add specific features if they want them.

GiacomoPope:
The topic of this work came up at a Sage days I am currently attending and with some of my research code finished I now have more time to start working more on this again. Latest commit was simply merging master back into this branch to be up to date with the work I've missed.

GiacomoPope:
@oscarbenjamin I'm being stupid with the local build...

If I run:

```
source bin/activate
./bin/build_inplace.sh   
```

I get an error from the current setup:

```
  File "/Users/Jack/Documents/GitHub/python-flint/setup.py", line 143, in <module>
    ext_modules=cythonize(ext_modules, compiler_directives=compiler_directives),
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Jack/Library/Python/3.12/lib/python/site-packages/Cython/Build/Dependencies.py", line 1010, in cythonize
    module_list, module_metadata = create_extension_list(
                                   ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Jack/Library/Python/3.12/lib/python/site-packages/Cython/Build/Dependencies.py", line 845, in create_extension_list
    for file in nonempty(sorted(extended_iglob(filepattern)), "'%s' doesn't match any files" % filepattern):
  File "/Users/Jack/Library/Python/3.12/lib/python/site-packages/Cython/Build/Dependencies.py", line 117, in nonempty
    raise ValueError(error_msg)
ValueError: 'src/flint/types/fmpz_mpoly_q.pyx' doesn't match any files
```

If I install the package

```
pip install .
```

Everything seems to work, but I have obviously got something wrong with the build as I get:

```
Python 3.11.9 (main, Apr  2 2024, 08:25:04) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> from flint import fmpz
  File "/Users/Jack/Library/Python/3.11/lib/python/site-packages/flint/__init__.py", line 27, in <module>
    from .types.fq_default import *
ModuleNotFoundError: No module named 'flint.types.fq_default'
```

Sorry for such a basic question... my usual way of working doesnt seem to match since some recent changes

oscarbenjamin:
Yes, it has changed because of the meson stuff which I need to properly document somewhere.

The setup.py is no longer needed but is just there for compatibility. The build_inplace script uses setup.py so should not be used any more. In meson there is no concept of an in-place build because all builds are always out-of-tree in a build directory.

There are several ways of doing this now:

First way:

Use something like `pip` as a build frontend which will invoke the `meson-python` build backend (see PEP 517). That means that e.g. `pip install .` will work.

You can also use this to create a meson-python editable install like:
```
pip install meson meson-python cython
pip install --no-build-isolation --editable .
```
This editable install will be built in the `./build` directory. Every time you `import flint` from Python it will trigger an incremental rebuild.

Second way:

Don't bother with `pip` and the whole PEP 517 business. Run meson commands directly to build things instead. First get a clean setup:
```
pip uninstall python-flint
rm -r build build-install
```
Now think of these commands as `./configure`, `make` and then `make install`:
```
meson setup build --pkg-config-path=$(pwd)/.local/lib/pkgconfig
meson compile -C build
meson install -C build --destdir ../build-install
```
Now the `build-install` directory looks like this:
```console
$ tree build-install/
build-install/
└── usr
    └── local
        └── lib
            └── python3.11
                └── site-packages
                    └── flint
                        ├── flint_base
                        │   ├── flint_base.cpython-311-x86_64-linux-gnu.so
                        │   ├── flint_context.cpython-311-x86_64-linux-gnu.so
                        │   ├── __init__.py
                        │   └── __pycache__
                        │       └── __init__.cpython-311.pyc
...
```
Add that directory to PYTHONPATH and run the tests:
```console
export PYTHONPATH=$(pwd)/build-install/usr/local/lib/python3.11/site-packages
export LD_LIBRARY_PATH=$(pwd)/.local/lib
python -m flint.test
```

Third way:

Running the meson commands directly is tedious so spin (`pip install spin`) acts as a frontend to make this nicer. Make sure you have a clean setup (no editable install, etc) first:
```
pip uninstall python-flint
rm -r build build-install
```
Now `spin` can build and run the tests for you. If you have flint installed system wide (rather than in a local directory) then it should be as simple as running `spin test` which will build everything, install it into `build-install`, add to `PYTHONPATH` and run all the tests. This builds in parallel and uses incremental rebuilds so it's a lot faster.

I don't have Flint installed system wide so I need to configure the build first like this:
```
spin build -- --pkg-config-path=$(pwd)/.local/lib/pkgconfig -Dadd_flint_rpath=true
```
That configuration is persistent so now I can just run e.g. `spin test` to rebuild and run tests.

Fourth way:

It is also possible with `spin` to use an editable install and `spin install` will set that up for you by running `pip`. Note that `spin` is just a convenience frontend. It prints out all of the commands that it executes in case you want to run them yourself with different arguments or something.

Note that these different ways are not compatible with each other so you have to choose one. You probably need a deep clean when switching from one way to another because they use the build directory in incompatible ways which is why I recommend:
```
pip uninstall python-flint
rm -r build build-install
```

I'm not sure what should be the recommended way right now. Probably using `spin` without an editable install should be recommended although the editable install is quite convenient. I have spent a long time learning how all the pieces work and so I find all of these things easy to use now but I'm not sure what others would prefer. I think @Jake-Moss prefers to use `spin` without an editable install.

With meson all building is out of tree so if you have an old checkout where you have been doing in-place builds then it is probably best to clean out all the old `.c` etc files with `git clean`. The meson build will not generate those files in the source tree any more.

GiacomoPope:
AMAZING! Thanks so much for such a detailed answer :)

I'll work at finally getting this type done this evening

GiacomoPope:
@oscarbenjamin this is still very much unfinished, but I wanted to point to the arithmetic boilerplate code in the `flint_scalar` class -- is this something valuable? I thought it might be worth having another PR / issue where we discuss some generic functions which we think *should* be included for all these base classes as a guide for people implementing new types?

This links again to what you said in the other PR with `pow_trunc` -- having expectations with boilerplates in the base class could help?

oscarbenjamin:
> I wanted to point to the arithmetic boilerplate code in the `flint_scalar` class -- is this something valuable?

Yes, I think so.

I think that for some very basic types like `nmod`, `fmpz`, etc it is important for the subclass to implement most/all of its own boiler-plate for maximum speed when e.g. adding `nmod + nmod` or `fmpz + int`. For most other types like polynomials, matrices etc it matters much less because each operation is generally more expensive. Cython's subclass v-tables are a fast dispatch mechanism in my experience so the overhead of splitting the implementation of a method between super and subclass is quite small for cdef classes.

In general though I think it would be better if most types could inherit something that manages most of the `__add__` etc boiler-plate like checking compatible types, coercing `int` etc. As you say it would help with defining what the expected methods should be as well.

> I thought it might be worth having another PR / issue where we discuss some generic functions which we think _should_ be included for all these base classes as a guide for people implementing new types?

I think that makes sense.

Jake-Moss:
> I think @Jake-Moss prefers to use spin without an editable install.

Yep that's correct! IIRC the editable install doesn't support prompting a compilation from anything other than importing. This didn't jell with me, I prefer my in-editor write, compile, test workflow with an explicit (or automatic) compilation step. 

However, not using an editable install means all my `python-flint` related tools need to be aware of spin to properly setup `PYTHON_PATH`, so I configure my emacs setup with the Python executable name as `spin` with the necessary options. This means I can have an explicit compilation step (with my choice of arguments) if I desire it via `spin build`, and any invocation of spin prompts an incremental build. So things like launching `pytest` or a python repl first call `spin`, building if necessary, then running as normal. 

IMO it's pretty close to optimal, only thing left for me to tinker with is to write some `spin` layout for `direnv` and drop the emacs specific configuration, but prefixing things with `spin` works well enough for now.

oscarbenjamin:
> IMO it's pretty close to optimal

Okay, so maybe I'll just write some docs that explain how to use spin like that. I can give some context for how to set things up the other ways but it is good to have a guide that clearly describes one way of doing things.

Another thing I'm not sure about:

Do either of you build against a system install of Flint or use a local build?

I always use a local build in `$(pwd)/.local` because I want to have local control over which version of Flint is being used. The `bin/build_dependencies_unix.sh` script sets up the local build of GMP, MPFR and Flint to make this work. Do you use that script?

Using a local build complicates things a bit because you need to fiddle with environment variables. Hence the `bin/activate` script:
https://github.com/flintlib/python-flint/blob/069d24d5d7a85bb8e264359ca9f3fcc8e744921d/bin/activate#L1-L3

With spin/meson there is a better way than using environment variables but it just complicates exactly the instructions that are needed for setting up the development build. Firstly meson uses pkgconfig to locate dependencies like Flint and so rather than setting `LIBRARY_PATH` or `C_INCLUDE_PATH` you set `PKG_CONFIG_PATH` which you can do with the `--pkg-config-path` command line argument. That needs to be done in the call to `meson setup` or `spin build`. With `spin` this means that you need to have a first explicit call to `spin build` before you can then use e.g. `spin test` because arguments like `--pkg-config-path` cannot be passed to `spin test` (its arguments are all passed to `pytest` rather than to `meson`).

Setting `--pkg-config-path` means that the build completes but it is also necessary to be able to find the `libflint` shared library at runtime. That is what `LD_LIBRARY_PATH` is for on Linux (or `DYLD_LIBRARY_PATH` on macos or `PATH` on Windows). At least on Linux or macos there is a better way than using environment variables which is to add an rpath entry. The meson-python editable install does this automatically. When using `spin` it happens automatically on macos but not on Linux. I've added a meson configuration option called `add_flint_rpath` for this. Possibly it makes sense to have this option on by default somehow but this is needed as well.

This is all why for me the first command I need to run is either of:
```
meson setup build --pkg-config-path=$(pwd)/.local/lib/pkgconfig -Dadd_flint_rpath=true
spin build -- --pkg-config-path=$(pwd)/.local/lib/pkgconfig -Dadd_flint_rpath=true
```
It might be handy to put this into a convenient script like `bin/setup_spin.sh` or something so then the instructions to set up a dev checkout are:
```
git clone git@github.com:flintlib/python-flint.git
cd python-flint
pip install -r requirements-dev.txt
bin/build_dependencies_unix.sh
bin/setup_spin.sh
```
And then `spin test`, `spin docs`, `spin sdist` etc should work. I think one thing missing right now is `spin doctest`.

I'm not sure though if I can assume though that:

1. Everyone wants to use a local build of Flint when working on python-flint.
2. The path `$(pwd)/.local/lib/pkgconfig` is always correct. It might be `lib64` or something on some systems.
3. It is always fine to add the rpath entry. I think at least if this is just advice for a dev checkout then probably it is fine. There are apparently reasons why meson does not do this by default (https://github.com/mesonbuild/meson/issues/13046).

GiacomoPope:
@oscarbenjamin small issue with doc tests -- the default type picked between 3.0 and 3.1 seems to be different, which means I can't get things to pass for both...

oscarbenjamin:
> the default type picked between 3.0 and 3.1 seems to be different, which means I can't get things to pass for both...

Could just mark the doctest to be skipped.

> One option would be simply to not have the context type included on the `__repr__`

It might be nicer if it displayed as something other than just an integer. The multivariate polynomials use an enum for monomial orderings. Perhaps something similar could be used here:
```python
In [1]: import flint

In [2]: ctx = flint.fmpz_mpoly_ctx(3, flint.Ordering.lex, ["x", "y", "z"])

In [3]: ctx
Out[3]: fmpz_mpoly_ctx(3, '<Ordering.lex: 0>', ('x', 'y', 'z'))
```
I might prefer that to just show as this and for this to be valid though:
```python
fmpz_mpoly_ctx(3, 'lex', ('x', 'y', 'z'))
```
Using integers makes sense in C but in Python strings are nicer. Enums are also good but it is nice if you can shortcut them with something like `'lex'` that is still meaningful.

Jake-Moss:
> Do either of you build against a system install of Flint or use a local build?

I've been using something pretty similar with a local build, just in a different location. I've controlled everything via a `.envrc` where I set `PKG_CONFIG_PATH`, `LIBRARY_PATH`, `LD_LIBRARY_PATH`, and `C_INCLUDE_PATH` appropriately. Very recently I've switched to using a Nix flake to manage everything where I can override various options without building from source manually.  Having `pkg-config` with nix is enough that I don't need to configure anything else.

> 2. The path $(pwd)/.local/lib/pkgconfig is always correct. It might be lib64 or something on some systems.

My certainly uses (or rather used) `lib64`.

> 3. It is always fine to add the rpath entry. I think at least if this is just advice for a dev checkout then probably it is fine. There are apparently reasons why meson does not do this by default (https://github.com/mesonbuild/meson/issues/13046).

After having read that issue I think it's _probably fine_ for a quick automated install script to add the rpath entry if they have not installed `flint` via a package manager. It's always possible for contributors to do things themselves if anything goes wrong.

Jake-Moss:
> It might be nicer if it displayed as something other than just an integer. The multivariate polynomials use an enum for monomial orderings. Perhaps something similar could be used here:
> 
> ```python
> In [1]: import flint
> 
> In [2]: ctx = flint.fmpz_mpoly_ctx(3, flint.Ordering.lex, ["x", "y", "z"])
> 
> In [3]: ctx
> Out[3]: fmpz_mpoly_ctx(3, '<Ordering.lex: 0>', ('x', 'y', 'z'))
> ```
> 
> I might prefer that to just show as this and for this to be valid though:
> 
> ```python
> fmpz_mpoly_ctx(3, 'lex', ('x', 'y', 'z'))
> ```
> 
> Using integers makes sense in C but in Python strings are nicer. Enums are also good but it is nice if you can shortcut them with something like `'lex'` that is still meaningful.

I should be able to fix that by using a Python based `enum.Enum` class instead of the `cpdef enum` as we can then specify strings as the constants. Not sure why I opted for the cython-based class over the Python one in the first place. I've been slacking on https://github.com/flintlib/python-flint/pull/164, I can make the change after that.

oscarbenjamin:
> For coverage, how best should I check this? `coverage run ?`

I'm not sure. This is one thing that does not currently work with meson because Cython gets confused about the out of tree build. I showed (in https://github.com/cython/cython/issues/6186#issuecomment-2104842388) how to patch Cython and then add something to python-flint's coverage config that would make it work I think.

Jake have you managed to measure coverage with the meson build somehow?

oscarbenjamin:
> I should be able to fix that by using a Python based `enum.Enum` class instead of the `cpdef enum` as we can then specify strings as the constants.

I don't think it needs the Python enum. Is it not just a case of checking for a string in the constructor and then also displaying a string in the repr?

GiacomoPope:
```py
In [18]: fq_default_ctx(5, 2, fq_type=1)
Out[18]: fq_default_ctx(5, 2, 'x', x^2 + 4*x + 2, 'FQ_ZECH')

In [19]: fq_default_ctx(5, 2, fq_type=2)
Out[19]: fq_default_ctx(5, 2, 'x', x^2 + 4*x + 2, 'FQ_NMOD')

In [20]: fq_default_ctx(5, 2, fq_type=3)
Out[20]: fq_default_ctx(5, 2, 'x', x^2 + 4*x + 2, 'FQ')
```

I have this set up, where I convert from a string if necessary in the `__init__` and print as a string in the `__repr__`

oscarbenjamin:
> After having read that issue I think it's _probably fine_ for a quick automated install script to add the rpath entry if they have not installed `flint` via a package manager.

I'm not sure that we can check how Flint was installed. I think that's what the discussion in the linked issue came down to i.e. that it is not easy to figure out if the rpath is needed or not.

Currently there is a project option `add_flint_rpath` and there are a number of different ways to set it:
```
meson setup build --pkg-config-path=$(pwd)/.local/lib/pkgconfig -Dadd_flint_rpath=true
spin build -- --pkg-config-path=$(pwd)/.local/lib/pkgconfig -Dadd_flint_rpath=true
pip install . --config-settings=setup-args=--pkg-config-path=$(pwd)/../sympy/.local/lib/pkgconfig --config-settings=setup-args=-Dadd_flint_rpath=true
```
I can just list those commands in the instructions somewhere and let people figure it out. There's no way round having someone specify the pkgconfig location. I think for `pip install --editable .` or `spin install` the rpath gets set anyway because meson-python has a fixer that sets it but only in an editable install.

I suspect that this is only ever needed for development work anyway because otherwise Flint will be installed system-wide or there will be some other system for making it available like with environment variables. For the binary wheels there is no issue because the cibuildwheel repair step bundles all the libs and fixes all the rpaths.

GiacomoPope:
I'm sure this is not perfect, but it's a good start at exposing most of the functionality of the `fq_default` type. Ready for review

Jake-Moss:
> Jake have you managed to measure coverage with the meson build somehow?

No I have not, I believe it blocked by the linked issue. If your patch is working I could automate the patching of cython via nix but I don't believe that is a nice solution.
To get my coverage reports I have a separate copy of `python-flint` that I build with `setup.py`. 

> I can just list those commands in the instructions somewhere and let people figure it out.

Sorry this is what I meant. You can only do so much, I think documenting it is enough

> > I should be able to fix that by using a Python based `enum.Enum` class instead of the `cpdef enum` as we can then specify strings as the constants.
> 
> I don't think it needs the Python enum. Is it not just a case of checking for a string in the constructor and then also displaying a string in the repr?

We certainly can, I believe I added it here https://github.com/flintlib/python-flint/pull/132/commits/17b3406758855ee3028c7f433c4d86442f872551 because I just wanted to move the ordering specific stuff out and away from the context classes. It would be very easy to just revert to using strings

GiacomoPope:
The main thing missing now is probably thorough coverage of all the branching for context creation... I'm not sure if I'm even 100% happy with the current context generation but it's a start

src/flint/test/test_all.py
```diff
@@ -3585,6 +3585,74 @@ def test_matrices_transpose():
...
+        assert gf.one().is_one() is True
+
+        a = gf.random_element(not_zero=True)
+        b = gf.random_element(not_zero=True)
+        c = gf.random_element(not_zero=True)
+
+        assert a + (-a) == gf.zero()
+        assert a + a == 2*a
+        assert a * a == a**2
+        assert a * a == a.square()
```
oscarbenjamin, line 3630: It would be good to add some careful testing that `__eq__` behaves as expected. These tests only work if `__eq__` works.

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,790 @@
...
+        # For small integers we can convert directly
+        if typecheck(obj, int) and obj.bit_length() < 32:
+            fq_default_set_si(fq_ele, <slong>obj, self.val)
+            return 0
+
+        # Assumes that the modulus of the polynomial matches
+        # the context for the fq_default
+        if typecheck(obj, fmpz_mod_poly):
+            fq_default_set_fmpz_mod_poly(fq_ele, (<fmpz_mod_poly>obj).val, self.val)
+            return 0
```
oscarbenjamin, line 340: If the modulus doesn't match then what happens? Does it just set without checking or does it crash or something?

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,790 @@
...
+
+        # Now set the value from the fmpz_poly
+        fq_default_set_fmpz_poly(fq_ele, poly.val, self.val)
+
+        return 0
+
+    cdef set_any_as_fq_default(self, fq_default_t fq_ele, obj):
+        # Converts the list to an fmpz_poly and then sets from this
+        if typecheck(obj, list):
+            return self.set_list_as_fq_default(fq_ele, obj)
```
oscarbenjamin, line 329: I think it is a bit dangerous to accept a list here if this function is used as part of `__add__` etc.

oscarbenjamin: It is even used in `__eq__`:
```python
In [4]: ctx = flint.fq_default_ctx(5, 2, 'y')

In [5]: y = ctx.gen()

In [6]: y
Out[6]: y

In [7]: y == [0, 1]
Out[7]: True
```

src/flint/types/fq_default.pyx
GiacomoPope: I think it simply takes the polynomial and ignores the modulus reducing modulo the fq context instead

src/flint/test/test_all.py
GiacomoPope: this has been added

src/flint/types/fq_default.pyx
GiacomoPope: I'd like to be able to do `GF([1,2,3])` though, as elements of the extensions are essentially polynomials

GiacomoPope: but maybe we need a different call for `__call__` on the context and on the arithmetic type casting

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,790 @@
...
+        fq_default_get_fmpz_mod_poly((<fmpz_mod_poly>pol).val, self.val, self.ctx.val)
+
+        return pol
+
+    def str(self):
+        return self.polynomial().str(var=self.ctx.var.decode())
+
+    def __repr__(self):
+        # TODO: what do we want here?
+        return str(self)
```
oscarbenjamin, line 478: I'm not sure but maybe it is better to define `.repr()` here and have the superclass figure out which one to call based on the pretty setting. I think that is done for some other types if I remember correctly.

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,790 @@
...
+
+    def __richcmp__(self, other, int op):
+        cdef bint res
+        if op != 2 and op != 3:
+            raise TypeError("fq_default cannot be ordered")
+
+        if not typecheck(other, fq_default):
+            other = self.ctx.any_as_fq_default(other)
+
+        if typecheck(other, fq_default):
```
oscarbenjamin, line 515: I'm not sure about using `any_as` in the `__richcmp__` context any more. I recently added something like this to `nmod_poly` and then afterwards thought that there must be a better way.

Correct me if I am wrong here but if  `F = fq(p^n)` then an element `x` of `F` is basically a polynomial over `G = fq(p)`. The only way that `x` is equal to anything that is not an element of `F` is if its polynomial is constant so that we can equate it to an element of `fq(p)` (or `Z/(p)`).

Is that correct?

If that is correct then I think it makes more sense rather than converting `other` to an `fq` we should check if self is a constant polynomial and if so downgrade `self` to an `fq(p)` or an `fmpz_mod(p)` or something for the comparison.

Or does it get more complicated than that? Are there conversions between e.g. `fq(p^d)` and `fq(p^e)`? Would there be a reason to equate elements of fields with same characteristic but different order?

We have to be careful here because `any_as_fq_default` is quite greedy but I think we want to be strict with `__eq__` in general. I'm not sure exactly what makes sense when it comes to equating finite field elements with other things that are not finite field elements or that are elements of different finite fields.

Does it make sense to seamlessly mix `nmod` and `fmpz_mod` in with `fq` types?

Also note this:
```python
In [8]: from flint import *

In [9]: two = nmod(2, 5)

In [10]: two == 2
Out[10]: True

In [11]: two == fmpz(2)
Out[11]: False

In [12]: fmpz(2) == fmpq(2)
Out[12]: True
```
I'm not entirely sure what is best here but I had the impression that this was an intentional decision that although `int` can mix in with nmod the flint types are clearly separated for zero vs nonzero characteristic. I don't know how that should extend to the case of mixing `nmod` etc with finite fields.

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,790 @@
...
+    def _neg_(self):
+        cdef fq_default res
+        res = self.ctx.new_ctype_fq_default()
+        fq_default_neg(res.val, self.val, self.ctx.val)
+        return res
+
+    def _add_(self, other):
+        other = self.ctx.any_as_fq_default(other)
+        if other is NotImplemented:
+            return NotImplemented
```
oscarbenjamin, line 538: One of the things that would be nice about having the `__add__` etc boilerplate in the superclass is if we could centralise the logic around how coercions happen. Ideally if we have the overhead of an `__add__` method that calls an `_add_` method then the `__add__` method should do something to simplify the situation like handling NotImplemented etc. It would be like:
```python
class Super:
    def __add__(self, other):
        # Happy fast path:
        if self.typecheck(other):
            return self._add_(other)
        elif typecheck(other, int):
            return self._add_int_(other)
        other = self.any_as_self(other)
        if other is NotImplemented:
            return NotImplemented
        return self._add_(other)

class Sub:
    def any_as_self(self, other):
        ...

    def _add_(Sub self, Sub other):
        # No type checking needed here
        # No NotImplemented etc. Types must be correct.
        r = Sub.__new__(Sub)
        fmpsub_add(r.val, self.val, other.val)
        return r
```
For some types like fmpz this would be too much overhead and we would still need to have `__add__` implemented fully with all boilerplate in th subclass. The intention though should be that the superclass is somehow simplifying the situation to make it easier for the subclass to implement these methods. Otherwise there isn't any point in having `__add__` call `_add_`.

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,790 @@
...
+
+        cdef fq_default res
+        res = self.ctx.new_ctype_fq_default()
+        fq_default_mul(res.val, self.val, (<fq_default>other).val, self.ctx.val)
+        return res
+
+    @staticmethod
+    def _div_(left, right):
+        # Case when left and right are already fq_default
```
oscarbenjamin, line 582: Having `__rdiv__` call here reversing the order of the arguments places us in a somewhat awkward position where we don't know which argument is actually an `fq_default`.

src/flint/types/fq_default.pyx
oscarbenjamin: I think that `__call__` is definitely a different case. Implicit coercions need to be limited especially if they are used in things like `__eq__`.

When someone calls `GF(obj)` that is an explicit conversion which is a very different situation.

src/flint/types/fq_default.pyx
GiacomoPope: I have tried to do something similar to this in the latest commit, although without the optimisation of things like `self._add_int_(other)` which I think we can do at a later date?

src/flint/types/fq_default.pyx
GiacomoPope: OK this has been fixed:

```py
>>> from flint import fq_default_ctx
>>> gf = fq_default_ctx(5)
>>> gf(0) == [0]
False
>>> gf.gen() == [0, 1]
False
```

src/flint/types/fq_default.pyx
GiacomoPope: yeah this is something we need to think about.

For GF(p) (when the degree is one) I think it's natural to allow:

fq_default(x) == fmpz_mod(x) == nmod(x) 

providing that the characteristic for fq_default matches the moduli for fmpz_mod, nmod -- in fact, even in FLINT the fmpz_mod type is used in some cases when the degree d = 1

For GF(p^d) for d != it's a little tricky there's two things which mathematically are interesting but im not sure how flint handles them.

1) If GF(p^d), GF(p^e) with e a divisor of d then there is an embedding of GF(p^e) into GF(p^d) and so elements of GF(p^e) all fit into GF(p^d). This embedding is only up to isomorphism though, because they could have different moduli

2) I suppose a special case of the above is when d = e but the moduli do not match -- there is a sense of equality here, but elements are only equal up to isomorphism, not just by comparing the polynomials.

I don't immediately see these fq_isomorphisms made available... but maybe they're here somewhere?

I think what might be sensible for now is:

1) if two elements are fq_default and have EXACTLY the same context then we can compare them
2) if one element is fq_default and another is fmpz_mod or nmod then they're equal if: a) the fq_default element has a lift to an integer AND the moduli match
3) if one element is fq_default and another is a int-like type, then they should match if int(fq_default) matches the int like type
4) if an fmpz_poly or nmod_poly has EXACTLY the same polynomial, then we could say these two things are equal? this is a little weird maybe?

src/flint/types/fq_default.pyx
GiacomoPope: i tried to fix this with a `swap=False` bool:

```py
    def __rsub__(self, other):
        other = self._any_as_self(other)
        if other is NotImplemented:
            return NotImplemented
        return self._sub_(other, swap=True)
```

src/flint/types/fq_default.pyx
GiacomoPope: I think i have done this as you want

src/flint/types/fq_default.pyx
GiacomoPope: Of course another option is to just allow equality on the type fq_default and not allow any other coercion

GiacomoPope: Oh! There's a whole thing for isomorphisms! This is what we should use for these checks then: https://flintlib.org/doc/fq_embed.html and not use equality here :)

GiacomoPope:
@fredrik-johansson do you know if there's a plan to make / include a `fq_default_embed` -- or should we have bindings for the three separate cases in python-flint

oscarbenjamin:
> if there's a plan to make / include a `fq_default_embed` -- or should we have bindings for the three separate cases in python-flint

It should be possible to just convert `fq_default` to `fq` and use the `fq_embed` functions. Then any returned `fq` can be converted back to the given `fq_default` contexts. The polynomials or matrices involved can use `fmpz_mod` although perhaps conversion to/from `nmod` could be used for the `fq_nmod/fq_zech` cases.

GiacomoPope:
I'd be interested if the fq_default_embed was planned though, as this limits the amount of internal python conversion needed?

src/flint/types/fq_default.pyx
GiacomoPope: OK i think I agree with this then

I'll summarise what I think could work.

- `fq_default` elements are only equal if their value is equal and their contexts are equal. This means `fq_default(p).one() != fq_default(p, d).one()` as they're in different fields (assume d != 1)
- `fq_default` elements are equal to `fmpz_mod` and `nmod` elements if the degree of the finite field is one and the prime matches `fq_default(p).one() == nmod(1, p)`

A trickier one, which requires changing `fmpz` maybe would be:

- `fmpz(x) = int(x)` for all `x`

as an example of this:

```
import gmpy2
assert gmpy2.mpz(1) == 1
```

holds and I think we should match this.

If this is true, then we need to decide about whether:

```
fmpz(1) == nmod(1, p) == fmpz_mod(1, p) == fq_default(1, p)
fmpz(-11) == nmod(p-1, p) == fmpz_mod(p-1, p) == fq_default(p-1, p)
```

When the degree of `fq_default` is one

GiacomoPope: This is what sage does:

```py
sage: R = Zmod(11)
sage: F = GF(11)
sage: R(1) == F(1) == int(1) == ZZ(1)
True
sage: R(-1) == F(-1) == int(-1) == ZZ(-1)
True
sage: R(10) == F(10) == int(10) == ZZ(10)
```

```py
sage: F = GF(11)
sage: F2 = GF(11**2)
sage: F(1) == F2(1)
True
sage: F(1) == F2(1) == ZZ(1) == int(1)
True
```

So they allow two elements to be coerced:

```py
sage: F2 = GF(11**2)
sage: F3 = GF(11**3)
sage: F2(1) == F3(1)
True
sage: F2(-1) == F3(-1)
True
```

oscarbenjamin: > At the moment we get something very very strange:

And also:
```python
In [9]: fmpz(2) == 2 == nmod(2, 3) != fmpz(2)
Out[9]: True

In [11]: fmpz(2) == 2 == 2.0 != fmpz(2)
Out[11]: True
```
If different pairs of objects use different notions of equality than transitivity is lost.

I doubt there is a unique good way to handle all of this especially when some things like `2 == 2.0` are outside of our control.

I think it is worth considering how other systems like Sage etc handle this sort of thing (not that any other system should be presumed "correct").

GiacomoPope: I think `mpz` having `mpz(1) == 1` convinces me that `fmpz(x) == int(x)` should always hold

I think if I have an element `nmod(x, p)` or `fmpz_mod(x, p)` etc it's natural to think of all equality modulo `p` so if we compare `nmod` types to `fmpz` or `int` types then the integer should be reducxed modulo p before comparison which i think would mean our design aligns with the sage snippet above

GiacomoPope: So maybe fq_defauly comparison could be something like...
```
if typecheck(other, fq_default):
    if element lifts to integers, only check value and prime matches
    else check value and whole context matches
elif typecheck(other fmpz):
    cast other to fq_default to reduce mod p and check equality
elif typecheck(other int):
    cast other to fq_default to reduce mod p and check equality
elif typecheck(other fmpz_mod):
    check modulus matches and value matches
elif typecheck(other nmod):
    check modulus matches and value matches
```

oscarbenjamin:
> It might be that no distros ship Flint 3.0 rather than 3.1 though in which case 3.1 would be a better baseline.

Another consideration is what version of Flint is used in Sage. I think python-flint is part of Sage now because it is used internally by SymPy and perhaps will be by mpmath in future. I assume that it is useful for Sage if python-flint has a bit of flexibility about which version of Flint it supports.

src/flint/types/fq_default.pyx
oscarbenjamin: > I think `mpz` having `mpz(1) == 1` convinces me that `fmpz(x) == int(x)` should always hold

So you would change the first two here:
```python
In [1]: from flint import *

In [2]: fmpz(2) == 2.0
Out[2]: False

In [3]: fmpz(2) == nmod(2, 3)
Out[3]: False

In [4]: fmpz(2) == int(2.0)
Out[4]: True

In [5]: fmpz(2) == int(nmod(2, 3))
Out[5]: True
```
Then we also have:
```python
In [6]: nmod(2, 3) == -1
Out[6]: True

In [7]: nmod(2, 3) == fmpz(-1)
Out[7]: False

In [8]: int(nmod(2, 3)) == fmpz(-1)
Out[8]: False
```
Perhaps it makes more sense to go the other way and say that when it comes to characteristic `p` the fmpz should be reduced modulo `p` for the comparison rather than lifting the nmod to fmpz. That is what happens for `nmod == int`.

oscarbenjamin: > Perhaps it makes more sense to go the other way and say that when it comes to characteristic `p`

If you think of `GF(p)` as `ZZ/(p)` then in some sense `ZZ` is the "ground domain" so it makes sense to promote from `ZZ` to `GF(p)` rather than vice versa...

GiacomoPope:
Attempting to resolve the discussion over equality and coercion I have written the above

oscarbenjamin:
> Attempting to resolve the discussion over equality and coercion I have written the above

The commit [f2443f8](https://github.com/flintlib/python-flint/pull/97/commits/f2443f80bd31e9f3390316870452395bee1067d3) looks reasonable to me. I think that it does imply that we need to change the way that `nmod` and `fmpz` compare though.

GiacomoPope:
i'd be tempted to do this in a second PR with an issue made to track it. The nmod implementation is less polished than other types atm and we could add a context and other bits while fixing the coercion?

oscarbenjamin:
> i'd be tempted to do this in a second PR

It doesn't all need to be done in one PR. We just need to be clear that that is the decision we are making.

GiacomoPope:
yeah, i wonder if we can use the GH discussions / wiki options to try and spell out some of the design decisions. Especially for things like coercion. 

Most of the pain of sagemath is because of the cursed coercion models they have, it'd be cool to try and centralise and simplify these coercions as much as possible

GiacomoPope:
@oscarbenjamin I had a quick go at trying to do some type checks to expose is_primitive, but I get set faults and floating point exception, so I'm probably doing something really stupid.

If you see the fix, let me know otherwise I'll undo the last commit

src/flint/flintlib/fq_default.pxd
```diff
@@ -0,0 +1,110 @@
...
+cdef extern from "flint/fq_default.h":
+    # TODO: how best to handle union
+    # https://stackoverflow.com/questions/12452210/cython-nesting-a-union-within-a-struct
+    ctypedef struct fq_default_struct:
+        fq_t fq
+        fq_nmod_t fq_nmod
+        fq_zech_t fq_zech
+        ulong nmod
+        fmpz_t fmpz_mod
+    ctypedef fq_default_struct fq_default_t[1]
```
oscarbenjamin, line 23: I'm not sure it matters but I think this should be `cdef union fq_default_struct`. The linked SO question refers to nested anonymous unions but this is just a union.

src/flint/types/fq_default.pyx
```diff
@@ -880,3 +880,15 @@ cdef class fq_default(flint_scalar):
         res = self.ctx.new_ctype_fq_default()
         fq_default_frobenius(res.val, self.val, <slong>e, self.ctx.val)
         return res
+
+    def is_primitive(self):
+        """
+        """
+        if self.ctx.fq_type == 1:
+            return 1 == fq_zech_is_primitive(<fq_zech_t>self.val, <fq_zech_ctx_t>self.ctx.val)
+        elif self.ctx.fq_type in [2, 4]:
```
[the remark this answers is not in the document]
GiacomoPope: this gets me new errors, and one I'm less likely to be able to fix as apparently there's some internal aliasing with `gr_ctx_struct`?

```
      [1/3] Compiling Cython source /Users/Jack/Documents/GitHub/python-flint/src/flint/types/fq_default.pyx
      [2/3] Compiling C object src/flint/types/fq_default.cpython-312-darwin.so.p/meson-generated_src_flint_types_fq_default.pyx.c.o
      FAILED: src/flint/types/fq_default.cpython-312-darwin.so.p/meson-generated_src_flint_types_fq_default.pyx.c.o
      cc -Isrc/flint/types/fq_default.cpython-312-darwin.so.p -Isrc/flint/types -I../../src/flint/types -I/usr/local/opt/python@3.12/Frameworks/Python.framework/Versions/3.12/include/python3.12 -I/usr/local/Cellar/gmp/6.3.0/include -I/usr/local/Cellar/mpfr/4.2.1/include -I/usr/local/Cellar/flint/3.1.3-p1/include -I/usr/local/opt/openssl@1.1/include -fvisibility=hidden -fdiagnostics-color=always -DNDEBUG -Wall -Winvalid-pch -O3 -MD -MQ src/flint/types/fq_default.cpython-312-darwin.so.p/meson-generated_src_flint_types_fq_default.pyx.c.o -MF src/flint/types/fq_default.cpython-312-darwin.so.p/meson-generated_src_flint_types_fq_default.pyx.c.o.d -o src/flint/types/fq_default.cpython-312-darwin.so.p/meson-generated_src_flint_types_fq_default.pyx.c.o -c src/flint/types/fq_default.cpython-312-darwin.so.p/src/flint/types/fq_default.pyx.c
      src/flint/types/fq_default.cpython-312-darwin.so.p/src/flint/types/fq_default.pyx.c:18934:118: error: no member named 'fq_zech' in 'gr_ctx_struct'
          __pyx_t_1 = __Pyx_PyBool_FromLong((1 == fq_zech_is_primitive(__pyx_v_self->val->fq_zech, __pyx_v_self->ctx->val->fq_zech))); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 888, __pyx_L1_error)
```

src/flint/flintlib/fq_default.pxd
GiacomoPope: I can modify this :)

GiacomoPope: Tests work locally for me, will push the change for CI

src/flint/types/fq_default.pyx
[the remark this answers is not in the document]
GiacomoPope: Looks like this might be associated to:

https://github.com/flintlib/flint/blob/05340d2a7349761aff2cd3370810c93e2174ce46/src/fq_default.h#L66-L73

oscarbenjamin:
I got it to compile with:
```diff
diff --git a/src/flint/types/fq_default.pyx b/src/flint/types/fq_default.pyx
index 7917dd9..828d6f1 100644
--- a/src/flint/types/fq_default.pyx
+++ b/src/flint/types/fq_default.pyx
@@ -7,6 +7,20 @@ from flint.types.fmpz_mod_poly cimport fmpz_mod_poly, fmpz_mod_poly_ctx
 from flint.types.nmod_poly cimport nmod_poly
 from flint.utils.typecheck cimport typecheck
 
+#define FQ_DEFAULT_CTX_FQ(ctx) ((fq_ctx_struct *) GR_CTX_DATA_AS_PTR(FQ_DEFAULT_GR_CTX(ctx)))
+#define FQ_DEFAULT_CTX_FQ_ZECH(ctx) ((fq_zech_ctx_struct *) GR_CTX_DATA_AS_PTR(FQ_DEFAULT_GR_CTX(ctx)))
+#define FQ_DEFAULT_CTX_FQ_NMOD(ctx) ((fq_nmod_ctx_struct *) GR_CTX_DATA_AS_PTR(FQ_DEFAULT_GR_CTX(ctx)))
+
+from flint.flintlib.fq cimport fq_ctx_t
+from flint.flintlib.fq_zech cimport fq_zech_ctx_t
+from flint.flintlib.fq_nmod cimport fq_nmod_ctx_t
+
+cdef extern from "flint/fq_default.h":
+    cdef fq_ctx_t FQ_DEFAULT_CTX_FQ(fq_default_ctx_t ctx)
+    cdef fq_zech_ctx_t FQ_DEFAULT_CTX_FQ_ZECH(fq_default_ctx_t ctx)
+    cdef fq_nmod_ctx_t FQ_DEFAULT_CTX_FQ_NMOD(fq_default_ctx_t ctx)
+
+
 cdef class fq_default_ctx:
     r"""
     Context object for creating :class:`~.fq_default`.
@@ -881,15 +895,16 @@ cdef class fq_default(flint_scalar):
         fq_default_frobenius(res.val, self.val, <slong>e, self.ctx.val)
         return res
 
-    # TODO: this crashes as the context cannot be found during compile time.
-    # def is_primitive(self):
-    #     """
-    #     """
-    #     if self.ctx.fq_type == 1:
-    #         return 1 == fq_zech_is_primitive((<fq_default>self).val.fq_zech, (<fq_default_ctx>self.ctx).val.fq_zech)
-    #     elif self.ctx.fq_type == 2:
-    #         return 1 == fq_nmod_is_primitive((<fq_default>self).val.fq_nmod, (<fq_default_ctx>self.ctx).val.fq_nmod)
-    #     elif self.ctx.fq_type == 3:
-    #         return 1 == fq_is_primitive((<fq_default>self).val.fq, (<fq_default_ctx>self.ctx).val.fq)
-    #     else:
-    #         return NotImplemented
+    def is_primitive(self):
+        """
+        """
+        cdef fq_default s
+        s = self
+        if self.ctx.fq_type == 1:
+            return 1 == fq_zech_is_primitive(s.val.fq_zech, FQ_DEFAULT_CTX_FQ_ZECH(s.ctx.val))
+        elif self.ctx.fq_type == 2:
+            return 1 == fq_nmod_is_primitive(s.val.fq_nmod, FQ_DEFAULT_CTX_FQ_NMOD(s.ctx.val))
+        elif self.ctx.fq_type == 3:
+            return 1 == fq_is_primitive(s.val.fq, FQ_DEFAULT_CTX_FQ(s.ctx.val))
+        else:
+            return NotImplemented
```
Then:
```python
In [1]: from flint import fq_default_ctx

In [2]: F = fq_default_ctx(3, 2)

In [3]: x = F.gen()

In [4]: x
Out[4]: x

In [5]: x.is_primitive()
Out[5]: True

In [6]: (1 + x).is_primitive()
Out[6]: False
```

oscarbenjamin:
It is probably worth having `cdef` helpers like
```python
self.ctx.get_fq_ctx_t()
```

GiacomoPope:
oh amazing, nice work! I'll make the change now :)

GiacomoPope:
Hmm I can't make the cdef helper functions, I'm being told the struct type is an array?

```py
    cdef fq_ctx_t get_fq_ctx_t(self):
        """
        Return the fq_ctx_t type from the context
        """
        cdef fq_default s
        s = self
        return FQ_DEFAULT_CTX_FQ(s.ctx.val)
```
      /Users/Jack/Documents/GitHub/python-flint/src/flint/types/fq_default.pyx:443:9: Function cannot return an array

GiacomoPope:
This also brings up another question -- Flint currently uses `FMPZ_MOD` and `NMOD` instead of the `FQ_` counterparks for degree 1 extensions which just makes some of these edge cases annoying. I wonder if we should enforce that python-flint only picks one of the 1, 2, 3 types and ignore the casting to NMOD and FMPZ_MOD when the degree is 1?

GiacomoPope:
:( it also looks like 3.0 doesnt have `FQ_DEFAULT_CTX_FQ_NMOD` -- not sure how to deal with this now. Maybe just remove the functionalities until 3.1 is the minimum supported version?

oscarbenjamin:
> I'm being told the struct type is an array?

Maybe just declare it as returning a pointer to the struct?

This also works:
```diff
diff --git a/src/flint/types/fq_default.pyx b/src/flint/types/fq_default.pyx
index 7917dd9..e43415a 100644
--- a/src/flint/types/fq_default.pyx
+++ b/src/flint/types/fq_default.pyx
@@ -7,6 +7,16 @@ from flint.types.fmpz_mod_poly cimport fmpz_mod_poly, fmpz_mod_poly_ctx
 from flint.types.nmod_poly cimport nmod_poly
 from flint.utils.typecheck cimport typecheck
 
+from flint.flintlib.fq cimport fq_ctx_struct
+from flint.flintlib.fq_zech cimport fq_zech_ctx_struct
+from flint.flintlib.fq_nmod cimport fq_nmod_ctx_struct
+
+cdef extern from "flint/fq_default.h":
+    cdef fq_ctx_struct *FQ_DEFAULT_CTX_FQ(fq_default_ctx_t ctx)
+    cdef fq_zech_ctx_struct *FQ_DEFAULT_CTX_FQ_ZECH(fq_default_ctx_t ctx)
+    cdef fq_nmod_ctx_struct *FQ_DEFAULT_CTX_FQ_NMOD(fq_default_ctx_t ctx)
+
+
 cdef class fq_default_ctx:
     r"""
     Context object for creating :class:`~.fq_default`.
@@ -881,15 +891,16 @@ cdef class fq_default(flint_scalar):
         fq_default_frobenius(res.val, self.val, <slong>e, self.ctx.val)
         return res
 
-    # TODO: this crashes as the context cannot be found during compile time.
-    # def is_primitive(self):
-    #     """
-    #     """
-    #     if self.ctx.fq_type == 1:
-    #         return 1 == fq_zech_is_primitive((<fq_default>self).val.fq_zech, (<fq_default_ctx>self.ctx).val.fq_zech)
-    #     elif self.ctx.fq_type == 2:
-    #         return 1 == fq_nmod_is_primitive((<fq_default>self).val.fq_nmod, (<fq_default_ctx>self.ctx).val.fq_nmod)
-    #     elif self.ctx.fq_type == 3:
-    #         return 1 == fq_is_primitive((<fq_default>self).val.fq, (<fq_default_ctx>self.ctx).val.fq)
-    #     else:
-    #         return NotImplemented
+    def is_primitive(self):
+        """
+        """
+        cdef fq_default s
+        s = self
+        if self.ctx.fq_type == 1:
+            return 1 == fq_zech_is_primitive(s.val.fq_zech, FQ_DEFAULT_CTX_FQ_ZECH(s.ctx.val))
+        elif self.ctx.fq_type == 2:
+            return 1 == fq_nmod_is_primitive(s.val.fq_nmod, FQ_DEFAULT_CTX_FQ_NMOD(s.ctx.val))
+        elif self.ctx.fq_type == 3:
+            return 1 == fq_is_primitive(s.val.fq, FQ_DEFAULT_CTX_FQ(s.ctx.val))
+        else:
+            return NotImplemented
```

GiacomoPope:
is this worth fixing if Flint 3.0 doesnt support it?

oscarbenjamin:
> is this worth fixing if Flint 3.0 doesnt support it?

It is possible to make this work, just a little awkward. It needs to be handled in the C preprocessor because Cython has no knowledge of the Flint version. We can insert preprocessor defines like this though:
https://github.com/flintlib/python-flint/blob/c4847e72fcc31c6d1f56653935705679cffa3d1d/src/flint/flintlib/flint.pxd#L39-L50

We just need a preprocessor define like:
```
#define FQ_DEFAULT_CTX_FQ_NMOD 12345 /* unlikely number */
```

GiacomoPope:
The inputs into the functions are different types (both for `fq_t` and `fq_ctx_t` so we would need some kind of additional conversion here.

An expensive solution for these two types would be to have a second context created to coerce into types 1,2,3 for these special functions. I don't know how to do this any better than calling `fq_default_ctx(mod=self.ctx.modulus() fq_type=(1/2))` and using this for the function call.

oscarbenjamin:
> An expensive solution for these two types would be to have a second context created to coerce into types 1,2,3 for these special functions. I don't know how to do this any better than calling `fq_default_ctx(mod=self.ctx.modulus() fq_type=(1/2))` and using this for the function call.

This is what I meant by keeping another context lying around. Performing a conversion here is fine but we don't want to create the context every time so we have:
```
cdef class fq_default_ctx:
     cdef fq_default_ctx _fq_ctx
```
Then `_fq_ctx` is initialised to `None` by default but  a method like `ctx.get_fq_ctx()` creates (and stores) a corresponding `FQ_DEFAULT_FQ` context. Then it is always possible to access any functionality that `fq` has regardless of what the internal type is:
```
fq_ctx = self.ctx.get_fq_ctx()
return fq_ctx(self).is_primitive()
```
Or inlining the C calls if speed is needed:
```
cdef fq_default_ctx fq_ctx = self.ctx.get_fq_ctx()
cdef fq_ctx_t c_fq_ctx = FQ_DEFAULT_CTX_FQ(ctx.val)
cdef fq_t s_fq
fq_init(s_fq)
fq_set_fmpz(s_fq, self.val, c_fq_ctx)
res = 1 == fq_is_primitive(s_fq, c_fq_ctx)
fq_clear(s_fq)
```
That would want some kind of helper function if it needs using more than once.

GiacomoPope:
> Maybe just declare it as returning a pointer to the struct?

I tried this but I couldnt get cython right to find out how to return the pointer as a C-type -- i get errors about it trying to convert the type to a Python object however i refactor it

GiacomoPope:
The above is also a nice  idea -- i need a break from this but if you want, please add this change (either as a diff or a commit) -- im a little cython confused atm :D

oscarbenjamin:
> i need a break from this

Fair enough maybe to make things easier we can just postpone adding `is_primitive` or comment it out for now.

src/flint/flint_base/flint_base.pyx
```diff
@@ -30,7 +30,127 @@ cdef class flint_elem:
...
+        other = self._any_as_self(other)
+        if other is NotImplemented:
+            return NotImplemented
+        return self._sub_(other)
+
+    def __rsub__(self, other):
+        other = self._any_as_self(other)
+        if other is NotImplemented:
+            return NotImplemented
+        return self._sub_(other, swap=True)
```
oscarbenjamin, line 97: Rather than having `swap=True` I would just have two methods `_sub_` and `_rsub_`.

oscarbenjamin: The idea is that `__sub__` handles all conditionality so that `_sub_` and `_rsub_` do not need to check anything. Also keyword arguments can add overhead that I would avoid at this level.

GiacomoPope: OK I've done this for the base class and added `rsub`, `rdiv`.

GiacomoPope:
>  postpone adding is_primitive

Maybe we can circle back to this when we deal with `fq_embed` which will need more complex use of the `fq_ctx` etc and we'll need to build a bunch of helpers to get correct contexts

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,933 @@
...
+        return res
+
+    def _add_(self, other):
+        """
+        Assumes that __add__() has ensured other is of type self
+        """
+        cdef fq_default res
+        res = self.ctx.new_ctype_fq_default()
+        fq_default_add(res.val, self.val, (<fq_default>other).val, self.ctx.val)
+        return res
```
oscarbenjamin, line 662: I would add types for the parameters here and make this cdef I think:
```python
cdef fq_default _add_(fq_default self, fq_default other):
    cdef fq_default res = self.ctx.new_ctype_fq_default()
    fq_default_add(res.val, self.val, other.val, self.ctx.val)
    return res
```

oscarbenjamin: I'm not sure if making this `cdef` makes much difference though.

oscarbenjamin:
I'll give this a full read later but I think it's probably good to go.

src/flint/types/fq_default.pyx
GiacomoPope: I tried doing this and it broke all the arithmetic... Is there a non-trivial change needed to be made to the base class too?

oscarbenjamin: I'm not sure what is the best way to make this work. If we want to use this widely though (for other types) then it would be good to benchmark, look at the generated C code, try alternatives etc.

I am concerned about the overhead of these methods. See e.g.: https://github.com/sympy/sympy/discussions/26932. I got 10% speedup on a macro benchmark that used fmpz just by inlining the type checking in `__add__`:
```diff
diff --git a/src/flint/types/fmpz.pyx b/src/flint/types/fmpz.pyx
index 0b16375..0a025b8 100644
--- a/src/flint/types/fmpz.pyx
+++ b/src/flint/types/fmpz.pyx
@@ -1,3 +1,5 @@
+from cpython.object cimport PyTypeObject, PyObject_TypeCheck
+
 from flint.flint_base.flint_base cimport flint_scalar
 from flint.utils.typecheck cimport typecheck
 from flint.utils.conversion cimport chars_from_str
@@ -185,6 +187,10 @@ cdef class fmpz(flint_scalar):
         return -self
 
     def __add__(s, t):
+        if PyObject_TypeCheck(t, <PyTypeObject*>fmpz):
+            u = fmpz.__new__(fmpz)
+            fmpz_add((<fmpz>u).val, (<fmpz>s).val, (<fmpz>t).val)
+            return u
         cdef fmpz_struct tval[1]
         cdef int ttype = FMPZ_UNKNOWN
         u = NotImplemented
```
It still came out notably slower than when using gmpy2 though so there is definitely some overhead in the Cython wrapper that could be reduced.

GiacomoPope: Maybe we should just prioritise speed for all these scalar types and accept we have to hand write all the __add__ methods etc. seems silly to lose a bunch of performance just for a few smaller functions.

oscarbenjamin: We need to measure the effects. I think it is fine to have a framework in the superclass but then inline common cases. Note that e.g. for SymPy the case that really matters is the common case like `fmpz*fmpz` and to a much lesser extent `fmpz*int`. We could have an `fmpz.__add__` method that optimises that common case as hard as possible but then just falls back on something slower that handles coercions etc:
```python
class fmpz(scalar):
    def __add__(self, other):
        if PyTypeCheck(...):
            # Fast optimised case
        else:
            # Fall back on slower generic routine
            return super().__add__(self, other)
```
It makes sense to start out having a nice generic routine in the superclass that handles everything correctly and then after doing timing measurements micro-optimise the cases that matter.

GiacomoPope: OK. We can work this way then. I think int * type and type * type are the ones worth doing first as I think it'll be comment for people to do things like 5 * type or type + 1 in their code and it would be nice for this to be fast without having to do one = type(one); type + one

oscarbenjamin: Exactly and almost always the ints would be small so the fast path could use `fmpz_mul_si` and friends rather than general `int -> fmpz` coercion.

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,886 @@
...
+        - `fq_default_ctx.FQ`: Using `fq_t`.
+        """
+        FQ_TYPES = {
+            1 : "FQ_ZECH",
+            2 : "FQ_NMOD",
+            3 : "FQ",
+            4 : "NMOD",
+            5 : "FMPZ_MOD",
+        }
+        return FQ_TYPES[self.fq_type]
```
oscarbenjamin, line 212: I think it should be possible to get the string from the enum if `self.fq_type` returns the enum like:
```
return self.fq_type._name_
```
Then we don't need to duplicate the names and numbers here.

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,886 @@
...
+    def __dealloc__(self):
+        if self._initialized:
+            fq_default_ctx_clear(self.val)
+
+    @staticmethod
+    def _parse_input_fq_type(fq_type):
+        # Allow the type to be denoted by strings or integers
+        FQ_TYPES = {
+            "FQ_ZECH" : 1,
+            "FQ_NMOD" : 2,
```
oscarbenjamin, line 43: This dictionary is already provided by the enum type:
```python
In [35]: fq_default_type._member_map_
Out[35]: 
{'DEFAULT': <fq_default_type.DEFAULT: 0>,
 'FQ_ZECH': <fq_default_type.FQ_ZECH: 1>,
 'FQ_NMOD': <fq_default_type.FQ_NMOD: 2>,
 'FQ': <fq_default_type.FQ: 3>,
 'NMOD': <fq_default_type.NMOD: 4>,
 'FMPZ_MOD': <fq_default_type.FMPZ_MOD: 5>}
```

oscarbenjamin: I guess this function can just be:
```python
try:
    return fq_default_type._member_map_[fq_type]
except KeyError:
    pass

try:
    return fq_default_type._value2member_map_[fq_type]
except KeyError:
    pass

raise ValueError(f"fq_type should be one of {fq_default_type._member_names_}")

File: src/flint/types/fq_default.pyx [hunk 14]
```diff
@@ -0,0 +1,886 @@
...
+
+cdef class fq_default_ctx:
+    r"""
+    Context object for creating :class:`~.fq_default`.
+
+    Finite fields can be initialized in one of two possible ways. The
+    first is by providing characteristic and degree:
+
+        >>> fq_default_ctx(5, 2, 'y', fq_type=1)
+        fq_default_ctx(5, 2, 'y', x^2 + 4*x + 2, 'FQ_ZECH')
```
oscarbenjamin, line 18: Probably better to use `fq_type='FQ_ZECH'` or `fq_type=fq_default_type.FQ_ZECH` here rather than `fq_type=1`. We shouldn't encourage writing code that uses magic values like that.

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,886 @@
...
+
+    For more details, see the documentation of :method:`~.set_from_order`
+    and :method:`~.set_from_modulus`.
+    """
+    def __cinit__(self):
+        pass
+
+    def __dealloc__(self):
+        if self._initialized:
+            fq_default_ctx_clear(self.val)
```
oscarbenjamin, line 36: There was some discussion about this in the mpoly PR so I need to find out what the situation is. Apparently Cython does not guarantee that `__dealloc__` is only ever called once. If that is the case then it means that we also need to set `self._initialized = False` here after calling `clear` to prevent a double-free.

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,886 @@
...
+    cdef _c_set_from_order(self, fmpz p, int d, char *var,
+                           fq_default_type fq_type=fq_default_type.DEFAULT):
+        self.var = var
+        fq_default_ctx_init_type(self.val, p.val, d, self.var, fq_type)
+        self._initialized = True
+
+    def set_from_order(self, p, d, var,
+                       fq_type=fq_default_type.DEFAULT, check_prime=True):
+        """
+        Construct a context for the finite field GF(p^d).
```
oscarbenjamin, line 122: I don't think we want these context-mutating methods to be public. I assume that it would wreak havoc if someone mutated the context after creating elements...

oscarbenjamin: Maybe make them `cdef` only with leading underscore names.

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,886 @@
...
+
+    def degree(self):
+        """
+        The extension degree of the finite field
+
+            >>> gf = fq_default_ctx(5, 2)
+            >>> gf.degree()
+            2
+        """
+        return fq_default_ctx_degree(self.val)
```
oscarbenjamin, line 222: Would it make sense to return fmpz rather than int here?

I guess `degree` is always small unlike `characteristic` and `order` so maybe `fmpz` does not make sense here...

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,886 @@
...
+        return 0
+
+    cdef set_any_scalar_as_fq_default(self, fq_default_t fq_ele, obj):
+        if typecheck(obj, int):
+            # For small integers we can convert directly
+            if obj < 0 and obj.bit_length() < 31:
+                fq_default_set_si(fq_ele, <slong>obj, self.val)
+            elif obj.bit_length() < 32:
+                fq_default_set_ui(fq_ele, <ulong>obj, self.val)
+            # For larger integers we first convert to fmpz
```
oscarbenjamin, line 360: The signed to unsigned cast here is unsafe:
```python
In [65]: F = fq_default_ctx(2)

In [66]: F(-2**31+1)
OverflowError                             Traceback (most recent call last)
Cell In[66], line 1
----> 1 F(-2**31+1)

File fq_default.pyx:467, in flint.types.fq_default.fq_default_ctx.__call__()

File fq_default.pyx:492, in flint.types.fq_default.fq_default.__init__()

File fq_default.pyx:388, in flint.types.fq_default.fq_default_ctx.set_any_as_fq_default()

File fq_default.pyx:359, in flint.types.fq_default.fq_default_ctx.set_any_scalar_as_fq_default()

OverflowError: can't convert negative value to ulong
```
I would just use a single `set_si` case here for both positive and negative values. Can probably catch OverflowError like:
```
slong cdef i
try:
    i = ...
```

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,886 @@
...
+        if self is other:
+            return True
+
+        if typecheck(other, fq_default_ctx):
+            return (self.fq_type == other.fq_type
+                    and self.var == other.var
+                    and self.prime() == other.prime()
+                    and self.modulus() == other.modulus())
+        else:
+            raise TypeError(f"Cannot compare fq_default_ctx with {type(other)}")
```
oscarbenjamin, line 451: `__eq__` should return `NotImplemented` rather than raising `TypeError`:
```
In [67]: F == None
TypeError                                 Traceback (most recent call last)
Cell In[67], line 1
----> 1 F == None

File fq_default.pyx:451, in flint.types.fq_default.fq_default_ctx.__eq__()

TypeError: Cannot compare fq_default_ctx with <class 'NoneType'>
```
Generally considered a faux-pas for `==` to raise an exception rather than turning into `False`.

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,886 @@
...
+        return coeffs
+
+    def str(self):
+        return self.polynomial().str(var=self.ctx.var.decode())
+
+    def repr(self):
+        return f"fq_default({self.to_list(), self.ctx.__repr__()})"
+
+    def __hash__(self):
+        return hash((self.to_polynomial(), hash(self.ctx)))
```
oscarbenjamin, line 556: I think we probably need coverage measurement to make sure that cases like this are checked:
```python
In [69]: hash(F(4))
AttributeError                            Traceback (most recent call last)
Cell In[69], line 1
----> 1 hash(F(4))

File fq_default.pyx:556, in flint.types.fq_default.fq_default.__hash__()

AttributeError: 'flint.types.fq_default.fq_default' object has no attribute 'to_polynomial'
```

src/flint/types/fq_default.pyx
```diff
@@ -0,0 +1,886 @@
...
+        # Ignores the modulus of nmod
+        if typecheck(obj, nmod):
+            fq_default_set_ui(fq_ele, <ulong>(<nmod>obj).val, self.val)
+            return 0
+
+        # For fmpz_mod we can also convert directly
+        # Assumes that the modulus of the ring matches the context
+        if typecheck(obj, fmpz_mod):
+            fq_default_set_fmpz(fq_ele, (<fmpz_mod>obj).val, self.val)
+            return 0
```
oscarbenjamin, line 381: If this function is being used by `__add__` etc then the modulus needs to be checked:
```python
In [70]: F = fq_default_ctx(3)

In [71]: F(2) + nmod(2, 7)
Out[71]: 1
```

oscarbenjamin (commented):
Broadly this is good but I found a bunch of small things.

src/flint/types/fq_default.pyx
GiacomoPope: I'm getting an error that `fq_default_type` doesn't have a member map... I'll try and figure out why

src/flint/types/fq_default.pyx
GiacomoPope: the degree should always be pretty small... but we could do fmpz to keep things in the flint types if you want -- i think i would lean towards int

src/flint/types/fq_default.pyx
GiacomoPope: oh i thought i handled this... i must have forgotten to do it.

src/flint/types/fq_default.pyx
oscarbenjamin: Let's keep it as `int` then.

src/flint/types/fq_default.pyx
GiacomoPope: I have it available in python:

```py
In [1]: from flint import fq_default_ctx, fq_default_type

In [2]: fq_default_type
Out[2]: <flag 'fq_default_type'>

In [3]: fq_default_type._member_map_
Out[3]: 
{'DEFAULT': <fq_default_type.DEFAULT: 0>,
 'FQ_ZECH': <fq_default_type.FQ_ZECH: 1>,
 'FQ_NMOD': <fq_default_type.FQ_NMOD: 2>,
 'FQ': <fq_default_type.FQ: 3>,
 'NMOD': <fq_default_type.NMOD: 4>,
 'FMPZ_MOD': <fq_default_type.FMPZ_MOD: 5>}

In [4]: type(fq_default_type)
Out[4]: enum.EnumType
```

But for compilation of the cython I get:

```
      Error compiling Cython file:
      ...
              return fq_type
      
          @staticmethod
          def _parse_input_var(var):
              try:
                  return fq_default_type._member_map_[fq_type]
                                        ^
      
      /Users/Jack/Documents/GitHub/python-flint/src/flint/types/fq_default.pyx:65:34: _member_map_ not a known value of fq_default_type
```

oscarbenjamin: I guess that the C-level enum is different in the cpdef enum.

Maybe using the enum type for validation isn't quite a good fit and we just need to have a dict somewhere to map values from one to the other.

Either way we need there to be a single place where the list of values is stored rather than several different dicts spread around.

src/flint/types/fq_default.pxd
```diff
@@ -0,0 +1,57 @@
...
+- 2. `fq_default_ctx.FQ_NMOD`: Use `fq_nmod_t`,
+- 3. `fq_default_ctx.FQ`: Use `fq_t`.
+- 4. `fq_default_ctx.NMOD`: Use `nmod` for degree = 1,
+- 5. `fq_default_ctx.FMPZ_MOD`: Use `fmpz_mod` for degree = 1.
+
+These can be manually selected, or type: `fq_default_ctx.DEFAULT` can be used
+for the implementation to be automatically decided by Flint (default),
+"""
+cpdef enum fq_default_type:
+    DEFAULT  = 0
```
oscarbenjamin, line 21: Does the docstring not need to inside the enum statement?

oscarbenjamin:
One thing I realised is that we need to add this to the docs.

oscarbenjamin:
We need a `doc/source/fq_default.rst`.

oscarbenjamin:
Let's get this one in now. It would be good to improve the test coverage but 

I suppose that can be done later. I think all the substantial issues were resolved.

Thanks!

GiacomoPope:
Maybe we can find a hacky way to make pytest run the doctests and have coverage test this? If we get a good coverage test done im happy to circle back and improve this

oscarbenjamin:
> Maybe we can find a hacky way to make pytest run the doctests and have coverage test this?

We can still use setuptools for it but I guess the setup.py just needs fixing:
```
$ PYTHONPATH=src bin/coverage.sh
```
I think this is enough to fix it:
```diff
diff --git a/setup.py b/setup.py
index d260834..82ce32d 100644
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,6 @@ ext_files = [
     ("flint.types.fmpz_mod_mat", ["src/flint/types/fmpz_mod_mat.pyx"]),
 
     ("flint.types.fmpq_mpoly", ["src/flint/types/fmpq_mpoly.pyx"]),
-    ("flint.types.fmpz_mpoly_q", ["src/flint/types/fmpz_mpoly_q.pyx"]),
 
     ("flint.types.arf", ["src/flint/types/arf.pyx"]),
     ("flint.types.arb", ["src/flint/types/arb.pyx"]),
```

oscarbenjamin:
I did have a look on Friday at getting Cython coverage to work with meson. The problem is that Cython's coverage plugin works by using heuristics based on setuptools. It basically needs a rewrite. Probably the rewritten version should be maintained as part of something like `spin` rather than Cython.

PR #104, opened 2023-11-10
maint: bump Flint to 3.0.1

oscarbenjamin:


oscarbenjamin:
Release announcement:

https://groups.google.com/g/flint-devel/c/2TOM_GkSyBc/m/SkmlzLsgAwAJ

When Flint 3.1.0 comes out we should make a separate CI job that builds and tests against 3.0.1 so that we can keep that as a minimum supported version. I don't think it is worth doing that for 3.0.0 though.

PR #106, opened 2023-11-15
Add fmpz_mod_mat

oscarbenjamin:
Since fmpz_mod and fmpz_mod_poly were recently added (gh-87 gh-83) this PR adds the corresponding matrix type fmpz_mod_mat.

The tests added are mostly generic tests applied to the 4 exact matrix types fmpz_mat, fmpq_mat, nmod_mat and fmpz_mod_mat. Various methods are added to some of the other types so that they have more consistent interfaces. Also some exception types were changed and some cases where Flint would abort are now handled with Python exceptions (e.g. charpoly on a non-square matrix).

Also nmod_mat now subclasses flint_mat along with the other matrix types and the matrix printing was made more consistent in how ctx.pretty is handled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ MANIFEST
 *.swp
 .python-version
 *.DS_Store
+.venv
--- a/doc/source/fmpz_mod_mat.rst
+++ b/doc/source/fmpz_mod_mat.rst
@@ -1,0 +1,8 @@
+**fmpz_mod_mat** -- matrices over integers mod n for arbitrary n
+===============================================================================
+
+.. autoclass :: flint.fmpz_mod_mat
+  :members:
+  :inherited-members:
+  :undoc-members:
+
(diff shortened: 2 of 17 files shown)


oscarbenjamin:
Does anyone want to review this @deinst @GiacomoPope or should I just merge?

GiacomoPope:
I had a quick look and it looks good. Nice work with the tests and aligning the other matrix types with this new addition.

oscarbenjamin:
Thanks.

Writing the generic test code would be a lot easier if we had a more uniform interface for the different types wrt contexts e.g.:
```
Rs = [fmpz_ctx(), fmpq_ctx(), nmod_ctx(7), fmpz_mod_ctx(7), ...]
for R in Rs:
    a = R(2)
    p = R.poly([1, 2])
    M = R.matrix([[1, 2], [3, 4]])
    assert M.charpoly() == R.poly([1, 2, 3])
    ...
```
Working on this has made it clear that a uniform interface is worthwhile so I think I will add that soon.

oscarbenjamin:
Okay, let's get this in.

PR #108, opened 2024-01-25
Declare build deps in `pyproject.toml`

mkoeppe:


oscarbenjamin:
Thanks. I had been thinking about doing this for a while but I wasn't sure about all implications given that setup.py uses distutils rather than setuptools (for Python < 3.12). I guess it is fine to call the backend setuptools in any case.

This should fix pip install for the special case that compilers and flint are already installed but a python-flint wheel is not available and should mean that `--no-build-isolation` is not needed any more.

There is also discussion about pyproject.toml in gh-52. Another thing that would be good to move to pyproject.toml would be all the cibuildwheel config which is currently duplicated in CI config/scripts:
https://github.com/flintlib/python-flint/blob/e5da576ed4f6a3e6419c038c1d37099b6c344442/.github/workflows/buildwheel.yml#L36-L53
https://github.com/flintlib/python-flint/blob/e5da576ed4f6a3e6419c038c1d37099b6c344442/bin/cibw.sh#L12-L36

oscarbenjamin:
Thanks Matthias

mkoeppe:
> `--no-build-isolation` is not needed any more

#111 updates the README accordingly

PR #109, opened 2024-01-25
Use exact division for `a/b` for `fmpz`, `fmpz_mat`, and `*_poly`

oscarbenjamin:
Use exact division for fmpz etc e.g.:
```
In [3]: fmpz(4)/fmpz(2)
Out[3]: 2

In [4]: fmpz(4)/fmpz(3)
DomainError
```
Prevously fmpz_mat and fmpz_poly when divided by fmpz would give convert to fmpq but now only exact division is supported:
```
In [5]: fmpz_poly([2, 4])
Out[5]: 4*x + 2

In [6]: fmpz_poly([2, 4]) / 2
Out[6]: 2*x + 1

In [7]: fmpz_poly([2, 4]) / 3
DomainError
```
Also add exact division for `fmpz_poly`, `fmpq_poly`, `nmod_poly` and `fmpz_mod_poly`:
```
In [12]: fmpz_poly([2, 4]) / fmpz_poly([1, 2])
Out[12]: 2

In [13]: fmpz_poly([2, 4]) / fmpz_poly([1, 3])
DomainError
```
This is all working as far as I can tell apart from the case of `nmod_poly` and `fmpz_mod_poly` with non-prime moduli e.g.:
```
In [15]: nmod_poly([2, 3], 10) / 2
    Cannot invert modulo 2*5
Aborted (core dumped)
```
The same core dump is seen with divmod (an existing problem, not changed in this PR):
```
divmod(nmod_poly([2, 3], 10), 2)
    Cannot invert modulo 2*5
Aborted (core dumped)
```

### DIFF
--- a/src/flint/test/test.py
+++ b/src/flint/test/test.py
@@ -547,7 +547,8 @@ def test_fmpz_mat():
     assert str(M(2,2,[1,2,3,4])) == '[1, 2]\n[3, 4]'
     assert M(1,2,[3,4]) * flint.fmpq(1,3) == flint.fmpq_mat(1, 2, [1, flint.fmpq(4,3)])
     assert flint.fmpq(1,3) * M(1,2,[3,4]) == flint.fmpq_mat(1, 2, [1, flint.fmpq(4,3)])
-    assert M(1,2,[3,4]) / 3 == flint.fmpq_mat(1, 2, [1, flint.fmpq(4,3)])
+    assert raises(lambda: M(1,2,[3,4]) / 3, DomainError)
+    assert M(1,2,[2,4]) / 2 == M(1,2,[1,2])
     assert M(2,2,[1,2,3,4]).inv().det() == flint.fmpq(1) / M(2,2,[1,2,3,4]).det()
     assert M(2,2,[1,2,3,4]).inv().inv() == M(2,2,[1,2,3,4])
     assert raises(lambda: M.randrank(4,3,4,1), ValueError)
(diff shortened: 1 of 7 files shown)


oscarbenjamin:
There is a bug in the `fmpz_mod_poly.exact_division` method on master (not caused by this PR) when the modulus is composite:
```python
In [1]: from flint import *

In [2]:     F_cmp = fmpz_mod_ctx(10)
   ...:     R_cmp = fmpz_mod_poly_ctx(F_cmp)
   ...:     f_cmp = R_cmp([1,2,3,4,5])
   ...:     f_bad = R_cmp([2,2,2,2,2])

In [3]: f_cmp.exact_division(R_cmp([2]))
Out[3]: 0
```
The code is here:
https://github.com/flintlib/python-flint/blob/ea65d5fd0e07ecb28114cd59050ad963d7f044b6/src/flint/types/fmpz_mod_poly.pyx#L457-L489
As far as I can tell everything works up to the call to `fmpz_mod_poly_divides` which returns 1 (meaning divisible) but sets the quotient to zero.

Maybe `fmpz_mod_poly_divides` should not be used with non-prime moduli although that is not documented. It is not clear to me if there is an alternative that could be used. I'm not sure how much to worry about the case of non-prime moduli in general. I don't personally have any use for it and it only seems to have only patchy support in Flint.

Maybe just anything related to any kind of division should be disallowed in python-flint for non-prime characteristic. Otherwise it seems like there will always be cases where python-flint calls some function that is not guaranteed to return sensible results or might abort on some inputs.

Checking if the modulus is prime can be done when creating the context so it does not need to slow down basic operations.

oscarbenjamin:
I think that non-prime moduli can be dealt with later. There would need to be a context object for nmod so that it can keep track of whether or not the modulus is prime.

For now this gets the right behaviour in the well-defined cases so let's get it in.

PR #110, opened 2024-01-25
pyproject.toml: Move project metadata here from setup.py

mkoeppe:
That's preparation for but independent of the decision on the build system to use (#52).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,3 +3,20 @@ requires = ["setuptools",
             "numpy",
             "Cython>=3"]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "python-flint"
+description = "Bindings for FLINT and Arb"
+version = "0.5.0"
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
Thanks

oscarbenjamin:
To be clear I think the decision about a build system is not difficult. I would just use meson but I'm happy to consider alternatives if others would prefer something else. I haven't see any strong opinions about it.

The hold up on switching to meson is just finding the time. I don't know meson that well so it would take me a little while to work out exactly how to do it.

The primary motivation for switching away from setuptools is really just that it doesn't manage building the dependencies like flint. A first step to moving to meson would just be to have a meson configuration that could do the equivalent of `bin/build_dependencies_unix.sh` which makes a local development build of GMP, MPFR and Flint. That part is already not managed by setuptools anyway.

The main constraint is just that whatever system is used needs to work somehow on Windows.

PR #112, opened 2024-01-25
pyproject.toml, setup.py: Only use numpy.distutils on win32

mkoeppe:
Using `setuptools` on all other platforms, as is already done when Python == 3.12 after #100.

Ref https://github.com/flintlib/python-flint/issues/52#issuecomment-1774196119 @oscarbenjamin

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from Cython.Distutils import build_ext
 from Cython.Build import cythonize
 
 
-if sys.version_info < (3, 12):
+if sys.platform == 'win32' and sys.version_info < (3, 12):
     from distutils.core import setup
     from distutils.extension import Extension
     from numpy.distutils.system_info import default_include_dirs, default_lib_dirs
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
 requires = ["setuptools",
-            "numpy",
+            "numpy; sys_platform == 'win32' and python_version < '3.12'",
             "Cython>=3"]
 build-backend = "setuptools.build_meta"


setup.py
```diff
@@ -6,7 +6,7 @@
 from Cython.Build import cythonize
 
 
-if sys.version_info < (3, 12):
+if sys.platform == 'win32':
```
oscarbenjamin, line 9: We need < 3.12 on Windows as well because the numpy.distutils module is removed in Python 3.12. I guess it could be:
```
if sys.platform == 'win32' and sys.version_info < (3, 12)
```

mkoeppe: but does the case `sys.platform == 'win32'` and `sys.version_info >= (3, 12)` work at all?

oscarbenjamin: Yes, it works. Every CI job normally passes but the change here has caused the Windows job to fail when building wheels for Python 3.12. These same CI jobs are what build the wheels that go to PyPI which has Windows wheels for Python 3.12:
https://pypi.org/project/python-flint/#files

Some sort of change in CPython itself means that mingw+setuptools works for Python 3.12 but not for Python < 3.12. I don't follow all the details but the clearest explanation I found is here:
https://github.com/cython/cython/issues/4470#issuecomment-1361506709

mkoeppe:
Thanks. I've updated it accordingly

oscarbenjamin:
Looks like its working. Thanks Matthias.

oscarbenjamin:
The Windows build with cibuildwheel is now failing CI in gh-113. I'm not sure if this or other changes is the cause.

mkoeppe:
https://github.com/flintlib/python-flint/actions/runs/7748025255/job/21130728629?pr=113#step:5:81
"Invalid --only='""', must be a build selector with a known platform"

oscarbenjamin:
I don't see why that error should be thrown now and only on Windows. The CI passes after merging this PR:
https://github.com/flintlib/python-flint/actions/runs/7672043828

I've just restarted that same CI it to see if it passes again and it has now failed.

The cibuildwheel version does not seem to have changed. The msys version does not seem to have changed. The Python versions have not changed.

The same argument `--only '""'` is passed for all other OS but the job only fails on Windows.

oscarbenjamin:
Maybe an upstream bug:
https://github.com/pypa/cibuildwheel/issues/1748

oscarbenjamin:
I think it is working after bumping the version of cibuildwheel to 2.16.5. Apparently the problem was a change in the GitHub Actions runners that needed an update in the pypa/cibuildwheel action.

PR #114, opened 2024-02-01
test CI failure

oscarbenjamin:
CI failed in gh-113 and I just want to check if a do nothing PR passes CI...


oscarbenjamin:
Okay it fails in the same way so the problem is not due to the changes in gh-113.

PR #119, opened 2024-02-02
Add prerelease testing for CPython 3.13 to CI

oscarbenjamin:
Add prerelease testing for CPython 3.13 to CI

It would also be good to add a job that builds flint from git trunk...

### DIFF
--- a/.github/workflows/buildwheel.yml
+++ b/.github/workflows/buildwheel.yml
@@ -137,17 +137,17 @@ jobs:
       - run: python -m flint.test --verbose
 
   test_flint_versions:
-    name: pip install ${{ matrix.target }} on ${{ matrix.python-version }}
+    name: Test flint ${{ matrix.flinttag }}
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
         # minimum supported version and latest git
-        flint: ['v3.0.0', 'main']
+        flinttag: ['v3.0.0', 'main']
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
           python-version: 3.12
-      - run: bin/pip_install_ubuntu.sh . ${{ matrix.flint }}
+      - run: bin/pip_install_ubuntu.sh . ${{ matrix.flinttag }}
       - run: python -m flint.test --verbose
--- a/.github/workflows/buildwheel.yml
+++ b/.github/workflows/buildwheel.yml
@@ -125,7 +125,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.11', '3.12', '3.13-dev']
-        # '.' means install from git checkout
+        # '.' means install from python-flint git checkout
         # 'python-flint' means install from PyPI sdist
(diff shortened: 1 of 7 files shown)


oscarbenjamin:
Looks like building against flint main fails for `fmpz_mod_mat`:
```
src/flint/types/fmpz_mod_mat.c:4860:5: error: too few arguments to function ‘fmpz_mod_mat_clear’
       4860 |     fmpz_mod_mat_clear(__pyx_v_self->val);
            |     ^~~~~~~~~~~~~~~~~~
      In file included from src/flint/types/fmpz_mod_mat.c:1227:
      /usr/local/include/flint/fmpz_mod_mat.h:48:6: note: declared here
         48 | void fmpz_mod_mat_clear(fmpz_mod_mat_t mat, const fmpz_mod_ctx_t ctx);
            |      ^~~~~~~~~~~~~~~~~~
```
Maybe there are incompatible changes to the signatures of functions that operate with `fmpz_mod_mat`...

oscarbenjamin:
Looks like this was changed in https://github.com/flintlib/flint/pull/1754

fredrik-johansson:
> Looks like this was changed in [flintlib/flint#1754](https://github.com/flintlib/flint/pull/1754)

Yes, this will be a notable compatibility-breaking change in 3.1. There could be others, but hopefully nothing as major.

oscarbenjamin:
I don't think it would be hard to change the `fmpz_mod_mat` code in python-flint for the change. Attempting to support both versions of Flint is possibly awkward but perhaps this is a good moment to see how well an `#ifdef` approach can work.

oscarbenjamin:
The python-flint fmpz_mod_mat type already stores a reference to an associated
`fmpz_mod_ctx`. If I understand correctly the change is mostly just that almost
all `fmpz_mod_mat` functions in flint now need this ctx to be passed explicitly
as an additional parameter.

We can probably do something like:
```cython
# fmpz_mod_mat.pxd
cdef extern from *:
    """
    /*
     * fmpz_mod_mat function signatures were changed in flint 3.1.0
    */
    #if __FLINT_RELEASE >= 30100
    void fmpz_mod_mat_clear_compat(fmpz_mod_mat_t mat, fmpz_mod_ctx_t ctx) {
        fmpz_mod_mat_clear(mat, ctx);
    }
    #else
    void fmpz_mod_mat_clear_compat(fmpz_mod_mat_t mat, fmpz_mod_ctx_t ctx) {
        fmpz_mod_mat_clear(mat);
    }
    #endif
    """

cdef extern from "flint/fmpz_mod_mat.h":
    void fmpz_mod_mat_clear_compat(fmpz_mod_mat_t mat, fmpz_mod_ctx_t ctx)

# fmpz_mod_mat.pyx
cdef class fmpz_mod_mat:
    ...
    def __dealloc__(self):
        fmpz_mod_mat_clear_compat(self.mat, self.ctx)
```

oscarbenjamin:
Okay, all sorted and works with latest Flint from git as well as previous versions.

PR #123, opened 2024-02-02
pyproject.toml: Move settings of CIBW_* here

mkoeppe:
As suggested in https://github.com/flintlib/python-flint/pull/108#issuecomment-1909901282 @oscarbenjamin 

Also increasing build parallelism in CI because GH Actions upgraded to machines with 4 CPUs in late 2023. (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). (`make -j6` was a first guess, reducing the wall time of "Build wheels for macos-12" from 28 min to 17 min and "Build wheels for windows-2019" from 28 min to 26 min.)

### DIFF
--- a/.github/workflows/buildwheel.yml
+++ b/.github/workflows/buildwheel.yml
@@ -34,7 +34,6 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.16.5
         env:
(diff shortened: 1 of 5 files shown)


oscarbenjamin:
Looks good. Thanks Matthias!

PR #125, opened 2024-02-25
maint: bump Flint to 3.1.0

oscarbenjamin:
Flint 3.1.0 is released:

https://groups.google.com/g/flint-devel/c/36-dOb0Sfqc/m/By2w0cs5AQAJ?utm_medium=email&utm_source=footer

### DIFF
--- a/.github/workflows/buildwheel.yml
+++ b/.github/workflows/buildwheel.yml
@@ -129,8 +129,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


oscarbenjamin:
The 3.13-dev builds fail due to https://github.com/python/cpython/issues/114626 which is already fixed in https://github.com/python/cpython/pull/115561 so it should be fine after the next alpha release of CPython 3.13. See also https://github.com/cython/cython/issues/4635#issuecomment-1963022775.

PR #129, opened 2024-03-20
Use meson-python instead of setuptools for wheels

oscarbenjamin:
This is just for wheel building for now and not building the dependencies.

Currently this works except on Windows and also there is a bug with meson-python and python 3.13.


oscarbenjamin:
> there is a bug with meson-python and python 3.13.

That bug is described in https://github.com/mesonbuild/meson/issues/12401. It will be resolved somehow before Python 3.13 is released so it is not a significant problem.

oscarbenjamin:
Okay, this is working on Windows now. 

It is still not working with `Flint < 3.10` for some reason...

oscarbenjamin:
Since GitHub did not create the automatic back-reference this is where the rest of the discussion happened about these changes: https://github.com/mesonbuild/meson-python/discussions/602

oscarbenjamin:
It seems that `PKG_CONFIG_PATH` is not inherited by msys2 in CI and that msys2 initialises `PKG_CONFIG_PATH` to add some mingw paths. Setting the environment variable in cibuildwheel config in pyproject.toml ensures that it gets set but overrides the mingw paths. Somehow just not setting `PKG_CONFIG_PATH` at all seems to work on Windows in CI.

oscarbenjamin:
> It seems that `PKG_CONFIG_PATH` is not inherited by msys2 in CI

https://www.msys2.org/docs/pkgconfig/
https://github.com/msys2/setup-msys2/issues/171

oscarbenjamin:
I've opened another discussion in meson-python about editable installs on OSX: https://github.com/mesonbuild/meson-python/discussions/604

oscarbenjamin:
The main outstanding problem here is just that building with `Flint < 3.1.0` fails when using pkgconfig. I think that this is due to a bug that was fixed by https://github.com/flintlib/flint/pull/1647.

oscarbenjamin:
> I think that this is due to a bug that was fixed by [flintlib/flint#1647](https://github.com/flintlib/flint/pull/1647).

I can confirm that fixing the Flint 3.0.1 flint.pc pkgconfig file fixes this problem. The question then is just whether python-flint should drop support for `Flint < 3.0.1` or if it would be better to add something to the meson build config that can fix this.

PR #135, opened 2024-05-08
build: add --local-flint to set rpath in dev build

oscarbenjamin:
This implements https://github.com/mesonbuild/meson/issues/13046#issuecomment-2099453043.

On Linux the current meson build needs `LD_LIBRARY_PATH` to be set when using a local build of flint. This adds a project option to install the rpaths and also customises spin's build command so that you can do:
```
spin build --local-flint=.local/lib/pkgconfig
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build/*
+build-install/*
 dist/*
 src/flint/**/*.c
 src/flint/*.html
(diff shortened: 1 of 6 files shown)


oscarbenjamin:
I've removed the custom spin command. To set the rpaths you can use:
```
spin build -- -Dpkg_config_path=$(pwd)/.local/lib/pkgconfig -Dadd_flint_rpath=true
```

PR #139, opened 2024-06-01
maint: install wheel explicitly for cibuildwheel on Windows

oscarbenjamin:
This is needed for cibuildwheel >= 2.18.0 since wheel is no longer installed implicitly.


oscarbenjamin:
See gh-138

PR #142, opened 2024-06-14
Exposing acb_theta_all

edgarcosta:
The only thing missing is that I do not know how to handle flint versions older than 3.1.


oscarbenjamin:
We could put this in a separate module that is built conditionally:
```diff
diff --git a/meson.build b/meson.build
index 7e941db..150e9f6 100644
--- a/meson.build
+++ b/meson.build
@@ -22,6 +22,9 @@ endif
 # flint.pc was missing -lflint until Flint 3.1.0
 if flint_dep.version().version_compare('<3.1')
   flint_dep = cc.find_library('flint')
+  have_acb_theta = true
+else
+  have_acb_theta = false
 endif
 
 pyflint_deps = [dep_py, gmp_dep, mpfr_dep, flint_dep]
diff --git a/src/flint/types/acb_mat.pyx b/src/flint/types/acb_mat.pyx
index 533bdc7..fff3531 100644
--- a/src/flint/types/acb_mat.pyx
+++ b/src/flint/types/acb_mat.pyx
@@ -20,6 +20,11 @@ from flint.flintlib.arf cimport *
 from flint.flintlib.acb cimport *
 from flint.flintlib.acb_mat cimport *
 
+try:
+    from .acb_theta import acb_mat_theta
+except ImportError:
+    acb_mat_theta = None
+
 cimport cython
 
 cdef acb_mat_coerce_operands(x, y):
@@ -814,3 +819,8 @@ cdef class acb_mat(flint_mat):
         if left:
             return Elist, L
         return Elist, R
+
+   def theta(tau, z, square=False):
+       if acb_mat_theta is None:
+           raise NotImplementedError("acb.theta() needs Flint >= 3.1.0")
+       ...
diff --git a/src/flint/types/meson.build b/src/flint/types/meson.build
index bf8caec..4fd6362 100644
--- a/src/flint/types/meson.build
+++ b/src/flint/types/meson.build
@@ -41,6 +41,10 @@ exts = [
   'fmpz_mpoly',
 ]
 
+if have_acb_theta
+  exts = exts + ['acb_theta']
+endif
+
 py.install_sources(
   pyfiles,
   pure: false,
```

edgarcosta:
I will try it out.

edgarcosta:
Thank you. That sped up things. Hopefully, all the tests pass now.
I kept the doctests on `acb_mad_theta` as having conditional doctests is a pain, and we should only have them in one place.

src/flint/test/__main__.py
```diff
@@ -80,6 +80,8 @@ def run_doctests(verbose=None):
                flint.types.acb_series,
                flint.types.dirichlet,
                flint.functions.showgood]
+    if hasattr(flint.types, 'acb_theta'):
+        modules.append(flint.types.acb_theta)
```
edgarcosta, line 84: I must be doing something wrong, as I only get 2816 doctests...

oscarbenjamin: `hasattr` will only work if the module was already imported. I think this needs a try/import.

oscarbenjamin:
```diff
diff --git a/src/flint/test/__main__.py b/src/flint/test/__main__.py
index 1bd13de..f1dcaf0 100644
--- a/src/flint/test/__main__.py
+++ b/src/flint/test/__main__.py
@@ -80,8 +80,11 @@ def run_doctests(verbose=None):
                flint.types.acb_series,
                flint.types.dirichlet,
                flint.functions.showgood]
-    if hasattr(flint.types, 'acb_theta'):
-        modules.append(flint.types.acb_theta)
+    try:
+        from flint.types import acb_theta
+        modules.append(acb_theta)
+    except ImportError:
+        pass
     results = [doctest.testmod(x) for x in modules]
 #    ffmpz, tfmpz = doctest.testmod(flint._fmpz, verbose=verbose)
  #   failed, total = doctest.testmod(flint.pyflint, verbose=verbose)
```

oscarbenjamin:
I think that `doc/source/acb_theta.rst` is needed as well.

oscarbenjamin:
The docs can be built with `spin docs`.

oscarbenjamin:
Looks good. Thanks!

oscarbenjamin:
We should add an automatically generated interface that makes it possible to call any Flint function (gh-54). Then it would be easier to access something like `acb_theta` without needing any changes in python-flint itself.

PR #147, opened 2024-06-19
Run strip on libflint.so after building Flint

oscarbenjamin:
Not sure if this might break anything but I expect it to make the wheels a lot smaller.


oscarbenjamin:
gh-148 was better.

Issue #124, opened 2024-02-13
Crash in `nmod_poly`


GiacomoPope:
For `fmpz_mod_poly` we get an error when we cannot compute an inverse when performing operations

```py
>>> from flint import fmpz_mod_poly_ctx, fmpz_mod_poly
>>> R = fmpz_mod_poly_ctx(10)
>>> f = R([1,2,3])
>>> f
3*x^2 + 2*x + 1
>>> f % f
0
>>> g = R([3,1,2])
>>> g
2*x^2 + x + 3
>>> g % g
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "src/flint/types/fmpz_mod_poly.pyx", line 659, in flint.types.fmpz_mod_poly.fmpz_mod_poly.__mod__
  File "src/flint/types/fmpz_mod_poly.pyx", line 652, in flint.types.fmpz_mod_poly.fmpz_mod_poly._mod_
ValueError: Cannot compute remainder of 2*x^2 + x + 3 modulo 2*x^2 + x + 3
```

But for `nmod_poly` we get a crash:

```py
Python 3.11.4 (main, Jul 25 2023, 17:07:07) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from flint import *
>>> f = nmod_poly([2,1,3], 10)
>>> f
3*x^2 + x + 2
>>> f % f
0
>>> f = nmod_poly([3, 1, 2], 10)
>>> f % f
Flint exception (Impossible inverse):
    Cannot invert modulo 2*5
zsh: abort      python3
```

oscarbenjamin:
I think what is needed is to have contexts for nmod so that there is somewhere to store whether or not the modulus is prime. Then operations that are not well defined for composite moduli could be rejected.

GiacomoPope:
Yeah I agree. I think this would also help users simply because then different polynomial objects would be used in the same way.

oscarbenjamin:
I can see potentially two different kinds of contexts. The low-level version is sort of what `fmpz_mod_poly_ctx` is except that it would be good to make it so that all objects have such a context and can quickly compare them in methods like `__add__` or can quickly check for prime modulus.

A higher level version could be a user interface like:
```python
R = Zmod(10)
n = R(5)
p = R.poly([1, 2, 1])
M = R.matrix([[1, 2], [3, 4]])
s = R.series(...)
```
[...]

PR #164, opened 2024-07-18
Fmpz mod mpoly and nmod mpoly

Jake-Moss:
This PR aims to add support for the `fmpz_mod_mpoly` and `nmod_mpoly` flint objects. I'm using the recently added `fmpz_mpoly` as a basis, there may be remants of the large copy-paste that have escaped my changes. 

I haven't given the docs any more attention that a quick search-replace as of yet so while they (and all tests locally) may be passing I'll have to give them a once over before I start on `nmod_mpoly`.


Jake-Moss:
For now I've crammed the `fmpz_mod_mpoly` into the mpoly tests with both a prime and non-prime modulus. Though it's not quite as clean as Id like. The division functions assume the modulus is prime, so I've set it to raise an exception on any division function with a non-prime modulus, even if it be within the domain. The same goes for sqrt and gcd. While the flint docs don't state they it's requires a prime modulus for gcd, it was raising an impossible inversion error so I added the exception to it as well.

oscarbenjamin:
> I've set it to raise an exception on any division function with a non-prime modulus, even if it be within the domain. The same goes for sqrt and gcd.

That seems right to me.

Since `nmod_mpoly` has a context (unlike other `nmod_*` types; see gh-124) we could also check whether the modulus is prime when the context is created.

Jake-Moss:
> Since `nmod_mpoly` has a context (unlike other `nmod_*` types; see [gh-124](https://github.com/flintlib/python-flint/issues/124)) we could also check whether the modulus is prime when the context is created.

I think that's also reasonable, currently I'm just checking the first time, storing the result and just returning that instead for later calls.

oscarbenjamin:
> I think that's also reasonable, currently I'm just checking the first time, storing the result and just returning that instead for later calls.

Either approach is reasonable I think. Not doing it when the context is created saves the one time cost of the check if it is never needed. In exchange we get a slight bit of overhead each time the attribute needs to be checked. I don't know whether that overhead is even measurable though.

What would be measurable I think is actually computing the prime check every time division happens. Particularly in the case of `nmod` rather than `nmod_mpoly` I would expect that to bring a measurable slowdown. This is why we need a context for `nmod` so that we can store whether or not the modulus is prime and only do the prime check once.

Jake-Moss:
> This is why we need a context for nmod so that we can store whether or not the modulus is prime and only do the prime check once.

That sounds reasonable, is that something we'd like in this PR?

I've brought the nmod_mpoly up to the same spec as fmpz_mod_mpoly. I've refrained from cramming them into the other nmod_poly and fmpz_mod_poly tests respectively just because of large API differences. Happy to have a crack at it but would either require a little more than just adding a for loop + some lambdas.

Whoops I looks like I missed that numpy isn't required here, was just using it for 32bit/64bit dependent memoryviews in place of a malloc of ulongs.

oscarbenjamin:
> > This is why we need a context for nmod so that we can store whether or not the modulus is prime and only do the prime check once.
> 
> That sounds reasonable, is that something we'd like in this PR?

That's up to you. If it makes things easier in this PR then go ahead. Otherwise we can leave that to a separate PR.

More generally I think that we want all types to have associated contexts so that there is a uniform interface like `ZZ.poly([1,2])` or `GF(3).poly([1,2])` and we can express operations like `ZZ.poly([1,2]).set_domain(QQ)`. That means we want contexts for fmpz, fmpq, arb, ...

We would also need contexts for all types for generic rings as well. I haven't looked enough into the generic rings interfaces to see if that should affect the overall design of contexts for the non-generic types.

For arb we need a contexts that can hold non-global precision. Likewise for the number of terms in a series. It isn't possible for SymPy, mpmath etc to make use of things like `arf` if the precision is global (e.g. SymPy and mpmath would both clobber each other). Even methods like `__repr__` depend on global variables so there needs to be some sort of printer context or something.

If you want to add contexts for `nmod` then I would suggest just keeping it simple so that it resembles `fmpz_mod`. I mention these other things here just to give a sense for where I would want to go in future.

oscarbenjamin:
The test failure is
```
  File "/opt/hostedtoolcache/Python/3.12.4/x64/lib/python3.12/site-packages/flint/test/__main__.py", line 36, in run_tests
    test()
  File "/opt/hostedtoolcache/Python/3.12.4/x64/lib/python3.12/site-packages/flint/test/test_all.py", line 2888, in test_mpolys
    assert P(ctx=ctx).is_zero() is True
           ^^^^^^^^^^^^^^^^^^
AttributeError: 'flint.types.fmpz_mod_mpoly.fmpz_mod_mpoly' object has no attribute 'is_zero'
```
I think that is a test added in gh-168 so you may need to merge master to see the same failure locally.

Jake-Moss:
> I think that is a test added in [gh-168](https://github.com/flintlib/python-flint/pull/168) so you may need to merge master to see the same failure locally.

Yep that was it thanks! Seems like CI was doing something and merging the PR before testing, definitely through me off for a minute.

> If you want to add contexts for nmod then I would suggest just keeping it simple so that it resembles fmpz_mod. I mention these other things here just to give a sense for where I would want to go in future.

I think something like that might be best suited in it's own PR, this one only touches on `nmod_mpoly` and `fmpz_mod_mpoly`

Jake-Moss:
> More generally I think that we want all types to have associated contexts so that there is a uniform interface like `ZZ.poly([1,2])` or `GF(3).poly([1,2])` and we can express operations like `ZZ.poly([1,2]).set_domain(QQ)`. That means we want contexts for fmpz, fmpq, arb, ...

This seems like it could be done as system that builds upon the existing contexts. It would have to be Python types rather than Cython types as `cdef class`s don't support multiple inheritance if something OO driven is desired.

> We would also need contexts for all types for generic rings as well. I haven't looked enough into the generic rings interfaces to see if that should affect the overall design of contexts for the non-generic types.

Just from a quick skim they seem like quite the beast.

> For arb we need a contexts that can hold non-global precision. Likewise for the number of terms in a series. It isn't possible for SymPy, mpmath etc to make use of things like `arf` if the precision is global (e.g. SymPy and mpmath would both clobber each other). Even methods like `__repr__` depend on global variables so there needs to be some sort of printer context or something.

Would this be replacing the global `FlintContext`? Instead moving it towards something instance specific? Or at least swappable?

oscarbenjamin:
> Seems like CI was doing something and merging the PR before testing, definitely through me off for a minute.

That is usually how CI works at least for Travis and GitHub Actions. We were using Cirrus as well before which did not seem to work like that.

When you push to a PR (or close and open it) the CI system will make a temporary branch somewhere effectively like:
```console
git checkout master
git checkout -b tmp_branch
git merge pr_branch # merge the PR into temporary copy of master
<run CI checks>
```
Sometimes you need to close and open a PR to get it to merge with latest master for CI (asking GA to "rerun" a job does not trigger a new merge).

There are also a bunch of complicated access rules in GitHub so if the PR modifies the GA workflow files then it might still run with the master version of the workflows depending on the event type that triggers the workflow.

Jake-Moss:
Looks like another possible memory error. I'll hunt that down tomorrow.

oscarbenjamin:
> Would this be replacing the global `FlintContext`? Instead moving it towards something instance specific?

These are details that need to be worked out. For `arf` for example I would like to use it for SymPy's RR domain: https://github.com/sympy/sympy/pull/26861. It also makes sense for `mpmath` to use `arf` at least as its internal representation for `mpf`. That would mean that both SymPy and mpmath start manipulating the global precision and would mess each other up. Likewise the global precision is not thread-safe even if SymPy is the only user of it.

The thread-safe part could be managed by using a thread-local variable in python-flint. I don't know what the overheads would be for that in the context of `arf` though. I still think SymPy and mpmath would clobber each other though because it is likely that SymPy wants to modify the precision exactly when it is doing arithmetic with mpf or calling mpmath functions. Using a thread-local but otherwise global precision would still require very careful usage of context managers like `with ctx.extraprec()` everywhere and that just seems fragile.

The cleanest solution is to use separate contexts with context-local precision etc settings. Then you can do e.g.:
```python
ctx = arb_ctx(53)
result = ctx.sin(1)
```
Then SymPy and mpmath can each make their own contexts and manipulate the precision separately. Of course if SymPy makes this context a global variable and modifies its precision then we're back to not thread-safe etc.

In the case of something like the RR domain though we can attach a separate context to each domain object and we never need to modify its precision. Ideally creating this `arb_ctx` would be a very cheap operation so that it can be say the first step in a call to something like `.evalf()`. Then everything is naturally thread-safe and robust since we've just avoided having any global variables to worry about.

The problem with having separate arb contexts everywhere is that then you need to do e.g. `ctx.add(x, y)` rather than `x + y`. For many purposes that is fine but it's not really nice for end users and wouldn't work for SymPy's `RR` domain where the elements need to work in generic code that uses elementary arithmetic. Now then what you need is to have each `arb`, `arf` etc hold a reference to the `arb_ctx`. Then in arithmetic like `x + y` the `x.__add__` method can check the context for the other object and if they have the same context then its precision can be used for the operation and the returned result can also hold a reference to the same context. 

Then we could ensure that any `arb` created as e.g. `flint.arb(2)` just ends up using the global context. That way users who just want to set the precision globally can do so but it doesn't mess up SymPy or mpmath who would never use the global context directly.

Now the problem is what are you going to do when you have `x + y` and they have different contexts with different precision or even different rounding modes? Let's see what mpmath does:
```python
In [1]: import mpmath

In [2]: ctx1 = mpmath.MPContext()

In [3]: ctx2 = mpmath.MPContext()

In [4]: ctx2.dps = 50

In [5]: ctx1.mpf(1) / ctx2.mpf(3)
Out[5]: mpf('0.33333333333333331')

In [6]: ctx2.mpf(1) / ctx1.mpf(3)
Out[6]: mpf('0.33333333333333333333333333333333333333333333333333311')
```
It basically just takes the context from the left-hand operand. I think that it would be better to say something like:

- If there are mixed context operands then take the one with higher precision (since dropping precision is irreversible)?
- Definitely error if they have different rounding modes though.
- Maybe actually just make all mixed context operations an error and require an explicit conversion or the use of context methods like `x1+ctx1(x2)` or `ctx1.add(x1,x2)`.

There are likely two very different groups of users here:

- End users who are probably happy using a global context. These users probably want implicit conversions.
- Library users like SymPy, mpmath etc who are probably happy using explicit conversions. From SymPy's perspective it is nicer for debugging if mixed context operations give an immediate error so that we can fix it and add the explicit conversion in the right place.

Perhaps a way to satisfy both groups is to say that implicit coercions between contexts are allowed only when mixing the global context with a non-global context (the global context can coerce all others). Then SymPy can never use the global context but end users who do use it can take the output of SymPy functions and mix it into their own types.

Similar considerations apply to all of the other operations like the number of terms in a series how exactly everything gets printed. The problem with printing is you might have something like:
```
result = str([obj1, obj2])
```
Now the list's `__str__` method is going to call `obj1.__repr__()`. Adding `obj1.repr()` doesn't help us here. This is basically the same problem as with `x + y`. One of two things is needed:

- Either we have a printer object that takes care of printing lists and things as well so you do `printer.str([obj1, obj2])`. Now `__repr__` and `__str__` are not used at all.
- Or otherwise each object needs to somehow hold a reference to a context that holds its printing settings but that adds to the memory footprint of every object.

Again the ideal solution here is probably different for different groups of users. SymPy for example already has its own complicated printing machinery so being able to call `printer.str(...)` or `obj.repr()` works fine. Many end users will find that inconvenient though and may want something more like the current global variables.

src/flint/test/test_all.py
```diff
@@ -2668,19 +2668,46 @@ def setbad(obj, i, val):
...
+        (
+            flint.nmod_mpoly,
+            lambda *args, **kwargs: flint.nmod_mpoly_ctx.get_context(*args, **kwargs, modulus=101),
+            int,
+            True,
+        ),
+        (
+            flint.nmod_mpoly,
+            lambda *args, **kwargs: flint.nmod_mpoly_ctx.get_context(*args, **kwargs, modulus=100),
+            int,
```
oscarbenjamin, line 2694: The type here should be `nmod` rather than `int`. The tests should also check for interoperability with `int` as well.

oscarbenjamin:
There should be some handling equality checking against scalars:
```python
In [26]: flint.fmpz_poly([1]) == 1
Out[26]: True

In [27]: ctx = flint.fmpz_mpoly_ctx(2, flint.Ordering.lex, ['x', 'y'])

In [28]: ctx.from_dict({(0,0):1})
Out[28]: 1

In [29]: ctx.from_dict({(0,0):1}) == 1
Out[29]: False
```

src/flint/test/test_all.py
Jake-Moss: I've just pushed some changes that add a lot more interop between `mpoly` types and their base counter part, that should resolve this

Jake-Moss:
> Looks like another possible memory error. I'll hunt that down tomorrow.

I believe this is still lurking, I don't expect CI to pass on windows. I'll try to get to this soon

Jake-Moss:
This took a while to catch but I believe it's been fixed. The `ctypedef unsigned long ulong` in `flintlib/flint.pxd` causes Cython to inline the definition of `ulong` in place of the `sizeof(ulong)`. This is the incorrect size on 64bit windows machines causing the wrong amount of memory to be allocated. Because it was such a small amount (only 8 bytes) in most tests the failures were sporadic while testing, mainly only failing on the freeing of the memory.

Jake-Moss:
@GiacomoPope just added missing argument in https://github.com/flintlib/python-flint/pull/164/commits/f15b55a29f68d9493a13ca8990a157df4262294e. I don't think the fix is worth a separate PR

GiacomoPope:
Thanks for fixing this!

Jake-Moss:
My local CI via `act` is not working, hopefully that's all that's need for the coverage

oscarbenjamin:
> This took a while to catch but I believe it's been fixed. The `ctypedef unsigned long ulong` in `flintlib/flint.pxd` causes Cython to inline the definition of `ulong` in place of the `ulong` in the `sizeof(ulong)`

Good work!

Should we remove the `ctypedef unsigned long ulong` line?

I guess it is not really accurate and so could maybe cause other problems as well...

At Cython time we do not know if we are building for Windows or not so I guess that `ulong` and `slong` should be treated as being opaque types. Maybe that would be awkward because it might prevent Cython from automatically casting e.g. `ulong` to Python int when returning from a function. I'd be happier with explicit casts than obscure bugs though...

We have this which fixed previous Windows bugs:
https://github.com/flintlib/python-flint/blob/f5b07e958b6b22ea4ef4d7e40a3d6cf42a342bba/src/flint/flintlib/flint.pxd#L63-L76

src/flint/flintlib/flint.pxd
```diff
@@ -60,6 +60,11 @@ cdef extern from "flint/flint.h":
     long flint_get_num_threads()
     void flint_cleanup()
 
+    ctypedef struct nmod_t:
+        mp_limb_t n
+        mp_limb_t ninv
+        flint_bitcnt_t norm
```
oscarbenjamin, line 66: This should also be removed from `nmod.pxd`.

Jake-Moss:
> Should we remove the `ctypedef unsigned long ulong` line?
> 
> I guess it is not really accurate and so could maybe cause other problems as well...
> 
> At Cython time we do not know if we are building for Windows or not so I guess that `ulong` and `slong` should be treated as being opaque types. Maybe that would be awkward because it might prevent Cython from automatically casting e.g. `ulong` to Python int when returning from a function. I'd be happier with explicit casts than obscure bugs though...

I think its preferable to remove it or use the true definition, though I'm unsure how that would be done. I'm sure removing will cause the Cython compiler to complain. I'll try some things and see if anything sticks tomorrow morning, I'm currently away from my desktop.

oscarbenjamin:
> I think its preferable to remove it or use the true definition, though I'm unsure how that would be done

I think the only way to use the true definition is to do the `extern from '*'` thing that I just showed. This is because in principle cython runs to generate the C code and then the C code gets moved to another computer to be compiled. In practice that doesn't really happen much but basically what it means is that we can't check OS, Flint version etc at Cython compile time: only the C compiler can do that.

At least defining it as `long long` would be more correct on 64 bit systems. Maybe we should add a CI job somewhere to check that python-flint works on at least some 32 bit OS...

oscarbenjamin:
> It seems it's possible to declare the types as opaque structs, Cython seems happy with this, however almost every single usage of them breaks.

Okay that's not acceptable I think.

I've asked how to handle this on the Cython mailing list:
https://groups.google.com/g/cython-users/c/sw1D4DdJI0U

Leaving aside the ctypedef ulong issue where are we at with this?

Is it just waiting for review?

Jake-Moss:
> > It seems it's possible to declare the types as opaque structs, Cython seems happy with this, however almost every single usage of them breaks.
> 
> Okay that's not acceptable I think.

Yep I agree. Thanks, I'll keep an eye on that list, I'm interested to see if there's a blessed solution.

> Leaving aside the ctypedef ulong issue where are we at with this?
> 
> Is it just waiting for review?

I believe so, it should be very similar to the `fmpz_mpoly` and `fmpq_mpoly` PR. Although I've made no attempts to address this comment of yours https://github.com/flintlib/python-flint/pull/164#issuecomment-2254563573, it's certainly something that should be done but I'm not sure where everything is currently standing with the other recent changes.

src/flint/test/test_all.py
```diff
@@ -2981,7 +3024,7 @@ def quick_poly():
                 (0, 1): 6
             })
 
-        for T in [int, S, flint.fmpz, lambda x: P(x, ctx=ctx)]:
+        for T in [int, S, int, lambda x: P(x, ctx=ctx)]:
```
oscarbenjamin, line 3027: This is checking `int` twice. Should it not check `fmpz`?

src/flint/test/test_all.py
```diff
@@ -2930,14 +2962,18 @@ def quick_poly():
 
         assert +quick_poly() \
             == quick_poly()
+
         assert -quick_poly() \
-            == mpoly({(0, 0): -1, (0, 1): -2, (1, 0): -3, (2, 2): -4})
+            == mpoly(
+                {(0, 0): -1, (0, 1): -2, (1, 0): -3, (2, 2): -4} if P is not flint.nmod_mpoly
+                else {k: ctx.modulus() + v for k, v in {(0, 0): -1, (0, 1): -2, (1, 0): -3, (2, 2): -4}.items()}
+        )
```
oscarbenjamin, line 2970: I would generally simplify these large expressions and bring out the conditional part to top level like:
```python
d = {(0, 0): -1, (0, 1): -2, (1, 0): -3, (2, 2): -4}
if P is not flint.nmod_mpoly:
    assert -quick_poly() == mpoly(d)
else:
    assert -quick_poly() == mpoly({m: v + ctx.modulus() for m, v in d.items()})
```
Does the `nmod_mpoly` constructor not take care of this automatically though?

I just checked:
```python
In [8]: ctx = nmod_mpoly_ctx(2, Ordering.lex, ['x', 'y'], 11)

In [9]: ctx.from_dict({(1, 2): 3})
Out[9]: 3*x*y^2

In [10]: ctx.from_dict({(1, 2): -3})
OverflowError                             Traceback (most recent call last)
Cell In[10], line 1
----> 1 ctx.from_dict({(1, 2): -3})

File nmod_mpoly.pyx:224, in flint.types.nmod_mpoly.nmod_mpoly_ctx.from_dict()

OverflowError: can't convert negative value to ulong
```
That should be fixed to be like other nmod types:
```python
In [11]: nmod_poly([0, -3], 11)
Out[11]: 8*x

In [12]: nmod(-3, 11)
Out[12]: 8

In [13]: nmod_mat([[1, 2], [3, -3]], 11)
Out[13]: 
[1, 2]
[3, 8]
```

src/flint/test/test_all.py
```diff
@@ -2930,14 +2962,18 @@ def quick_poly():
...
 
         assert quick_poly() \
             + mpoly({(0, 0): 5, (0, 1): 6, (1, 0): 7, (2, 2): 8}) \
             == mpoly({(0, 0): 6, (0, 1): 8, (1, 0): 10, (2, 2): 12})
 
-        for T in [int, S, flint.fmpz, lambda x: P(x, ctx=ctx)]:
+        for T in [int, S, lambda x: P(x, ctx=ctx)]:
             p = quick_poly()
             p += T(1)
```
oscarbenjamin, line 2978: Does `fmpz` not always work here?
```python
In [15]: ctx.from_dict({(1, 2): 3}) * fmpz(3)
TypeError                                 Traceback (most recent call last)
Cell In[15], line 1
----> 1 ctx.from_dict({(1, 2): 3}) * fmpz(3)

TypeError: unsupported operand type(s) for *: 'flint.types.nmod_mpoly.nmod_mpoly' and 'flint.types.fmpz.fmpz'
```
If the exception is only for `nmod_mpoly` then I think it would be better to write that explicitly in the test:
```python
        for T in [int, S, lambda x: P(x, ctx=ctx)]:
            p = quick_poly()
            if type(p) is not nmod_mpoly:
                ...
```
Really though I think that we want `fmpz` and `int` to be compatible with all types (see also gh-178):
```python
In [16]: nmod_poly([1, 2], 3) * fmpz(2)
Out[16]: x + 2

In [17]: nmod_poly([1, 2], 3) + fmpz(2)
Out[17]: 2*x

In [15]: ctx.from_dict({(1, 2): 3}) * fmpz(3)
TypeError                                 Traceback (most recent call last)
Cell In[15], line 1
----> 1 ctx.from_dict({(1, 2): 3}) * fmpz(3)

TypeError: unsupported operand type(s) for *: 'flint.types.nmod_mpoly.nmod_mpoly' and 'flint.types.fmpz.fmpz'
```

src/flint/test/test_all.py
```diff
@@ -2995,65 +3038,89 @@ def quick_poly():
...
+            assert quick_poly() // S(1) == quick_poly()
+            assert quick_poly() % S(1) == P(ctx=ctx)
+            assert divmod(quick_poly(), S(1)) == (quick_poly(), P(ctx=ctx))
 
         if is_field:
-            assert quick_poly() / 3 == mpoly({(0, 0): S(1, 3), (0, 1): S(2, 3), (1, 0): S(1), (2, 2): S(4, 3)})
+            if (P is flint.fmpz_mod_mpoly or P is flint.nmod_mpoly):
+                assert quick_poly() / 3 == mpoly({(0, 0): S(34), (0, 1): S(68), (1, 0): S(1), (2, 2): S(35)})
+            else:
+                assert quick_poly() / 3 == mpoly({(0, 0): S(1, 3), (0, 1): S(2, 3), (1, 0): S(1), (2, 2): S(4, 3)})
```
oscarbenjamin, line 3071: It should be possible to use e.g. `S(1)/3` here and not need to split the two cases.

src/flint/test/test_all.py
```diff
@@ -3075,19 +3142,29 @@ def quick_poly():
         # # XXX: Not sure what this should do in general:
         assert raises(lambda: pow(P(1, ctx=ctx), 2, 3), NotImplementedError)
 
-        if is_field:
-            assert (f * g).gcd(f) == f / 4
+        if (P is not flint.fmpz_mod_mpoly and P is not flint.nmod_mpoly) or f.context().is_prime():
```
oscarbenjamin, line 3145: Give the number of places where this condition is checked it should probably have a name like:
```
composite_characteristic = (P in (fmpz_mod_mpoly, nmod_mpoly)) and not P.is_prime()
```

src/flint/types/fmpz_mod_mpoly.pyx
```diff
@@ -0,0 +1,1192 @@
...
+                fmpz_mod_mpoly_push_term_fmpz_ffmpz(res.val, (<fmpz>o).val, exp_vec.val, self.val)
+
+        fmpz_mod_mpoly_sort_terms(res.val, self.val)
+        return res
+
+
+cdef class fmpz_mod_mpoly(flint_mpoly):
+    """
+    The *fmpz_mod_mpoly* type represents sparse multivariate polynomials over
+    the integers.
```
oscarbenjamin, line 249: Over the integers mod `n` for large `n` or something like that.

src/flint/types/fmpz_mod_mpoly.pyx
```diff
@@ -0,0 +1,1192 @@
...
+            res = create_fmpz_mod_mpoly(self.ctx)
+            res2 = create_fmpz_mod_mpoly(self.ctx)
+            fmpz_mod_mpoly_divrem(res.val, res2.val, (<fmpz_mod_mpoly>other).val, (<fmpz_mod_mpoly>self).val, res.ctx.val)
+            return (res, res2)
+        return NotImplemented
+
+    def __floordiv__(self, other):
+        cdef fmpz_mod_mpoly res
+        if not self.ctx.is_prime():
+            raise DomainError("division with non-prime modulus is not supported")
```
oscarbenjamin, line 518: This applies to other cases as well. We need to check the type of `other` and return `NotImplemented` before checking anything else. It is important to return `NotImplemented` rather than raise a `TypeError` because that is how Python's cooperative dispatch works for binary operators. If we return `NotImplemented` here then it gives a chance for the other type to implement `__rfloordiv__`. Other future types might (in principle) want to define this operation with this type for whatever reason even if the modulus is not prime.

src/flint/types/fmpz_mod_mpoly.pyx
```diff
@@ -0,0 +1,1192 @@
...
+            raise ZeroDivisionError("fmpz_mod_mpoly division by zero")
+        o = any_as_fmpz(other)
+        if o is NotImplemented:
+            return NotImplemented
+        res = create_fmpz_mod_mpoly(self.ctx)
+        fmpz_mod_mpoly_set_fmpz(res.val, (<fmpz>o).val, self.ctx.val)
+        return res / self
+
+    def __mod__(self, other):
+        return divmod(self, other)[1]
```
oscarbenjamin, line 597: It is not quite right to return `divmod(...)` here because it means that we never return `NotImplemented` which in turn means that the other type cannot override this with `__rmod__`.

src/flint/types/fmpz_mod_mpoly.pyx
```diff
@@ -0,0 +1,1192 @@
...
+        """
+        Factors self into irreducible factors, returning a tuple
+        (c, factors) where c is the content of the coefficients and
+        factors is a list of (poly, exp) pairs.
+
+            >>> from flint import Ordering
+            >>> Zm = fmpz_mod_mpoly
+            >>> ctx = fmpz_mod_mpoly_ctx.get_context(3, Ordering.lex, 11, 'x,y,z')
+            >>> p1 = Zm("2*x + 4", ctx)
+            >>> p2 = Zm("3*x*z +  + 3*x + 3*z + 3", ctx)
```
oscarbenjamin, line 992: Is this meant to have ` +  + `?

src/flint/types/nmod_mpoly.pyx
```diff
@@ -0,0 +1,1276 @@
...
+        elif other is None:
+            return op == Py_NE
+        elif typecheck(self, nmod_mpoly) and typecheck(other, nmod_mpoly):
+            if (<nmod_mpoly>self).ctx is (<nmod_mpoly>other).ctx:
+                return (op == Py_NE) ^ <bint>nmod_mpoly_equal(self.val, (<nmod_mpoly>other).val, self.ctx.val)
+            else:
+                return op == Py_NE
+        elif typecheck(other, nmod):
+            if other.modulus() != self.ctx.modulus():
+                raise ValueError(f"cannot compare with different modulus {other.modulus()} vs {self.ctx.modulus()}")
```
oscarbenjamin, line 295: This should return False rather than raising an exception.

src/flint/types/nmod_mpoly.pyx
```diff
@@ -0,0 +1,1276 @@
...
+        return <bint>nmod_mpoly_is_zero(self.val, self.ctx.val)
+
+    def is_one(self):
+        return <bint>nmod_mpoly_is_one(self.val, self.ctx.val)
+
+    def __richcmp__(self, other, int op):
+        if not (op == Py_EQ or op == Py_NE):
+            return NotImplemented
+        elif other is None:
+            return op == Py_NE
```
oscarbenjamin, line 287: It shouldn't be necessary to check for `None` explicitly here.

src/flint/types/nmod_mpoly.pyx
```diff
@@ -0,0 +1,1276 @@
...
+            raise ValueError("not enough arguments provided")
+        elif nargs > nvars:
+            raise ValueError("more arguments provided than variables")
+
+        args = [int(x) for x in args]
+
+        cdef:
+            # Using sizeof(ulong) here breaks on 64 windows machines because of the `ctypedef unsigned long ulong` in
+            # flintlib/flint.pxd. Cython will inline this definition and then allocate the wrong amount of memory.
+            ulong *vals = <ulong *>libc.stdlib.malloc(nargs * (FLINT_BITS // 4))
```
oscarbenjamin, line 699: Should this not be `FLINT_BITS // 8`?

oscarbenjamin:
> I've asked how to handle this on the Cython mailing list:
> https://groups.google.com/g/cython-users/c/sw1D4DdJI0U

The suggested solution from Cython maintainer is to do:
```
cdef extern from *:
    """
    #define SIZEOF_ULONG sizeof(ulong)
    """
    int SIZEOF_ULONG
```
That is marginally better than using `FLINT_BITS // 8`.

It is also apparently considered to be a Cython bug that the code for `sizeof(ulong)` is not generated correctly. Apparently it should be okay to have an extern typedef like:
```
cdef extern from "flint/flint.h":
   ctypedef unsigned long ulong
   ctypedef long slong
```
Cython should be expected to respect the signed-ness of the types but otherwise not make any assumptions about the actual size or whether it really is typedef'd to `unsigned long` in the C code.

It doesn't look like we actually use `sizeof(ulong)` or `sizeof(slong)` anywhere else so hopefully there are no other problems lurking as a result of this.

src/flint/test/test_all.py
Jake-Moss: It certainly should be, whoops

src/flint/test/test_all.py
Jake-Moss: It should, just an oversight. I'm adding `fmpz` support now, commits will come tomorrow evening

src/flint/test/test_all.py
Jake-Moss: Yep it is thanks. This also revealed that I wasn't used `fmpz_mod` for the `fmpz_mod_mploy` tests, which has opened another rabbit whole of things to add.

src/flint/test/test_all.py
```diff
@@ -2995,65 +3038,89 @@ def quick_poly():
...
         g = mpoly({(0, 1): 2, (1, 0): 2})
-        assert f * g / mpoly({(0, 1): 1, (1, 0): 1}) \
-            == mpoly({(1, 1): 8, (0, 0): 2})
-
-        if not is_field:
-            assert raises(lambda: 1 / quick_poly(), DomainError)
-            assert raises(lambda: quick_poly() / P(2, ctx=ctx), DomainError)
+        if (P is flint.fmpz_mod_mpoly or P is flint.nmod_mpoly) and not ctx.is_prime():
+            pass
+        else:
```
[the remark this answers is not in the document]
Jake-Moss: They do not when `ctx.is_prime` is false, division is not even attempted there. The type check is just to short-circuit because the others don't have an `is_prime` method.

src/flint/types/nmod_mpoly.pyx
Jake-Moss: It certainly should, whoops! But I'll favour the suggested fix over this

src/flint/types/nmod_mpoly.pyx
```diff
@@ -0,0 +1,1451 @@
...
+        elif typecheck(other, nmod):
+            if other.modulus() != self.ctx.modulus():
+                raise ValueError(f"cannot multiply with different modulus {other.modulus()} vs {self.ctx.modulus()}")
+            res = create_nmod_mpoly(self.ctx)
+            nmod_mpoly_scalar_mul_ui(res.val, (<nmod_mpoly>self).val, int(other), self.ctx.val)
+            return res
+        elif isinstance(other, int):
+            res = create_nmod_mpoly(self.ctx)
+            nmod_mpoly_scalar_mul_ui(res.val, (<nmod_mpoly>self).val, other, res.ctx.val)
+            return res
```
oscarbenjamin, line 535: Converting the `int` to `ulong` here can overflow e.g.:
```python
In [1]: from flint import *

In [2]: ctx = nmod_mpoly_ctx(2, Ordering.lex, ['x', 'y'], 11)

In [3]: p = ctx.from_dict({(1, 2): 3})

In [4]: p
Out[4]: 3*x*y^2

In [5]: p * 100**100
OverflowError                             Traceback (most recent call last)
Cell In[5], line 1
----> 1 p * 100**100

File flint/types/nmod_mpoly.pyx:534, in flint.types.nmod_mpoly.nmod_mpoly.__mul__()

OverflowError: Python int too large to convert to C unsigned long
```
It also doesn't work for negative numbers:
```
In [6]: -1 * p
OverflowError                             Traceback (most recent call last)
Cell In[6], line 1
----> 1 -1 * p

File flint/types/nmod_mpoly.pyx:543, in flint.types.nmod_mpoly.nmod_mpoly.__rmul__()

OverflowError: can't convert negative value to ulong
```
I would say that the common case is a small integer and so it is good to use `nmod_mpoly_scalar_mul_ui` for the fast-path but it can be something like:
```python
cdef ulong o
try:
    o = other
except OverflowError:
    o = other % mod
```
There could perhaps a be a `cdef inline` function if this is something that is needed a lot. It could be a method on the context.

There is a lot of repetitive code in different methods like `__add__` for handling all of these different cases. Maybe we need a context method like `ctx.any_scalar_as_ulong(other)` that could handle the `int`, `fmpz`, `nmod`, `fmpz_mod` cases all in one.

Actually I'm not sure about `fmpz_mod`... It seems to be not supported for mixing:
```
In [1]: from flint import *

In [2]: ctx = fmpz_mod_ctx(127)

In [3]: nmod_poly([1, 2], 3) * ctx(2)
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 nmod_poly([1, 2], 3) * ctx(2)

TypeError: unsupported operand type(s) for *: 'flint.types.nmod_poly.nmod_poly' and 'flint.types.fmpz_mod.fmpz_mod'

In [4]: ctx = fmpz_mod_poly_ctx(127)

In [5]: ctx([1, 2]) * nmod(3, 127)
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 ctx([1, 2]) * nmod(3, 127)

TypeError: unsupported operand type(s) for *: 'flint.types.fmpz_mod_poly.fmpz_mod_poly' and 'flint.types.nmod.nmod'
```
I guess we can continue with that for `mpoly` as well.

In reality there isn't much benefit in mixing them because you should always just use `nmod` for smaller moduli and `fmpz_mod` for larger. Particularly in Python it would probably have been better if these were merged together somehow like `fq_default`.

oscarbenjamin:
Tests are failing but only under Python 3.12. Specifically this test:
```
            assert p.subs({"x1": S(0), 0: S(0)}) == ctx.from_dict({(0, 0): 1})
>           assert p.compose(p.subs({"x1": 0}), ctx.from_dict({(0, 1): 1})) == mpoly({
                (2, 2): 36,
                (1, 2): 24,
                (1, 0): 9,
                (0, 2): 4,
                (0, 1): 2,
                (0, 0): 4
            })
E           assert 64*x0^4*x1^2 ...^2 + 2*x1 + 10 == 36*x0^2*x1^2 ...1^2 + 2*x1 + 4
```
The output is definitely quite different from that expected in the test:
```
(Pdb) type(p)
<class 'flint.types.fmpz_mod_mpoly.fmpz_mod_mpoly'>
(Pdb) p p.compose(p.subs({"x1": 0}), ctx.from_dict({(0, 1): 1}))
64*x0^4*x1^2 + 96*x0^3*x1^2 + 31*x0^2*x1^2 + 12*x0^2 + 72*x0*x1^2 + 9*x0 + 36*x1^2 + 2*x1 + 10
```
It looks like subs is the problem:
```
(Pdb) p p
4*x0^2*x1^2 + 3*x0 + 2*x1 + 1
(Pdb) p p.subs({"x1":0})
4*x0^2 + 3*x0 + 3
```
If we set `x1 -> 0` then we should just get `3*x0 + 1` here. It seems to be setting `x1 -> 1` instead.

The `fmpz_mod_mpoly.subs` method is here:
https://github.com/flintlib/python-flint/blob/043ca9dd9d047c78af7fe4ca8a01ee0ffa9341de/src/flint/types/fmpz_mod_mpoly.pyx#L894-L925

src/flint/types/fmpz_mod_mpoly.pyx
```diff
@@ -0,0 +1,1321 @@
...
+
+        fmpz_mod_mpoly_set(res.val, self.val, self.ctx.val)
+        for i, arg in args:
+            if typecheck(arg, fmpz_mod):
+                fmpz_mod_mpoly_evaluate_one_fmpz(res.val, res.val, i, (<fmpz_mod>arg).val, self.ctx.val)
+            else:
+                arg_fmpz = any_as_fmpz(arg)
+                if arg_fmpz is NotImplemented:
+                    raise TypeError(f"cannot coerce {type(arg)} to fmpz")
+                fmpz_mod_mpoly_evaluate_one_fmpz(res.val, res.val, i, (<fmpz>arg).val, self.ctx.val)
```
oscarbenjamin, line 923: I think it should be `(<fmpz>arg_fmpz).val` here.

Otherwise this is casting a PyLong to fmpz I guess which doesn't work.

The tests pass under Python 3.12 if this is changed to `arg_fmpz`.

Jake-Moss: It certainly should be, thanks.

src/flint/types/nmod_mpoly.pyx
Jake-Moss: > Converting the `int` to `ulong` here can overflow e.g.:
> It also doesn't work for negative numbers:

This is largely intentional, though through the absence of handling rather than an explicit case. I can see why we'd want it to work as well.

> There is a lot of repetitive code in different methods like `__add__` for handling all of these different cases. Maybe we need a context method like `ctx.any_scalar_as_ulong(other)` that could handle the int, fmpz, nmod, fmpz_mod cases all in one.

I completely agree, it's turning into a huge mess and I'm not particularly happy with it but haven't opted to rip it all out in favour of something smarter yet. I certainly can, and am now considering if I could complete it today.

Just enumerating the type conversion cases for a binary operator on something like `nmod_mpoly` we have
- `int`, try casting to `ulong`
  - No overflow,  `ui` functions
  - Overflow, `int` -> `fmpz` -> `ulong` via `fmpz_get_nmod`, `ui` functions. Or `int` mod modulus in python -> `ulong`, `ui` functions
- `nmod` -> `ulong` via `.val`, ui functions
- `fmpz` -> `ulong` via `fmpz_get_nmod`, `ui` functions
- `fmpz_mod` -> `ulong` via `fmpz_get_nmod`, `ui` functions
- `nmod_mpoly`, `nmod_mpoly` functions

Some functions like `div` require going further to construct a `nmod_mpoly` and set it to a `ulong`. 

and for `fmpz_mod_mpoly`
- `int` -> `fmpz`, `fmpz` functions
- `nmod` -> `ulong` via `.val`, `ui` functions, however, flint just converts the `ulong` to a `fmpz` immediately
- `fmpz`, `fmpz` functions 
- `fmpz_mod` -> `fmpz` via `.val`, `fmpz` functions
- `fmpz_mod_mpoly`, `fmpz_mod_mpoly` functions

As it stand, the `_any_as_self` system is a step to far with respect to the optimal paths for various types. We could have a `ctx.any_as_scalar` and `ctx.scalar_as_mpoly` or similar, which depending on the type either returns the respective scalar/mpoly, provided the scalars used in the C functions are homogeneous . This can also be extended to the `fmpq_mpoly` type as well since flint just converts arguments of the `fmpz` functions to `fmpq`s immediately, so we can as well.

I'm not sure of timezones (Brisbane, UTC+10 here) but I may just go ahead and implement the above for this PR, if it works I'll port `fmpz_mpoly` and `fmpq_mpoly` over in another.

oscarbenjamin: For `nmod_mpoly` in particular is it not a bit simpler to manage scalar conversion?

We basically always want to get down to ulong (or `mp_limb_t` or something) and in the case of overflow we can use `x % mod` and then cast. This works because we know that `x % mod` will always fit the basic C type (since `mod` does).

Jake-Moss: > For `nmod_mpoly` in particular is it not a bit simpler to manage scalar conversion?

It is indeed. I'd prefer a generic solution that works for most if not all cases across the various types and I'm writing one now. I'll have some commits up soon. 
If we have one scalar type for each of the mpoly classes then a similar approach should work for everything, given that flint seems to convert all the different types into a single one, e.g. `ulong`s into `fmpz`s for `fmpz_mod_mpoly` and `fmpq`s for `fmpq_mpoly` for example

> ```c
> // flint/src/fmpz_mod_mpoly/add_fmpz.c
> void fmpz_mod_mpoly_add_ui(
>     fmpz_mod_mpoly_t A,
>     const fmpz_mod_mpoly_t B,
>     ulong c,
>     const fmpz_mod_mpoly_ctx_t ctx)
> {
>     fmpz_t cc;
>     fmpz_init(cc);
>     fmpz_mod_set_ui(cc, c, ctx->ffinfo);
>     fmpz_mod_mpoly_add_fmpz_mod(A, B, cc, ctx);
>     fmpz_clear(cc);
> }
> 
> // flint/src/fmpq_mpoly/add_fmpq.c
> void fmpq_mpoly_add_ui(fmpq_mpoly_t A, const fmpq_mpoly_t B,
>                                            ulong c, const fmpq_mpoly_ctx_t ctx)
> {
>     fmpq_t t;
>     fmpz_init_set_ui(fmpq_numref(t), c);
>     *fmpq_denref(t) = 1;
>     fmpq_mpoly_add_fmpq(A, B, t, ctx);
>     fmpz_clear(fmpq_numref(t));
> }
> ```

I think we can as well, which will dramatically simplify things

Jake-Moss: I've just pushed an example with `nmod_mpoly`s addition and divmod here https://github.com/flintlib/python-flint/pull/164/commits/9abc8d21a8de41788083ece5896e7deb115ffb21

src/flint/test/test_all.py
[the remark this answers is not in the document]
oscarbenjamin: But should there not be some tests in the `pass` block?

Like `assert raises(..., lambda: quick_poly() / 2)`?

Also presumably at least the `TypeError` cases below apply equally whether the modulus is prime or not.

src/flint/test/test_all.py
```diff
@@ -3093,19 +3146,29 @@ def quick_poly():
...
-        assert raises(lambda: quick_poly().gcd(None), TypeError)
-        assert raises(lambda: quick_poly().gcd(P(ctx=ctx1)), IncompatibleContextError)
+            assert raises(lambda: (f * g).gcd(f), DomainError)
 
-        assert (f * g).factor() == (S(2), [(mpoly({(0, 1): 1, (1, 0): 1}), 1), (f, 1)])
+        if P is flint.fmpz_mod_mpoly or P is flint.nmod_mpoly:
+            if is_field:
+                assert (f * g).factor() == (S(8), [(mpoly({(0, 1): 1, (1, 0): 1}), 1), (f / 4, 1)])
+        else:
+            assert (f * g).factor() == (S(2), [(mpoly({(0, 1): 1, (1, 0): 1}), 1), (f, 1)])
```
oscarbenjamin, line 3163: If factor would raise an exception for non-prime modulus then that should be tested here so that the test shows the behaviour in all cases.

Also can these not be written with `g` in the output?
```python
        characteristic_zero = not (P is flint.fmpz_mod_mpoly or P is flint.nmod_mpoly)

        if characteristic_zero:
            # Primitive factors over Z for fmpz_mpoly and fmpq_mpoly
            assert (f * g).factor() == (S(2), [(g / 2, 1), (f, 1)])
        elif is_field:
            # Monic polynomials over Z/pZ for nmod_mpoly and fmpz_mod_mpoly
            assert (f * g).factor() == (S(8), [(g / 2, 1), (f / 4, 1)])
        else:
            # Factorisation not allowed over Z/nZ for n not prime.
            # Flint would abort so we raise an exception instead:
            assert raises(...)
```
One thing that I hadn't noticed until now is that `fmpq_mpoly.factor` behaves differently compared to `fmpq_poly.factor`. With `fmpq_poly` factors are always monic:
```python
In [29]: f = fmpq_poly([1, 4])

In [30]: g = fmpq_poly([2, 2])

In [31]: f
Out[31]: 4*x + 1

In [32]: g
Out[32]: 2*x + 2

In [33]: (f*g).factor()
Out[33]: (8, [(x + 1/4, 1), (x + 1, 1)])
```
However `fmpq_mpoly` gives factors primitive over Z:
```python
In [35]: ctx = fmpq_mpoly_ctx(1, Ordering.lex, ["x"])

In [36]: f = fmpq_mpoly('4*x + 1', ctx)

In [37]: g = fmpq_mpoly('2*x + 2', ctx)

In [38]: (f*g).factor()
Out[38]: (2, [(x + 1, 1), (4*x + 1, 1)])
```
I'm wondering whether it makes sense to try to make this consistent somehow. One sort of consistency is that whenever we have a field all factors can be monic which would suggest that fmpq_mpoly should also return monic factors.

On the other hand though it is computationally better for Q to use primitive factors and absorb any denominator into the constant factor as fmpq_mpoly does so we could change `fmpq_poly` to return primitive factors instead:
https://github.com/flintlib/python-flint/blob/ba1a7fe20ce8a4798896219c449b2ebc381c625a/src/flint/types/fmpq_poly.pyx#L396-L416
In SymPy it is `fmpq_poly` that is inconsistent with SymPy's behaviour and needs to be special-cased to remove precisely the denominators that python-flint inserts:
https://github.com/sympy/sympy/blob/31e957f47a636df9b7e62cd3951ec0afcc266369/sympy/polys/polyclasses.py#L2285-L2304

oscarbenjamin:
I had a couple of comments about the tests but generally I think this looks good and could be merged.

Is there anything outstanding to do?

src/flint/test/test_all.py
Jake-Moss: > I'm wondering whether it makes sense to try to make this consistent somehow. One sort of consistency is that whenever we have a field all factors can be monic which would suggest that fmpq_mpoly should also return monic factors.
> 
> On the other hand though it is computationally better for Q to use primitive factors and absorb any denominator into the constant factor as fmpq_mpoly does so we could change fmpq_poly to return primitive factors instead:

In my mind I would expect python-flint to be largely transparent with the results it returns from flint, if flint is returning primitive factors then python-flint should as well, if not to just to be consistent. I've seen your #189 PR and will take a look now

Jake-Moss:
> Is there anything outstanding to do?

I don't believe so, nothing comes to mind. I will however open another PR to remove the `if isinstance ... elif typecheck` ladders from `fmpz_mpoly` and `fmpq_mpoly` as well once this is merged.

oscarbenjamin:
Okay, this looks good. Let's get it in!

oscarbenjamin:
This is nice because it is mote of the mpoly types from Flint now.

The only ones missing I think are `gr_mpoly` (all of `gr` is missing) and `fq_nmod_mpoly` (which makes less sense since we use `fq_default` rather than `fq_nmod`).

Certainly these are all of the mpoly types that SymPy would currently use until we add `gr`.

Issue #165, opened 2024-07-21
Segmentation fault when running tests in python-3.10 with 0.7.a2 and 0.7.0a3, 0.6.0 is fine


edgarcosta:
Here is an edited log. 

I couldn't try 0.7.a1, as it fails to find a wheel, and to build, as expected.

```
edgarcosta@diophantus ~/test ⚡ source bin/activate
(test) edgarcosta@diophantus ~/test ⚡ pip3 install python-flint
Collecting python-flint
  Downloading python_flint-0.6.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (41.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 41.1/41.1 MB 18.6 MB/s eta 0:00:00
Installing collected packages: python-flint
Successfully installed python-flint-0.6.0
(test) edgarcosta@diophantus ~/test ⚡ python3 -m flint.test
Running tests...
[...]
----------------------------------------
OK: Your installation of python-flint seems to be working just fine!
----------------------------------------
(test) edgarcosta@diophantus ~/test ⚡ pip3 install python-flint==0.7.0a1 # it can't build, as expected
Collecting python-flint==0.7.0a1
  Downloading python_flint-0.7.0a1.tar.gz (229 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 229.9/229.9 KB 4.5 MB/s eta 0:00:00
[...]
      ../meson.build:10:12: ERROR: Dependency "flint" not found, tried pkgconfig and make
[...]
(test) edgarcosta@diophantus ~/test 💩⚡ pip3 install python-flint==0.7.0a2
Collecting python-flint==0.7.0a2
  Downloading python_flint-0.7.0a2-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (10.6 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 10.6/10.6 MB 29.5 MB/s eta 0:00:00
Installing collected packages: python-flint
  Attempting uninstall: python-flint
    Found existing installation: python-flint 0.6.0
    Uninstalling python-flint-0.6.0:
      Successfully uninstalled python-flint-0.6.0
Successfully installed python-flint-0.7.0a2
[...]

PR #166, opened 2024-07-21
fix: set --host=x86_64-* when building Flint and MPFR

oscarbenjamin:
This is needed when building portable wheels for PyPI upload.

Hopefully fixes gh-165


oscarbenjamin:
@edgarcosta I've added `--disable-arch` here so the CI job currently running will produce wheels that I think should work. I don't have any way of testing them for now but it would be great if you could test them when you get teh chance.

edgarcosta:
The wheels generated here or in #170 work on my servers.

oscarbenjamin:
Thanks. I've merged #170

PR #169, opened 2024-07-22
add host to flint

edgarcosta:


edgarcosta:
I didnt see #166

oscarbenjamin:
> I didnt see #166

The two wheels that you said did not work came from the two commits of that PR.

PR #172, opened 2024-07-29
add fmpz_is_square

roos-j:


src/flint/types/fmpz.pyx
```diff
@@ -683,11 +683,30 @@ cdef class fmpz(flint_scalar):
...
+ 
+            >>> fmpz(81).is_perfect_power()
+            True
+            >>> fmpz(1234).is_perfect_power()
+            False
+        """
         cdef int k
         cdef fmpz v = fmpz()
         k = fmpz_is_perfect_power(v.val, self.val)
         return k != 0
```
oscarbenjamin, line 697: We can have both methods but more useful than `is_perfect_power` would be a method that returns the power like SymPy's `perfect_power` function:
```python
In [1]: perfect_power(81)
Out[1]: (3, 4)

In [2]: perfect_power(80)
Out[2]: False
```
I'm not sure what is best to return if it is not a perfect power. Returning False is perhaps a bit strange...

Alternatives:

1. Raise an exception.
2. Return clearly invalid number like `(0, -1)`.
3. Return a tuple with a boolean like `(3, 4, True)` or `(0, -1, False)`.
4. Return None.

Maybe returning `False` is good as anything else...

oscarbenjamin:
Can you add tests in `flint/tests/test_all.py`?

Doctests are good for some simple things but generally it is better to have comprehensive tests including for edge cases outside of the doctests. We want full test coverage from the test suite and then doctests are good for showing tested examples in the documentation rather than actually serving as the test suite.

I'm not sure about these ones:
```python
In [6]: flint.fmpz(-1).is_perfect_power()
Out[6]: True

In [7]: flint.fmpz(0).is_perfect_power()
Out[7]: True

In [8]: flint.fmpz(1).is_perfect_power()
Out[8]: True
```
Do we actually want that?

SymPy's perfect power rejects -1, 0 and 1. Wikipedia suggests that 0 and 1 are sometimes considered perfect powers but no mention of -1. It seems that flint's function like SymPy's accepts negative odd powers though so I guess it makes sense to have -1 if we also have e.g. -8.

src/flint/types/fmpz.pyx
```diff
@@ -683,11 +683,30 @@ cdef class fmpz(flint_scalar):
         return fmpz_is_probabprime(self.val)
 
     def is_perfect_power(self):
+        r"""
+        Return True if this integer is of the form `r^k`, False otherwise.
```
oscarbenjamin, line 687: We need to clarify what the convention is here. At least we need `k > 1` for this to be meaningful.

It seems to me that it is only really useful to say that a number is a perfect power if `r != self` and `k > 1` although that is not how flint's `is_perfect_power` function handles -1, 0 and 1.

src/flint/types/fmpz.pyx
oscarbenjamin: Or perhaps it can just be:
```python
>>> perfect_power(80)
(80, 1)
```
So `perfect_power(n)` always returns `(r, k)` such that `n = r^k` where `k` is either maximal or `1` if no maximal `k` exists:
```python
>>> perfect_power(1)
(1, 1)
```

roos-j: `is_perfect_power` was there before this PR, but when I saw it, I also thought it should best return `(r,k)` since that data is provided by the C function (in that case I would lean towards `None` if not a perfect power), so I'd be happy to change that.
Are you also suggesting it be renamed to `perfect_power`? That could make sense since one might expect a method called `is_something` to return a boolean. On the other hand, it seems the currently prevailing convention is to mirror the C library names, so maybe it should stay `is_perfect_power`?

oscarbenjamin: Sorry I missed the fact that the function was already there. I guess it is better not to change it then.

We could add a separate `.perfect_power()` method though to return `(r, k)`. The convention it seems in other places is that a perfect power should have `r,k>=2` although I guess that the `.perfect_power()` method should be consistent with `is_perfect_power()`.

src/flint/types/fmpz.pyx
roos-j: I agree that `k>1` may have to be added; it is also missing in the C library documentation, but I doubt that anyone has been confused because of it. Regarding -1, 0, 1: that is explicitly mentioned in the documentation, https://flintlib.org/doc/fmpz.html
I seems odd to deviate from the C library behavior in Python bindings.

oscarbenjamin: > I seems odd to deviate from the C library behavior in Python bindings.

I agree. Also the `is_perfect_power` method has been there a long time so it should not be changed without a strong reason.

I just have a hard time imagining that defining 0 and 1 to be perfect powers is particularly useful for anything. In practice the value of knowing that something is a perfect power is if it means that you can reduce a problem down to a smaller number. Finding perfect powers of negative numbers is also not really that useful because it is better to handle the sign separately and find a power representation for the absolute value:
```python
In [24]: perfect_power(-4)
Out[24]: False

In [25]: perfect_power(4)
Out[25]: (2, 2)
```

src/flint/types/fmpz.pyx
roos-j: Regarding tests, `fmpz_is_perfect_power` appears well-tested already: https://github.com/flintlib/flint/blob/main/src/fmpz/test/t-is_perfect_power.c

src/flint/types/fmpz.pyx
roos-j: Regarding tests, `fmpz_is_perfect_power` appears well-tested already: https://github.com/flintlib/flint/blob/main/src/fmpz/test/t-is_perfect_power.c

src/flint/types/fmpz.pyx
roos-j: I think that would be good. The issue is though that `(r,k)` returned by `fmpz_is_perfect_power` is not guaranteed to have maximal `k`; the right way to proceed would be to implement that feature in flint's `fmpz_is_perfect_power` and then use it for python-flint to implement a method `.perfect_power()` you are suggesting.

src/flint/types/fmpz.pyx
oscarbenjamin: The testing needed in python-flint is a bit different from the testing needed in Flint. Here in python-flint we can usually assume that what Flint does is mathematically correct so we don't need to exhaustively test different values.

Firstly we need at least one test just to see that python-flint does not segfault because of a bug in its own wrapper code.

Secondly though the problem is that Flint does things like segfaulting or aborting as a way of handling invalid inputs. For example it would not have surprised me if `fmpz(-8).is_perfect_power()` would just crash in much the same way that this does:
```python
In [28]: flint.nmod_poly([1, 2, 1], 4).factor()
    Cannot invert modulo 2*2
Aborted (core dumped)
```
We generally want to avoid exposing that core dump to Python users because it is not friendly if you are working e.g. in a notebook or something. That means that python-flint needs a bit more validation of arguments to raise exceptions before calling Flint. It also means that python-flint needs to test edge cases to make sure that Flint behaves reasonably for the values that are passed to Flint.

Lastly there can often be reasons to have different behaviour in python-flint as compared to Flint because the most natural/convenient way of handling edge cases might be different in Python.

src/flint/types/fmpz.pyx
oscarbenjamin: Ah, okay. I suppose we could just call `fmpz_is_perfect_power` in a loop until we've reduced down to a number that is not a perfect power.

Maybe returning a result with non-maximal `k` is useful. I would be more inclined to reduce as far as possible (i.e. find maximal `k`). That is what Sage's `perfect_power` apparently does:
https://doc.sagemath.org/html/en/reference/rings_standard/sage/rings/integer.html#sage.rings.integer.Integer.perfect_power

This seems like the sort of situation where what is best in python-flint might be a bit different from what Flint itself does. I'm not sure what the performance implications would be though of calling `fmpz_is_perfect_power` in a loop (with successively smaller arguments) compared to just calling it once.

src/flint/types/fmpz.pyx
roos-j: There is such a test for `is_perfect_power` [here](https://github.com/flintlib/python-flint/blob/591f900c2b8824c4e950a39afca33321b1c36192/src/flint/test/test_all.py#L295C34-L295C50)

Do you mean this should be expanded further?
Added test for `is_square`

oscarbenjamin: > Do you mean this should be expanded further?

No, I just forgot that `is_perfect_power` was already there.

oscarbenjamin:
Okay, looks good.

We can merge this once tests are complete unless you want to add a `.perfect_power()` method as well.

roos-j:
Thanks, I prefer to merge this now.

oscarbenjamin:
Great. Thanks!

Issue #73, opened 2023-09-07
nmod does not gracefully handle errors


GiacomoPope:
```py
>>> nmod_poly([12, 7, 1], 10)
x^2 + 7*x + 2
>>> nmod_poly([12, 7, 1], 10).factor()
Flint exception (Impossible inverse):
    Cannot invert modulo 2*0
zsh: abort      python3
```

oscarbenjamin:
In general I think that python-flint should avoid segfaults and aborts (except perhaps if it exposes a very low-level API).

There are some cases like this though where it is tricky because how can we know ahead of time that this will operation will or will not succeed?
```python
In [3]: nmod_poly([1, 1], 10).factor()
Out[3]: (1, [(x + 1, 1)])
```
If the modulus is prime it should always succeed. The Flint docs don't give any clue as to when this is expected to work or under what circumstances it might crash. At least python-flint should warn in its docs whether a particular function might abort because aborts are considered unfriendly in Python-land.

GiacomoPope:
In SageMath a common trick is with the `sig_on` and `sig_off` to check for issues, we could do something like that here?

oscarbenjamin:
> check for issues

What check could be done besides just warning if the modulus is not prime?

GiacomoPope:
Oh for the `sig_on()` and `sig_off()` I think it just prints the error from C without the abort? So you have:

```
sig_on()
... some dangerous code
sig_off()
```

If the modulus is prime, then every element is invertible, a softer check is just ensuring that there's no non-trivial gcd before an inversion is called?

oscarbenjamin:
Perhaps there could be an argument like `.factor(check=True)` or something.

PR #174, opened 2024-08-05
add compose_mod and powmod with large exp

GiacomoPope:
this PR adds two features which i need for a current project:

1) optimised polynomial composition when you want to reduce the result modulo another polynomial
2) f^e mod h when e is very large.

I'm a little rusty as I haven't thought about flint for a while, and there's one TODO for the `__pow__` test as now fmpz_mod_poly does not raise an assertion error when a modulus is given generically


src/flint/test/test_all.py
```diff
@@ -2635,9 +2662,10 @@ def setbad(obj, i, val):
         assert P([1, 1]) ** 2 == P([1, 2, 1])
         assert raises(lambda: P([1, 1]) ** -1, ValueError)
         assert raises(lambda: P([1, 1]) ** None, TypeError)
-
-        # # XXX: Not sure what this should do in general:
-        assert raises(lambda: pow(P([1, 1]), 2, 3), NotImplementedError)
+        
+        # XXX: Not sure what this should do in general:
+        # TODO: this now fails as fmpz_mod_poly allows modulus
+        # assert raises(lambda: pow(P([1, 1]), 2, 3), NotImplementedError)
```
oscarbenjamin, line 2668: If we want to use 3-arg pow with poly types then we should just change this test. I'm pretty sure that I added this test and comment just as part of increasing test coverage.

I don't know if this operation makes sense for all types. It looks like `fmpz_poly` and `fmpq_poly` only have `pow_trunc` rather than `powmod`.

Maybe it would make sense to allow only special cases like:
```python
x = fmpz_poly([0, 1])
r = pow(x + 1, 100, x**3)
r = pow(x + 1, 100, 3)
```
Perhaps a `pow_trunc` method would be a good method for all types.

GiacomoPope: I think Flint has these methods for many types, so it's definitely worth looking into. For example `fmpz_mod_poly` has `fmpz_mod_poly_pow_trunc` and `fmpz_poly` has `fmpz_poly_pow_trunc`

GiacomoPope: is this something you want to be fixed before merging this PR?

src/flint/types/nmod_poly.pyx
```diff
@@ -191,6 +198,115 @@ cdef class nmod_poly(flint_poly):
...
+            >>> f.inverse_series_trunc(4)
+            23*x^3 + 159*x^2 + 151*x + 110
+            >>> f.inverse_series_trunc(5)
+            45*x^4 + 23*x^3 + 159*x^2 + 151*x + 110
+        """
+        cdef nmod_poly res
+        res = nmod_poly.__new__(nmod_poly)
+        nmod_poly_init_preinv(res.val, self.val.mod.n, self.val.mod.ninv)
+        nmod_poly_inv_series(res.val, self.val, n)
+        return res
```
oscarbenjamin, line 252: More checking is needed here:
```python
In [3]: flint.nmod_poly(2, 3).inverse_series_trunc(-3)
Segmentation fault (core dumped)

In [17]: flint.nmod_poly(0, 3).inverse_series_trunc(3)
Flint exception (General error):
    Exception (nmod_poly_inv_series). Division by zero.
Aborted (core dumped)
```

GiacomoPope: Urgh, im tempted to remove the commit for nmod_poly and simply not touch the class until I have time to write a proper context object to do the necessary checks

GiacomoPope: From the docs we have:

> Given Q find Qinv such that Q * Qinv is 1 modulo . The constant coefficient of Q must be invertible modulo the modulus of Q. An exception is raised if this is not the case or if n = 0. This function can be viewed as inverting a power series.

So I guess we need something like

```
check = self.leading_coefficient().is_invertible()
```

Also we need `n != 0`. In fact, does allowing `n < 1` make sense in this context? Why does flint use an `slong` instead of a `ulong`...

src/flint/types/nmod_poly.pyx
```diff
@@ -355,17 +471,85 @@ cdef class nmod_poly(flint_poly):
...
+            >>> f.powmod(2**64, mod)
+            52*x^3 + 96*x^2 + 136*x + 9
+            >>> mod_rev_inv = mod.reverse().inverse_series_trunc(4)
+            >>> f.powmod(2**64, mod, mod_rev_inv)
+            52*x^3 + 96*x^2 + 136*x + 9
+        """
+        cdef nmod_poly res
+
+        if e < 0:
+            raise ValueError("Exponent must be non-negative")
```
oscarbenjamin, line 510: Would it make sense to use `inverse_mod` here?

Or maybe the caller should just do that if needed.

GiacomoPope: Yeah good question. I don't think we invert where e is negative elsewhere, but this is a reasonable thing to do.

src/flint/types/nmod_poly.pyx
```diff
@@ -355,17 +471,85 @@ cdef class nmod_poly(flint_poly):
...
-            raise NotImplementedError("nmod_poly modular exponentiation")
+            return self.powmod(exp, mod)
         if exp < 0:
             raise ValueError("negative exponent")
         res = nmod_poly.__new__(nmod_poly)
         nmod_poly_init_preinv(res.val, (<nmod_poly>self).val.mod.n, (<nmod_poly>self).val.mod.ninv)
         nmod_poly_pow(res.val, self.val, <ulong>exp)
         return res
 
+    def powmod(self, e, modulus, mod_rev_inv=None):
```
oscarbenjamin, line 485: Maybe `pow_mod` is more consistent as a name here along with `inverse_mod` and `compose_mod`.

src/flint/types/nmod_poly.pyx
GiacomoPope: i dont think we currently have the set up to fix worrying about the bad inversion here... It's something I noticed before #73 and #124 -- i think the whole class needs to have a huge refactor

src/flint/types/nmod_poly.pyx
GiacomoPope: ive done this

src/flint/types/nmod_poly.pyx
oscarbenjamin: As long as it works for prime modulus that is okay.

I think we just need to check here that `self` is not zero and `n` is not negative and also make it clear in code comment and/or docstring that the modulus needs to be prime.

In SymPy I have now disabled the use of python-flint for non-prime modulus because there are too many operations that fail. As you say a context is needed.

src/flint/test/test_all.py
oscarbenjamin: We should fix the test by checking for `nmod_poly` or `fmpz_mod_poly` and having a test that 3-arg `pow` works for them and raises `NotImplementedError` for other types.

We don't need to add methods to other types for this here and now. I just always want to consider what might make sense for the other types when designing any new interface/behaviour.

src/flint/types/nmod_poly.pyx
oscarbenjamin: Inverting does bring additional complications though because I think then we need the modulus to be irreducible.

src/flint/types/nmod_poly.pyx
```diff
@@ -191,6 +198,132 @@ cdef class nmod_poly(flint_poly):
...
+            >>> f = nmod_poly([123, 129, 63, 14, 51, 76, 133], 163)
+            >>> f.inverse_series_trunc(3)
+            159*x^2 + 151*x + 110
+            >>> f.inverse_series_trunc(4)
+            23*x^3 + 159*x^2 + 151*x + 110
+            >>> f.inverse_series_trunc(5)
+            45*x^4 + 23*x^3 + 159*x^2 + 151*x + 110
+        """
+        if n <= 0:
+            raise ValueError("n must be positive")
```
oscarbenjamin, line 263: We should check self.is_zero here as well.

GiacomoPope:
Hopefully addressed the above comments

oscarbenjamin:
Yep, looks good. Thanks

PR #179, opened 2024-08-09
fix(nmod): Add nmod_ctx to store is_prime

oscarbenjamin:
Work in progress...

Have nmod use a context object so that we can store the additional information that the modulus is prime to avoid operations that might fail for non-prime modulus (gh-124).

### DIFF
--- a/src/flint/types/nmod.pyx
+++ b/src/flint/types/nmod.pyx
@@ -12,31 +12,142 @@ from flint.flintlib.nmod_vec cimport *
 from flint.flintlib.fmpz cimport fmpz_fdiv_ui, fmpz_init, fmpz_clear
 from flint.flintlib.fmpz cimport fmpz_set_ui, fmpz_get_ui
 from flint.flintlib.fmpq cimport fmpq_mod_fmpz
-from flint.flintlib.ulong_extras cimport n_gcdinv
+from flint.flintlib.ulong_extras cimport n_gcdinv, n_is_prime
 
-cdef int any_as_nmod(mp_limb_t * val, obj, nmod_t mod) except -1:
-    cdef int success
-    cdef fmpz_t t
-    if typecheck(obj, nmod):
-        if (<nmod>obj).mod.n != mod.n:
-            raise ValueError("cannot coerce integers mod n with different n")
-        val[0] = (<nmod>obj).val
-        return 1
-    z = any_as_fmpz(obj)
-    if z is not NotImplemented:
-        val[0] = fmpz_fdiv_ui((<fmpz>z).val, mod.n)
-        return 1
-    q = any_as_fmpq(obj)
-    if q is not NotImplemented:
-        fmpz_init(t)
-        fmpz_set_ui(t, mod.n)
-        success = fmpq_mod_fmpz(t, (<fmpq>q).val, t)
-        val[0] = fmpz_get_ui(t)
-        fmpz_clear(t)
-        if not success:
-            raise ZeroDivisionError("%s does not exist mod %i!" % (q, mod.n))
-        return 1
-    return 0
+
+#cdef int any_as_nmod(mp_limb_t * val, obj, nmod_t mod) except -1:
+#    return mod.ctx.any_as_nmod(val, obj)
+
+
+_nmod_ctx_cache = {}
+
+
+cdef nmod_ctx any_as_nmod_ctx(obj):
+    """Convert an int to an nmod_ctx."""
+    if typecheck(obj, nmod_ctx):
+        return obj
+    if typecheck(obj, int):
+        ctx = _nmod_ctx_cache.get(obj)
+        if ctx is None:
+            ctx = nmod_ctx(obj)
+            _nmod_ctx_cache[obj] = ctx
+        return ctx
+    return NotImplemented
+
+
+cdef class nmod_ctx:
+    """
+    Context object for creating :class:`~.nmod` initalised 
+    with modulus :math:`N`.
+
+        >>> nmod_ctx(17)
+        nmod_ctx(17)
+
+    """
+    def __init__(self, mod):
+        cdef mp_limb_t m
+        m = mod
+        nmod_init(&self.mod, m)
+        self._is_prime = n_is_prime(m)
+
+    cdef int any_as_nmod(self, mp_limb_t * val, obj) except -1:
+        """Convert an object to an nmod element."""
+        cdef int success
+        cdef fmpz_t t
+        if typecheck(obj, nmod):
+            if (<nmod>obj).ctx != self:
+                raise ValueError("cannot coerce integers mod n with different n")
+            val[0] = (<nmod>obj).val
+            return 1
+        z = any_as_fmpz(obj)
+        if z is not NotImplemented:
+            val[0] = fmpz_fdiv_ui((<fmpz>z).val, self.mod.n)
+            return 1
+        q = any_as_fmpq(obj)
+        if q is not NotImplemented:
+            fmpz_init(t)
+            fmpz_set_ui(t, self.mod.n)
+            success = fmpq_mod_fmpz(t, (<fmpq>q).val, t)
+            val[0] = fmpz_get_ui(t)
+            fmpz_clear(t)
+            if not success:
+                raise ZeroDivisionError("%s does not exist mod %i!" % (q, self.mod.n))
+            return 1
+        return 0
+
+    def modulus(self):
+        """Get the modulus of the context.
+
+        >>> ctx = nmod_ctx(17)
+        >>> ctx.modulus()
+        17
+
+        """
+        return fmpz(self.mod)
+
+    def is_prime(self):
+        """Check if the modulus is prime.
+
+        >>> ctx = nmod_ctx(17)
+        >>> ctx.is_prime()
+        True
+
+        """
+        return self._is_prime
+
+    def zero(self):
+        """Return the zero element of the context.
+
+        >>> ctx = nmod_ctx(17)
+        >>> ctx.zero()
+        0
+
+        """
+        return self(0)
+
+    def one(self):
+        """Return the one element of the context.
+
+        >>> ctx = nmod_ctx(17)
+        >>> ctx.one()
+        1
+
+        """
+        return self(1)
+
+    def __hash__(self):
+        return hash(self.mod)
+
+    def __eq__(self, other):
+        if not typecheck(other, nmod_ctx):
+            return NotImplemented
+        else:
+            return self.mod == other.mod
+
+    def __str__(self):
+        return f"Context for nmod with modulus: {self.modulus()}"
+
+    def __repr__(self):
+        return f"nmod_ctx({self.modulus()})"
+
+    cdef nmod _new(self, mp_limb_t * val):
+        cdef nmod r = nmod.__new__(nmod)
+        r.val = val[0]
+        r.ctx = self
+        return r
+
+    def __call__(self, val):
+        """Create an nmod element from an integer.
+
+        >>> ctx = nmod_ctx(17)
+        >>> ctx(10)
+        10
+
+        """
+        cdef mp_limb_t v
+        v = val
+        return self._new(&v)
+
 
 cdef class nmod(flint_scalar):
     """
@@ -46,16 +157,15 @@ cdef class nmod(flint_scalar):
         3
 
     """
-
     def __init__(self, val, mod):
-        cdef mp_limb_t m
-        m = mod
-        nmod_init(&self.mod, m)
-        if not any_as_nmod(&self.val, val, self.mod):
+        ctx = any_as_nmod_ctx(mod)
+        if ctx is NotImplemented:
+            raise TypeError("Invalid context/modulus for nmod: %s" % mod)
+        if not ctx.any_as_nmod(&self.val, val):
             raise TypeError("cannot create nmod from object of type %s" % type(val))
 
     def repr(self):
-        return "nmod(%s, %s)" % (self.val, self.mod.n)
+        return "nmod(%s, %s)" % (self.val, self.ctx.mod.n)
 
     def str(self):
         return str(int(self.val))
@@ -64,7 +174,7 @@ cdef class nmod(flint_scalar):
         return int(self.val)
 
     def modulus(self):
-        return self.mod.n
+        return self.ctx.mod.n
 
     def __richcmp__(s, t, int op):
         cdef mp_limb_t v
@@ -73,13 +183,13 @@ cdef class nmod(flint_scalar):
             raise TypeError("nmods cannot be ordered")
         if typecheck(s, nmod) and typecheck(t, nmod):
             res = ((<nmod>s).val == (<nmod>t).val) and \
-                  ((<nmod>s).mod.n == (<nmod>t).mod.n)
+                  ((<nmod>s).ctx.mod.n == (<nmod>t).ctx.mod.n)
             if op == 2:
                 return res
             else:
                 return not res
         elif typecheck(s, nmod) and typecheck(t, int):
-            res = s.val == (t % s.mod.n)
+            res = s.val == (t % s.ctx.mod.n)
             if op == 2:
                 return res
             else:
@@ -97,100 +207,108 @@ cdef class nmod(flint_scalar):
 
     def __neg__(self):
         cdef nmod r = nmod.__new__(nmod)
-        r.mod = self.mod
-        r.val = nmod_neg(self.val, self.mod)
+        r.ctx = self.ctx
+        r.val = nmod_neg(self.val, self.ctx.mod)
         return r
 
     def __add__(s, t):
-        cdef nmod r
+        cdef nmod r, s2
         cdef mp_limb_t val
-        if any_as_nmod(&val, t, (<nmod>s).mod):
+        s2 = s
+        if s2.ctx.any_as_nmod(&val, t):
             r = nmod.__new__(nmod)
-            r.mod = (<nmod>s).mod
-            r.val = nmod_add(val, (<nmod>s).val, r.mod)
+            r.ctx = s2.ctx
+            r.val = nmod_add(val, s2.val, s2.ctx.mod)
             return r
         return NotImplemented
 
     def __radd__(s, t):
-        cdef nmod r
+        cdef nmod r, s2
         cdef mp_limb_t val
-        if any_as_nmod(&val, t, (<nmod>s).mod):
+        s2 = s
+        if s2.ctx.any_as_nmod(&val, t):
             r = nmod.__new__(nmod)
-            r.mod = (<nmod>s).mod
-            r.val = nmod_add((<nmod>s).val, val, r.mod)
+            r.ctx = s2.ctx
+            r.val = nmod_add(s2.val, val, s2.ctx.mod)
             return r
         return NotImplemented
 
     def __sub__(s, t):
-        cdef nmod r
+        cdef nmod r, s2
         cdef mp_limb_t val
-        if any_as_nmod(&val, t, (<nmod>s).mod):
+        s2 = s
+        if s2.ctx.any_as_nmod(&val, t):
             r = nmod.__new__(nmod)
-            r.mod = (<nmod>s).mod
-            r.val = nmod_sub((<nmod>s).val, val, r.mod)
+            r.ctx = s2.ctx
+            r.val = nmod_sub(s2.val, val, s2.ctx.mod)
             return r
         return NotImplemented
 
     def __rsub__(s, t):
         cdef nmod r
         cdef mp_limb_t val
-        if any_as_nmod(&val, t, (<nmod>s).mod):
+        s2 = s
+        if s2.ctx.any_as_nmod(&val, t):
             r = nmod.__new__(nmod)
-            r.mod = (<nmod>s).mod
-            r.val = nmod_sub(val, (<nmod>s).val, r.mod)
+            r.ctx = s2.ctx
+            r.val = nmod_sub(val, s2.val, s2.ctx.mod)
             return r
         return NotImplemented
 
     def __mul__(s, t):
-        cdef nmod r
+        cdef nmod r, s2
         cdef mp_limb_t val
-        if any_as_nmod(&val, t, (<nmod>s).mod):
+        s2 = s
+        if s2.ctx.any_as_nmod(&val, t):
             r = nmod.__new__(nmod)
-            r.mod = (<nmod>s).mod
-            r.val = nmod_mul(val, (<nmod>s).val, r.mod)
+            r.ctx = s2.ctx
+            r.val = nmod_mul(val, s2.val, s2.ctx.mod)
             return r
         return NotImplemented
 
     def __rmul__(s, t):
-        cdef nmod r
+        cdef nmod r, s2
         cdef mp_limb_t val
-        if any_as_nmod(&val, t, (<nmod>s).mod):
+        s2 = s
+        if s2.ctx.any_as_nmod(&val, t):
             r = nmod.__new__(nmod)
-            r.mod = (<nmod>s).mod
-            r.val = nmod_mul((<nmod>s).val, val, r.mod)
+            r.ctx = s2.ctx
+            r.val = nmod_mul(s2.val, val, s2.ctx.mod)
             return r
         return NotImplemented
 
     @staticmethod
     def _div_(s, t):
-        cdef nmod r
+        cdef nmod r, s2, t2
         cdef mp_limb_t sval, tval, x
-        cdef nmod_t mod
+        cdef nmod_ctx ctx
         cdef ulong tinvval
 
         if typecheck(s, nmod):
-            mod = (<nmod>s).mod
-            sval = (<nmod>s).val
-            if not any_as_nmod(&tval, t, mod):
+            s2 = s
+            ctx = s2.ctx
+            sval = s2.val
+            if not ctx.any_as_nmod(&tval, t):
                 return NotImplemented
         else:
-            mod = (<nmod>t).mod
-            tval = (<nmod>t).val
-            if not any_as_nmod(&sval, s, mod):
+            t2 = t
+            ctx = t2.ctx
+            tval = t2.val
+            if not ctx.any_as_nmod(&sval, s):
                 return NotImplemented
 
         if tval == 0:
-            raise ZeroDivisionError("%s is not invertible mod %s" % (tval, mod.n))
+            raise ZeroDivisionError("%s is not invertible mod %s" % (tval, ctx.mod.n))
         if not s:
             return s
 
-        g = n_gcdinv(&tinvval, <ulong>tval, <ulong>mod.n)
+        g = n_gcdinv(&tinvval, <ulong>tval, <ulong>ctx.mod.n)
         if g != 1:
-            raise ZeroDivisionError("%s is not invertible mod %s" % (tval, mod.n))
+            raise ZeroDivisionError("%s is not invertible mod %s" % (tval, ctx.mod.n))
 
         r = nmod.__new__(nmod)
-        r.mod = mod
-        r.val = nmod_mul(sval, <mp_limb_t>tinvval, mod)
+        r.ctx = ctx
+        r.val = nmod_mul(sval, <mp_limb_t>tinvval, ctx.mod)
         return r
 
     def __truediv__(s, t):
@@ -200,19 +318,22 @@ cdef class nmod(flint_scalar):
         return nmod._div_(t, s)
 
     def __invert__(self):
-        cdef nmod r
+        cdef nmod r, s
+        cdef nmod_ctx ctx
         cdef ulong g, inv, sval
-        sval = <ulong>(<nmod>self).val
-        g = n_gcdinv(&inv, sval, self.mod.n)
+        s = self
+        ctx = s.ctx
+        sval = <ulong>s.val
+        g = n_gcdinv(&inv, sval, ctx.mod.n)
         if g != 1:
-            raise ZeroDivisionError("%s is not invertible mod %s" % (sval, self.mod.n))
+            raise ZeroDivisionError("%s is not invertible mod %s" % (sval, ctx.mod.n))
         r = nmod.__new__(nmod)
-        r.mod = self.mod
+        r.ctx = ctx
         r.val = <mp_limb_t>inv
         return r
 
     def __pow__(self, exp, modulus=None):
-        cdef nmod r
+        cdef nmod r, s
         cdef mp_limb_t rval, mod
         cdef ulong g, rinv
 
@@ -223,8 +344,10 @@ cdef class nmod(flint_scalar):
         if e is NotImplemented:
             return NotImplemented
 
-        rval = (<nmod>self).val
-        mod = (<nmod>self).mod.n
+        s = self
+        ctx = s.ctx
+        rval = s.val
+        mod = ctx.mod.n
 
         # XXX: It is not clear that it is necessary to special case negative
         # exponents here. The nmod_pow_fmpz function seems to handle this fine
@@ -237,6 +360,6 @@ cdef class nmod(flint_scalar):
             e = -e
 
         r = nmod.__new__(nmod)
-        r.mod = self.mod
-        r.val = nmod_pow_fmpz(rval, (<fmpz>e).val, self.mod)
+        r.ctx = ctx
+        r.val = nmod_pow_fmpz(rval, (<fmpz>e).val, ctx.mod)
         return r
--- a/src/flint/types/nmod_poly.pyx
+++ b/src/flint/types/nmod_poly.pyx
@@ -4,7 +4,7 @@ from flint.utils.typecheck cimport typecheck
 from flint.types.fmpz cimport fmpz, any_as_fmpz
 from flint.types.fmpz_poly cimport any_as_fmpz_poly
 from flint.types.fmpz_poly cimport fmpz_poly
-from flint.types.nmod cimport any_as_nmod
+from flint.types.nmod cimport any_as_nmod_ctx
 from flint.types.nmod cimport nmod
 
 from flint.flintlib.nmod_vec cimport *
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,11 @@ classifiers = [
 file = "README.md"
 content-type = "text/markdown"
 
+[tool.meson-python.args]
+setup = [
+    "-Dadd_flint_rpath=true",
+]
+
 [tool.spin]
 package = "flint"
 
--- a/src/flint/types/nmod.pxd
+++ b/src/flint/types/nmod.pxd
@@ -2,8 +2,16 @@ from flint.flint_base.flint_base cimport flint_scalar
 from flint.flintlib.flint cimport mp_limb_t
 from flint.flintlib.nmod cimport nmod_t
 
-cdef int any_as_nmod(mp_limb_t * val, obj, nmod_t mod) except -1
+#cdef int any_as_nmod(mp_limb_t * val, obj, nmod_t mod) except -1
+cdef nmod_ctx any_as_nmod_ctx(obj)
+
+cdef class nmod_ctx:
+    cdef nmod_t mod
+    cdef bint _is_prime
+
+    cdef int any_as_nmod(self, mp_limb_t * val, obj) except -1
+    cdef nmod _new(self, mp_limb_t * val)
 
 cdef class nmod(flint_scalar):
     cdef mp_limb_t val
-    cdef nmod_t mod
+    cdef nmod_ctx ctx


oscarbenjamin:
> I want to try to get back down to at least the same time if not faster than master.

It is potentially not worth slowing down nmod for prime modulus just to have better error handling for non-prime modulus...

oscarbenjamin:
After playing around with this a bit I think that it is not possible to store an `nmod_ctx` as a Python object attached to an `nmod` without incurring some overhead. I haven't found any way to match let alone beat the current code as long as there is an `nmod_ctx` attribute on `nmod`.

I show a lot of timings below and they vary quite a bit from run to run (e.g. if I quit ipython and restart it). The general trends are as described though. This time I am building with `-Dbuildtype=release` i.e. C compiler optimisations on for the Cython code.

Timings for master:
```python
In [1]: from flint import *

In [2]: ai = 10   # CPython caches small ints

In [3]: %timeit ai * ai
27 ns ± 0.169 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [4]: ai = 1000  # Larger ints not cached

In [5]: %timeit ai * ai
49.4 ns ± 1.36 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [7]: ag = gmpy2.mpz(10)  # gmpy2 slower for small ints

In [8]: %timeit ag * ag
74.2 ns ± 1.15 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [9]: af = nmod(10, 17)  # nmod similar

In [10]: %timeit af * af
75.3 ns ± 0.255 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
So `nmod` times about the same as `gmpy2.mpz` (for small ints). Both are about 50% slower than CPython's `int`. I'm not sure if that is because `int` somehow gets a built-in advantage or whether there is some way for both gmpy2 and python-flint to be faster here.

For very small integers CPython caches them in memory which avoids the heap allocation. That makes it 3x faster than `nmod`. We could contemplate doing the same thing for `nmod` when the modulus is small: just store every element in an array. That wouldn't work for large modulus though and we would need a separate array for each modulus. I'm not sure how worthwhile it is to try to optimise the small modulus case for this.

By contrast with my best efforts (not pushed) to micro-optimise the approach in this PR I get:
```python
In [3]: %timeit af * af
104 ns ± 1.56 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
That's a 30% slowdown. I believe that this is purely due to the unavoidable overhead of needing to assign a Python object in Cython like:
```
(<nmod>result).ctx = (<nmod>self).ctx
```
This happens for every operation that creates a new `nmod` and because `ctx` is an `nmod_ctx` Python object it results in INCREF/DECREF etc. The Cython-generated C code looks like:
```c
  /* "flint/types/nmod.pyx":403
 *             raise ValueError("cannot coerce integers mod n with different n")
 *         r = nmod.__new__(nmod)
 *         r.ctx = s.ctx             # <<<<<<<<<<<<<<
 *         r.val = nmod_mul(s.val, t.val, s.ctx.mod)
 *         return r
 */
  __pyx_t_2 = ((PyObject *)__pyx_v_s->ctx);
  __Pyx_INCREF(__pyx_t_2);
  __Pyx_GIVEREF(__pyx_t_2);
  __Pyx_GOTREF((PyObject *)__pyx_v_r->ctx);
  __Pyx_DECREF((PyObject *)__pyx_v_r->ctx);
  __pyx_v_r->ctx = ((struct __pyx_obj_5flint_5types_4nmod_nmod_ctx *)__pyx_t_2);
  __pyx_t_2 = 0;
```
I don't know what all these macros are doing but the analogous code when we store an `nmod_t` for the modulus is just one line that copies the struct.

Also here are timings for a more macro benchmark where SymPy uses python-flint's `nmod` for `GF(p)` matrices. This is SymPy's pure Python matrix implementation but with `nmod` as the elementary matrix elements. We can invert a 100x100 matrix over `GF(17)` in 11 milliseconds with python-flint master:
```python
In [13]: from sympy import *

In [14]: M = randMatrix(100)

In [15]: dM = M.to_DM(GF(17))

In [26]: %timeit dM.inv()
11.2 ms ± 164 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
Converting that to a Flint `nmod_mat` we get:
```python
In [27]: fM = dM.to_dense().rep.rep

In [28]: type(fM)
Out[28]: flint.types.nmod_mat.nmod_mat

In [30]: %timeit fM.inv()
1.16 ms ± 35.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
So that is about SymPy's pure Python matrix implementation operating with python-flint's `nmod` is about 10x slower than FLINT's `nmod_mat_inv` function. Obviously we expect a pure Python implementation to be slower but it is good to get within a factor of 10 of C here. That depends critically on python-flint's `nmod` speed though.

With this PR instead of python-flint master we get:
```python
In [8]: %timeit dM.inv()
15.9 ms ± 158 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
That's 16 milliseconds rather than 11 milliseconds so a 50% slowdown.

To control all extraneous factors I made these two simpler demo types:
```cython
from cpython.object cimport PyTypeObject, PyObject_TypeCheck

cdef class int_mod1:
    cdef unsigned long long val
    cdef unsigned long long mod

    def __init__(self, val, mod):
        self.val = val
        self.mod = mod

    def __repr__(self):
        return f"{self.val} mod {self.mod}"

    def __mul__(self, other):
        cdef int_mod1 result
        if not PyObject_TypeCheck(other, <PyTypeObject*>int_mod1):
            return NotImplemented
        if self.mod != (<int_mod1>other).mod:
            raise ValueError("cannot multiply integers mod n with different n")
        result = int_mod1.__new__(int_mod1)
        result.val = (self.val * (<int_mod1>other).val) % self.mod
        result.mod = self.mod
        return result

cdef class int_mod2_ctx:
    cdef unsigned long long mod

    def __init__(self, mod):
        self.mod = mod

cdef class int_mod2:
    cdef unsigned long long val
    cdef int_mod2_ctx ctx

    def __init__(self, val, mod):
        self.val = val
        self.ctx = int_mod2_ctx(mod)

    def __repr__(self):
        return f"{self.val} mod {self.ctx.mod}"

    def __mul__(self, other):
        cdef int_mod2 result
        if not PyObject_TypeCheck(other, <PyTypeObject*>int_mod2):
            return NotImplemented
        if self.ctx is not (<int_mod2>other).ctx:
            raise ValueError("cannot multiply integers mod n with different n")
        result = int_mod2.__new__(int_mod2)
        result.val = (self.val * (<int_mod2>other).val) % self.ctx.mod
        result.ctx = self.ctx
        return result
```
The only difference between these two is whether we have a `mod` attribute that is an `unsigned long` or an `int_mod2_ctx` attribute that is a Python object storing the `mod`. These are the timings:
```python
In [10]: a1 = int_mod1(1000, 10000)

In [11]: a2 = int_mod2(1000, 10000)

In [12]: %timeit a1 * a1
91.4 ns ± 2.16 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [13]: %timeit a2 * a2
142 ns ± 0.38 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
So that one difference (using an `nmod_ctx` Python object rather than an `unsigned long` modulus) brings a 50% slowdown. Timings on this vary from run to run but `int_mod2` is always slower. In absolute terms here it is about 50 nanoseconds which seems a lot but for my best efforts with `nmod` the difference was more like 20 nanoseconds. Clock speed here is apparently 1.8GHz so 20ns is about 10 CPU instructions I guess. It seems plausible that the INCREF/DECREF operations amount to that...

Note also:
```python
In [17]: sys.getsizeof(a1)
Out[17]: 32

In [18]: sys.getsizeof(a2)
Out[18]: 48
```
That sort of explains why in this PR `nmod` is still 48 bytes rather than 32. It is because the `nmod_ctx` pointer somehow cannot just increase the size from 24 bytes to 32. I suppose this is some alignment issue or to do with cyclic GC but it is just what happens with Cython when we swap an 8 byte object for a `PyObject*` apparently...

I think that proves then that we can't have an `nmod_ctx` attribute attached to an individual `nmod` so along with `fmpz` and `fmpq` we need an `nmod` to not have a context attribute. There can still be a context from an interface/API perspective it is just that we cannot store the context as an attribute on each `nmod` instance.

For `nmod_poly` and `nmod_mat` though I think that this overhead is likely acceptable e.g. the time to invert a 100x100 matrix is not measurably affected:
```python
In [15]: %timeit fM.inv()
1.12 ms ± 12.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
That is because a 30 nanosecond overhead just doesn't register in the context of a 1 millisecond operation.

The conclusion then is this: we have `nmod_ctx` but it is only used by `nmod_poly` and `nmod_mat` (`nmod_mpoly` already has a context). All other types besides `fmpz*` and `fmpq*` already have contexts. We should still add contexts for `fmpz` and `fmpq` but just they can be global variables. An `nmod.get_ctx()` function would need to retrieve a context object from some global cache.

Part of the reason for wanting to have contexts (apart from uniformity across types) is to store whether or not the modulus is prime. This is actually not needed in the case of `nmod` because the only operations that would have needed it use `n_gcdinv` and `n_sqrtmod` which don't crash and which can detect whether the result exists. It is only `nmod_poly` and `nmod_mat` that need to store the `is_prime` attribute and so we can add contexts for them.

In conclusion then I need to revert a lot of the changes to `nmod` here but not those to `nmod_poly` and `nmod_mat`...

oscarbenjamin:
> That sort of explains why in this PR `nmod` is still 48 bytes rather than 32. It is because the `nmod_ctx` pointer somehow cannot just increase the size from 24 bytes to 32. I suppose this is some alignment issue or to do with cyclic GC but it is just what happens with Cython when we swap an 8 byte object for a `PyObject*` apparently...

I asked on the Cython mailing list and this is apparently due to cyclic GC:
https://groups.google.com/g/cython-users/c/b_LGThvSj8M/m/pPw8c_IRCQAJ?utm_medium=email&utm_source=footer
Basically having any reference to a Python object adds 16 bytes to the memory footprint for the doubly linked list used by the cyclic garbage collector.

oscarbenjamin:
> I asked on the Cython mailing list and this is apparently due to cyclic GC:

Using `@cython.no_gc` seems to bring some speedups for some benchmarks.

> Also here are timings for a more macro benchmark where SymPy uses python-flint's `nmod` for `GF(p)` matrices. This is SymPy's pure Python matrix implementation but with `nmod` as the elementary matrix elements. We can invert a 100x100 matrix over `GF(17)` in 11 milliseconds with python-flint master:
> ```python
> In [13]: from sympy import *
>
> In [14]: M = randMatrix(100)
>
> In [15]: dM = M.to_DM(GF(17))
>
> In [26]: %timeit dM.inv()
>11.2 ms ± 164 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
> ```

Turns out I was completely wrong about this. I forgot that in this particular case what happens is that SymPy's sparse implementation converts the matrix to the dense representation which basically means it uses `nmod_mat_inv` to compute the inverse. The fact that is slowed down from 10 milliseconds to 15 milliseconds is still a problem but it is not due to slower arithmetic but rather slower conversions.

This is a better comparison of these using python-flint master:
```python
In [1]: from sympy import *

In [2]: M = randMatrix(100)

In [3]: dM_sympy_sparse = M.to_DM(GF(17)).to_sdm()

In [4]: dM_sympy_dense = M.to_DM(GF(17)).to_ddm()

In [5]: dM_nmod_mat = M.to_DM(GF(17)).to_dfm().rep

In [6]: %timeit dM_sympy_sparse.rref()
136 ms ± 181 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit dM_sympy_dense.rref()
81.6 ms ± 694 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: %timeit dM_nmod_mat.rref()
313 µs ± 2.19 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
So in this PR with `no_gc` the pure Python implementation using nmod for the arithmetic is about 300-500x slower than Flint's `nmod_mat_rref` function depending on whether the dense or sparse implementation is used.

If I remove the `no_gc` here:
```diff
diff --git a/src/flint/types/nmod.pyx b/src/flint/types/nmod.pyx
index 3b99b54..145b15d 100644
--- a/src/flint/types/nmod.pyx
+++ b/src/flint/types/nmod.pyx
@@ -195,7 +195,6 @@ cdef class nmod_ctx:
         return self._new(&v)
 
 
-@cython.no_gc
 cdef class nmod(flint_scalar):
     """
     The nmod type represents elements of Z/nZ for word-size n.
```
Then timings are:

With `no_gc`:
```python
In [4]: %timeit dM_sympy_dense.rref()
93.2 ms ± 1.85 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [5]: %timeit dM_sympy_dense.rref()
91.6 ms ± 203 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [6]: %timeit dM_sympy_dense.rref()
94.6 ms ± 2.86 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
Without `no_gc`:
```python
In [8]: %timeit dM_sympy_dense.rref()
110 ms ± 1.95 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [9]: %timeit dM_sympy_dense.rref()
112 ms ± 2.9 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [10]: %timeit dM_sympy_dense.rref()
113 ms ± 4.67 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
With master:
```python
In [4]: %timeit dM_sympy_dense.rref()
80.1 ms ± 280 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [5]: %timeit dM_sympy_dense.rref()
82.1 ms ± 2.3 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [6]: %timeit dM_sympy_dense.rref()
84 ms ± 5.9 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
I've repeated this a few times (different matrices, restart process, etc) and it is consistent so `no_gc` makes a difference of 110 ms vs 90 ms but master is still 80 ms. That means that this PR measures about 10% slower than master but it was 30% slower before adding `@cython.no_gc` so clearly `no_gc` makes a big difference.

src/flint/types/nmod.pyx
```diff
@@ -99,100 +255,108 @@ cdef class nmod(flint_scalar):
...
+        cdef nmod r, s2
         cdef mp_limb_t val
-        if any_as_nmod(&val, t, (<nmod>s).mod):
+        s2 = s
+        if any_as_nmod(&val, t, s2.ctx.mod):
             r = nmod.__new__(nmod)
-            r.mod = (<nmod>s).mod
-            r.val = nmod_mul(val, (<nmod>s).val, r.mod)
+            r.ctx = s2.ctx
+            r.val = nmod_mul(val, s2.val, s2.ctx.mod)
```
oscarbenjamin, line 313: Need an `nmod_new` function for this sort of thing.

src/flint/types/nmod.pyx
```diff
@@ -99,100 +255,108 @@ cdef class nmod(flint_scalar):
...
 
         if typecheck(s, nmod):
-            mod = (<nmod>s).mod
-            sval = (<nmod>s).val
-            if not any_as_nmod(&tval, t, mod):
+            s2 = s
+            ctx = s2.ctx
+            sval = s2.val
+            if not ctx.any_as_nmod(&tval, t):
                 return NotImplemented
```
oscarbenjamin, line 340: Calling `ctx.any_as_nmod` method is slower than calling the `any_as_nmod` function...

src/flint/types/nmod_mat.pyx
```diff
@@ -530,8 +571,8 @@ cdef class nmod_mat(flint_mat):
         if self.nrows() != self.ncols():
             raise ValueError("fmpz_mod_mat charpoly: matrix must be square")
 
-        res = nmod_poly.__new__(nmod_poly)
-        nmod_poly_init(res.val, self.val.mod.n)
+        # XXX: don't create a new context for the polynomial
+        res = nmod_poly_new_init(any_as_nmod_poly_ctx(self.ctx.mod.n))
         nmod_mat_charpoly(res.val, self.val)
```
oscarbenjamin, line 576: Need to avoid creating new contexts. I think we need an `nmod_mat_ctx` that holds an `nmod_ctx` and an `nmod_poly_ctx` as attributes.

src/flint/types/nmod_poly.pyx
```diff
@@ -4,50 +4,103 @@ from flint.utils.typecheck cimport typecheck
...
+    return p
+
+
+cdef class nmod_poly_ctx:
+    """
+    Context object for creating :class:`~.nmod_poly` initalised 
+    with modulus :math:`N`.
+
+        >>> nmod_poly_ctx(17)
+        nmod_poly_ctx(17)
```
oscarbenjamin, line 57: We should have `__init__` raise an exception so you have to use something like `nmod_poly_ctx.new()`. Then we can make all contexts unique on construction.

src/flint/types/nmod_poly.pyx
```diff
@@ -711,7 +808,7 @@ cdef class nmod_poly(flint_poly):
         if n == 1:
             return self, int(n)
         else:
-            v = nmod_poly.__new__(nmod_poly)
-            nmod_poly_init(v.val, self.val.mod.n)
+            v = nmod_poly_new_init(self.ctx)
             nmod_poly_deflate(v.val, self.val, n)
+            v.ctx = self.ctx
```
oscarbenjamin, line 813: Should have a `new_nmod_poly(self.ctx)` function.

src/flint/types/nmod.pyx
oscarbenjamin: If we declare them as `@cython.final` to avoid the virtual method table then it is just as fast I think...

src/flint/types/nmod_mat.pyx
oscarbenjamin: Now an `nmod_mat_ctx` holds an `nmod_poly_ctx` which holds an `nmod_ctx`.

src/flint/types/nmod_poly.pyx
oscarbenjamin: Done

src/flint/types/nmod_poly.pyx
oscarbenjamin: Rather `self.ctx.new_nmod_poly()` as a `@cython.final` method.

oscarbenjamin:
@GiacomoPope @Jake-Moss I don't know if either of you wants to review this.

I ended up doing a bunch of somewhat unrelated things like improving the coverage plugin adding tests for full coverage and then needing to make various types consistent with error-handling etc. I also found some problems with the newly added methods like `inverse_series_trunc` that they could crash under some conditions.

The main change here though is that `nmod`, `nmod_poly` and `nmod_mat` all have associated context types `nmod_ctx`, `nmod_poly_ctx`, and `nmod_mat_ctx`. Apart from meaning that we can handle the non-prime modulus case gracefully (no crashes) I am hoping that the pattern here is something that can work for all domains in python-flint.

On master an `nmod` has an `nmod_t` struct as part of the PyOjbect struct so it ends up being 48 bytes which is 16 bytes for `PyObject_HEAD` (ref count and pointer to type object) + 8 bytes for the value + 8 bytes for the `nmod_t` adding up to 48 bytes. Working in C you would not attach the `nmod_t` to each instance which is otherwise just an 8 byte integer but it is potentially the simplest way in Python/Cython.

Here in the PR each `nmod` instance actually holds a pointer to an `nmod_ctx` instance which holds the `nmod_t` as well as an `_is_prime` attribute that is computed when the context is created. This exchanges the 24 byte nmod_t struct for an 8 byte pointer bringing the size down from 48 bytes to 32.

Initially I found that the size was still 48 bytes but that is because Cython implicitly decided that an nmod should be GC-tracked if it has a PyObject pointer as one of its fields. The GC-tracking added not just 16 bytes but also significant time overhead when creating nmod instances. After discussing on the Cython mailing list I see that the `@cython.no_gc` decorator can disable this removing all of the overhead as far as I can tell from my measurements.

I also moved functions like `any_as_nmod` to be methods on the `nmod_ctx` context object but this also brought measurable overhead because of the virtual method table dispatch mechanism. Marking all `cdef` methods on the `nmod_ctx` either `staticmethod` or `cython.final` means that they carry no overhead compared to a plain function call (I think...).

If you have e.g.:
```cython
cdef some_func(nmod_ctx ctx, val):
   ...

cdef class nmod_ctx:
    cdef some_func(self, val):
        ...
```
then from my measurements calling the method will be significantly slower than calling the function:
```cython
ctx.some_func(val) # slow
some_func(ctx, val) # fast
```
The slowdown is due to the virtual method table because a subclass might override the `some_func` method. Declaring the method to be `@staticmethod` or `@cython.final` turns it back into a plain function call at the C level which seems to eliminate the overhead. I don't think that the actual lookup time of the virtual method table is the main issue but rather the fact that it prevents the C compiler from just completely inlining everything so that there is no actual function call overhead in the machine code. This distinction is important because it means that you really need to build with `-Dbuildtype=release` for timings to see the full effect of the comiler inlining everything.

I believe that attaching `nmod_ctx` instead of `nmod_t` to all `nmod` instances does not cause any slowdowns although it is hard to be sure because the timing measurements I get vary a lot from run to run and after rebuilding etc. Likewise I believe that all overhead of using context methods is also eliminated by using `@staticmethod` and `@cython.no_gc` as is done here. I don't think that there are slowdowns here then but I would be interested to hear any timing comparisons measured by others on different machines etc.

I wanted the context objects to be unique in memory which is not possible if using the ordinary constructor like:
```cython
ctx = nmod_ctx(17)
```
I have disabled that so constructing a context that way gives an error requiring instead that a context be created like:
```cython
ctx = nmod_ctx.new(17)
```
This (static) method first tries to lookup the context from a dict that holds all contexts previously created and if not found then it creates a new context. This ensures that the contexts are always unique in memory and so do not need `__eq__` or `__hash__` methods and can be compared as a pointer comparison. Ensuring that the context never gets created twice means that we can absorb the cost of checking whether the modulus is prime up front when the context is created (it takes about 50x longer to check if a large modulus is prime than it does to lookup a context from the dict).

This does all mean that the contexts are never freed from memory though so if someone uses many different moduli then they will gradually consume more and more memory because of context objects that remain attached to the dict in the `nmod` module. I don't believe it is possible to solve that using weakref etc without incurring precisely the GC-type overhead that I referred to above so if a solution is needed that does not just store all the contexts forever then I think it means that the contexts cannot be attached to the nmod instances as literal fields. It wouldn't be hard to change this to do that though in future but it would mean reverting some of the changes I made to `nmod` here.

Regardless of whether the `nmod` instances literally have a field that is a pointer to an `nmod_ctx` context object I still think that it is good from a UI perspective at some level to have an `nmod_ctx` context and that there should also be contexts for `fmpz` and `fmpq` and all types so that there is not a distinction between types that have contexts and types that don't. Ideally all the contexts would have some uniform interface that would be consistent across types and domains so that the types can can be used interchangeably in generic code rather than this sort of awkwardness:
https://github.com/flintlib/python-flint/blob/9a0327f258530adce8052d2aaf1dd349edb3798a/src/flint/test/test_all.py#L2343-L2347
I would want to rewrite that as:
```cython
 Z = flint.fmpz_ctx()
 Q = flint.fmpq_ctx()
 F17 = flint.nmod_ctx(17)
 F163 = flint.fmpz_mod_ctx(163) 
```
Then ideally you can use them all the same way like `F17(2) * F(3)` or `Z(2) * Z(3)` etc.

I changed the nmod constructor so that the second argument can be either an `int` or an `nmod_ctx` i.e.:
```cython
ctx = nmod_ctx(17)
a = nmod(3, 17)
b = nmod(3, ctx)
```
Internally `nmod(3, 17)` will go and look up the context which means a dict-lookup that adds some overhead but note that in Python just calling `nmod(...)` requires looking up the name `nmod` in a dict... Ideally the way to create an instance would be through the context like
```cython
c = ctx(3)
```
in which case there does not need to be any overhead. In any case the cost of a call like `nmod(3, 17)` is much less important than the cost of something like `a * b` which I think does not incur any overhead here.

I also added contexts for `nmod_poly` and `nmod_mat` that work in the same way: all `@no_gc` and static methods and contexts stored in a dict etc. An `nmod_poly_ctx` sometimes needs to be able to create an `nmod` so it internally holds a pointer to an `nmod_ctx` which is retrieved when the `nmod_poly_ctx` is created. This means that it is never necessary to create a context on the fly. An `nmod_mat_ctx` needs to be able to create both `nmod` and `nmod_poly` so it internally holds both an `nmod_ctx` and an `nmod_poly_ctx` and there are no circular references.

At the higher level in future I would envisage having an object like
```cython
Z17 = Zmod(17)
```
which would internally hold references to all three kinds of context and then you could do e.g.:
```cython
a = Z17(2)
x = Z17.poly([0, 1])
M = Z17.matrix([[1, 2], [3, 4]])
mat17 = Z17.mat_ctx()
M2 = mat17([[3], [4]])
```
It might make most sense for the dict management to take place at that higher Zmod level and for Zmod to be a GC-managed object that can be in a weakref dict. The individual nmod and nmod_poly instances would not hold a pointer to Z17 but rather to the `no_gc` context objects added here. Then having `Z17` be GC-tracked would not slow down basic operations but we would be able to free the memory for contexts no longer in use.

I am hoping that the basic structure of the context objects here is a reasonable design that we can use for all types and domains which is why I put some time into micro-optimising and measuring timings etc to see exactly how to organise methods like `new_nmod`, `any_as_nmod` etc in a clean way but without incurring runtime overhead (I hope!).

A separate consideration is how to give these contexts a uniform Python-level interface. I don't think that we can afford the cost of a virtual method table lookup for something like `ctx.new_nmod` but there should be some more generic methods that can be used by Python or Cython code in situations where it is not quite so important that the compiler inlines everything.

Jake-Moss:
> `@GiacomoPope` @Jake-Moss I don't know if either of you wants to review this.

I've just had a quick read over everything here + the diff but won't have the time this week for a proper review.

> The main change here though is that nmod, nmod_poly and nmod_mat all have associated context types nmod_ctx, nmod_poly_ctx, and nmod_mat_ctx. Apart from meaning that we can handle the non-prime modulus case gracefully (no crashes) I am hoping that the pattern here is something that can work for all domains in python-flint.

:tada: 

> On master an nmod has an nmod_t struct as part of the PyOjbect struct so it ends up being 48 bytes which is 16 bytes for PyObject_HEAD (ref count and pointer to type object) + 8 bytes for the value + 24 bytes for the nmod_t adding up to 48 bytes. Working in C you would not attach the nmod_t to each instance which is otherwise just an 8 byte integer but it is potentially the simplest way in Python/Cython.
> 
> Here in the PR each nmod instance actually holds a pointer to an nmod_ctx instance which holds the nmod_t as well as an _is_prime attribute that is computed when the context is created. This exchanges the 24 byte nmod_t struct for an 8 byte pointer bringing the size down from 48 bytes to 32.
> 
> Initially I found that the size was still 48 bytes but that is because Cython implicitly decided that an nmod should be GC-tracked if it has a PyObject pointer as one of its fields. The GC-tracking added not just 16 bytes but also significant time overhead when creating nmod instances. After discussing on the Cython mailing list I see that the @cython.no_gc decorator can disable this removing all of the overhead as far as I can tell from my measurements.

That is pretty interesting, I was unaware of that decorator. Seems like it might also be applicable to other types. I can't imagine any of the `python-flint` types forming cycles, but it's benefits might not be as noticeable.

> I also moved functions like any_as_nmod to be methods on the nmod_ctx context object but this also brought measurable overhead because of the virtual method table dispatch mechanism. Marking all cdef methods on the nmod_ctx either staticmethod or cython.final means that they carry no overhead compared to a plain function call (I think...).

This what I thought https://github.com/flintlib/python-flint/pull/192 would achieve, AFAIK unfortunately for the arithmetic operators, that virtual function look up is required.

> This does all mean that the contexts are never freed from memory though so if someone uses many different moduli then they will gradually consume more and more memory because of context objects that remain attached to the dict in the nmod module.

I think despite this it's certainly something worth doing. AFAIK this is the case with all mpoly contexts ATM. These objects are small enough that millions would have to be made before anyone would look at it and notice something is off. It should be possible to reuse the [C implementation](https://github.com/python/cpython/blob/58ce131037ecb34d506a613f21993cde2056f628/Modules/_functoolsmodule.c#L783) of `functools.lru_cache` if we wanted something that was fixed size, but it would be a little pre-emptive IMO. We could also expose some cache busting functions. If a user was to create a ton of contexts, and was aware of it, they could remove them when they're done, explicitly freeing them.

> Regardless of whether the nmod instances literally have a field that is a pointer to an nmod_ctx context object I still think that it is good from a UI perspective at some level to have an nmod_ctx context and that there should also be contexts for fmpz and fmpq and all types so that there is not a distinction between types that have contexts and types that don't. Ideally all the contexts would have some uniform interface that would be consistent across types and domains so that the types can can be used interchangeably in generic code rather than this sort of awkwardness:

 I agree.

> It might make most sense for the dict management to take place at that higher Zmod level and for Zmod to be a GC-managed object that can be in a weakref dict. The individual nmod and nmod_poly instances would not hold a pointer to Z17 but rather to the no_gc context objects added here. Then having Z17 be GC-tracked would not slow down basic operations but we would be able to free the memory for contexts no longer in use.

I think this is also a good idea.

> I am hoping that the basic structure of the context objects here is a reasonable design that we can use for all types and domains which is why I put some time into micro-optimising and measuring timings etc to see exactly how to organise methods like new_nmod, any_as_nmod etc in a clean way but without incurring runtime overhead (I hope!).

This is quite interested to me, I haven't put a lot of time into micro-optimisations and benchmarks for Cython because my work outside of `python-flint` has always been numeric where we just get down to C++ stdlib calls and call it a day.  Hopefully I get the time to expand my suite to include some micro-benchmarks. I should be able to measure the GC impacts once this is merged and SymPy is able to make use of it. I already can measure GC differences but something is wrong with my post processing ATM.

oscarbenjamin:
> I've just had a quick read over everything here + the diff but won't have the time this week for a proper review.

No problem. There is no rush so I will leave this here if you intend to review it later.

> I can't imagine any of the `python-flint` types forming cycles, but it's benefits might not be as noticeable.

The benefits are going to be most extreme for `nmod` as compared to anything else in python-flint. It is the smallest possible object that in C would only ever be a 64 bit integer so any overhead adds up. That's not the case for other python-flint types e.g. `nmod_mat` for a large matrix has much smaller *relative* overhead compared to working in C. The `~10` nanoseconds that I worried about for `nmod` are not significant for `nmod_mat` unless perhaps we have very small matrices like 2 x 2 or something. Of course ideally someone does not need to do heavy calculations in Python with nmod rather than `nmod_mat` or `nmod_poly` but it is nice if someone can do that and if python-flint makes this as fast as possible for them.

Also it is worth trying to minimise the overhead for trivial cases like the zero polynomial. Consider e.g.:
```python
In [1]: M = randMatrix(10, percent=10) + randMatrix(10, percent=10)*x

In [3]: M.to_DM().to_ddm()
Out[3]: DDM([[7*x, 0, 0, 0, 0, 0, 0, 5*x, 0, 0], [0, 69, 0, 0, 0, 0, 0, 0, 0, 0], [0, 0, 0, 0, 0, 0, 0, 0, 82*x, 0], [69, 96*x, 0, 0, 33, 0, 0, 0, 0, 0], [0, 0, 15*x + 21, 0, 0, 0, 0, 0, 0, 24*x], [14*x, 0, 0, 0, 0, 36, 0, 0, 0, 0], [0, 0, 0, 0, 0, 0, 65*x, 0, 0, 35], [23, 0, 0, 0, 0, 0, 25*x, 0, 31, 14*x], [0, 0, 0, 0, 0, 0, 0, 0, 0, 0], [0, 0, 0, 55, 0, 87, 0, 0, 0, 0]], (10, 10), ZZ[x])
```
Here you have a dense matrix that is full of zeros but those zeros are actually zero *polynomials*. The speed of e.g. matrix multiplication with this representation depends predominantly on how fast you can multiply and add mostly zero polynomials so the overhead in trivial cases does matter.

As for the reference cycles, consider this: an `nmod_mat` sometimes needs to create an `nmod_poly` e.g. for `charpoly`. To make that work an `nmod_mat_ctx` holds a reference to an `nmod_poly_ctx`. If we wanted to add a method to `nmod_poly` that would create a matrix like `p.companion_matrix()` or something then the `nmod_poly_ctx` would somehow need to get the `nmod_mat_ctx`. If the `nmod_poly_ctx` stored a reference to the `nmod_mat_ctx` we would have a reference cycle. If it didn't store the `nmod_mat_ctx` then it would need to use some slower way to get it when the method is called. One fix is to keep things acyclic if possible e.g. we add the `companion_matrix` method to `nmod_mat` instead like:
```python
M = p.companion_matrix() # cyclic
M = nmod_mat.companion_matrix(p) # acyclic
```
Hopefully that acyclic structure:
```
nmod_mat_ctx -> nmod_poly_ctx -> nmod_ctx
```
is something that can reasonably be used for all types and domains.

oscarbenjamin:
> The `~10` nanoseconds that I worried about for `nmod` are not significant for `nmod_mat` unless perhaps we have very small matrices like 2 x 2 or something.

Here is a timing comparison of that. With master:
```python
In [2]: M = nmod_mat([[1, 2], [3, 4]], 17)

In [3]: %timeit M*M
404 ns ± 2.52 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
With the PR:
```python
In [4]: %timeit M*M
425 ns ± 16.7 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
So maybe there is some slowdown...

Oh, wait it is because I forgot to add the `@cython.no_gc` decorator to `nmod_mat`. Another point: The decorator only works if used in the `.pyx` file. If it is only in the `.pxd` file it seems to be ignored...

If I fix that by adding `@cython.no_gc` then:
```python
In [3]: %timeit M*M
413 ns ± 12.6 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit M*M
405 ns ± 15.6 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
The timings can vary a lot though so it is difficult to be conclusive. They usually seem consistent if I run them repeatedly in the same ipython session but if I exit ipython and then start it again they might suddenly change. They also can change significantly if I checkout a different branch and rebuild, so I guess measuring timings this short is just tricky somehow because it depends on so many things. The timings I show here are just the ones that I get as I write the message but I have seen significantly slower and faster timings as well...

It is possible that `timeit` is not a good measure. It doesn't work well for SymPy for example because lots of things are cached there. I also think that trying to measure the time of a single fast operation is difficult. Here is how I would more often time things by making a single slow operation. This is the PR:
```python
In [9]: import math

In [10]: matrices = [M] * 10 ** 6

In [11]: %time math.prod(matrices)
CPU times: user 417 ms, sys: 107 µs, total: 417 ms
Wall time: 416 ms
Out[11]: 
[1, 0]
[0, 1]
```
And this is master:
```python
In [6]: %time math.prod(matrices)
CPU times: user 413 ms, sys: 0 ns, total: 413 ms
Wall time: 412 ms
Out[6]: 
[1, 0]
[0, 1]
```
So basically the same time.

oscarbenjamin:
> It should be possible to reuse the [C implementation](https://github.com/python/cpython/blob/58ce131037ecb34d506a613f21993cde2056f628/Modules/_functoolsmodule.c#L783) of `functools.lru_cache` if we wanted something that was fixed size, but it would be a little pre-emptive IMO. We could also expose some cache busting functions.

The question is whether we guarantee uniqueness of contexts or not. A cache that has a maximum size or that can be cleared cannot guarantee uniqueness.

It is not necessarily critical that we should guarantee uniqueness but just it has some other implications for how you compare the contexts. Suppose for example that we have a `gr_` type with some big hairy complicated context. The `gr.__add__(a, b)` method needs to check that the arguments have matching contexts as a first step in every operation. With unique contexts that comparison is always just a pointer comparison like `a.ctx is b.ctx`. Otherwise the comparison needs to compare by value using the context's `__eq__` methods.

On the other hand if we don't guarantee uniqueness then it becomes a lot easier to ensure that memory is cleared etc. The idea I suggested above about moving the dict management up to the `Zmod` level would basically mean that we compare contexts by value at this lower-level and don't ensure uniqueness but then at the `Zmod` level we make it very unlikely that the contexts are not unique and could ensure that the cost of creating them is low perhaps by using `lru_cache` or similar.

The approach used in `fmpz_mod_ctx` is that uniqueness is not guaranteed but identity is checked as a first step to handle the expected common case:
```python
def __eq__(ctx1, ctx2):
    if ctx1 is ctx2:
        return True
    else:
       return _compare_by_value(ctx1, ctx2)
```
Then in the happy fast path where the contexts match it is not much worse than a pointer comparison. I don't think any timings have been measured to see how that compares with having guaranteed unique contexts and not defining `__eq__` at all as I did for `nmod_ctx` here.

GiacomoPope:
I had a look through this and it seems good -- i dont think i have anything big or urgent to add to what has already been discussed.

oscarbenjamin:
The last two rebases were just because of merge conflicts after all the linting PRs and this time I pushed a commit to handle some remaining lint complaints in this branch.

GiacomoPope:
IMO I think we can include this, with the only "maybe" being the caching of the context, because im not sure what's best here with whether `nmod_ctx` should be cached or just some wrapper `Zmod` later down the line

oscarbenjamin:
It is simple enough to change the caching of the context later. The basic question that needs resolving is whether the context can be attached to an nmod as a field (attribute). There are two kinds of type in python-flint:

1. Types like fmpz where every instance of the type is compatible with every other.
2. Types like nmod, fmpz_mod etc where some additional data (e.g. modulus) is needed to know if the elements are compatible.

In the first case a context can always just be a global object and so no data needs to be attached to instances to identify a context. In the second case we either need to attach a pointer to the context object or some data that identifies the context and could be used to construct a context object on demand.

There is at least in principle some overhead in attaching a pointer to the context object because of reference counting. I don't know whether it might also be problematic in a multithreaded scenario to have all instances sharing a context across multiple threads:

https://peps.python.org/pep-0703/#deferred-reference-counting

Some types like `gr` will definitely need to have pointers to some sort of context object. For other types like `nmod` we can possibly have a pointer but we could also choose not to if it seems to have unacceptable overhead and instead we just attach an `nmod_t` directly to each instance which simplifies memory management for instances.

Since `nmod` is the smallest type, I think we can say that if the overhead is acceptable for `nmod` then it is acceptable for all types. If that is the case then every type either uses a global context or has the context attached as a direct attribute. That is a simple and clean model in which retrieving a context from an object is always a quick operation in either Python or Cython code. Then many low-level operations can quickly retrieve the context and use it do something such as performing coercions or accessing cached data about whether an operation is well defined. The same patterns can be applied across the codebase consistently for all types and some code can be shared generically in context superclasses.

If we can't attach the context object directly at the C level then we need to have some way to get a context from an object but it potentially needs to be created on demand (or looked up from a cache). Then for e.g. `nmod` retrieving the context as an object is not a quick operation any more at the C level. Then we need to have special case code for some types to handle things that would have been handled by the context objects.

Then again we need to special case the types that don't have a context object anyway. Otherwise we could just do:
```cython
cdef class flint_elem:
    flint_ctx ctx
```
Then it would always be fast in Cython to access the `ctx` field of any Flint type directly from the instance struct. That would make it possible to have fast Cython code that operates with different types generically. For this to work though `fmpz` and `fmpq` etc would need to have the `ctx` field in the instance struct as well.

GiacomoPope:
If all elements have an associated context, then we can use these context objects for the coercion. Eg we can have a `ZZ` and `QQ` (global) context which handle coercion to integers, in the same way that our `nmod_ctx` and `fmpz_mod` contexts work. So we would change `any_to_fmpz` to somehting like `ZZ.any_to_fmpz`

GiacomoPope:
yeah i really like the idea of this design decision, having the contexts done like this will help with coercion which is one of the bottlenecks in sage due to all the various parsing needed between libraries. By solving this for python-flint we should have something significantly faster

oscarbenjamin:
> If you look at the C code generated here you can see that this final method ends up just being a normal C function as do any static methods:
> 
> ```c
> static int __pyx_f_5flint_5types_4nmod_8nmod_ctx_any_as_nmod(
>     struct __pyx_obj_5flint_5types_4nmod_nmod_ctx *__pyx_v_ctx, 
>     mp_limb_t *__pyx_v_val,
>     PyObject *__pyx_v_obj
> ) 
> ```

Actually this is only true in the `nmod.pyx` module itself. In e.g. `nmod_poly.ctx` the code there accesses this from a struct containing function pointers.

Looking at it if you want it to be a plain function call everywhere then you need to move the body of these methods in to the .pxd file and declare them `cdef inline`.

I've just pushed a commit to make all `nmod_ctx`, `nmod_poly_ctx`, `nmod_mat_ctx` methods cdef inline so that the comiler can fully inline these functions in all modules.

PR #185, opened 2024-08-14
Remove doubled `mul_mod` functions from `fmpz_mod_poly`

### LINKED ISSUE #183: fmpz_mod_poly has mulmod and mul_mod
It's my mistake but making an issue so we can decide on which name of the two to keep

GiacomoPope:
Fixes #183 by removing the doubled function and picking `mul_mod` over `mulmod` to match things like `compose_mod`.


oscarbenjamin:
Looks good!

PR #190, opened 2024-08-19
Refactor fmpz_mpoly and fmpq_mpoly arithmatic functions

Jake-Moss:
This refactors fmpz_mpoly and fmpq_mpoly to use the new mpoly structure added in https://github.com/flintlib/python-flint/pull/164. It is based on https://github.com/flintlib/python-flint/pull/189, I'll rebase this PRs [one real commit](https://github.com/flintlib/python-flint/commit/b16f1401a92336327527e5148e7b720dde78c651) once https://github.com/flintlib/python-flint/pull/189 is merged

### DIFF
--- a/src/flint/types/fmpq_mpoly.pyx
+++ b/src/flint/types/fmpq_mpoly.pyx
@@ -94,6 +94,22 @@ cdef class fmpq_mpoly_ctx(flint_mpoly_context):
         fmpq_mpoly_ctx_init(self.val, nvars, ordering_py_to_c(ordering))
         super().__init__(nvars, names)
 
+    def any_as_scalar(self, other):
+        if isinstance(other, int):
+            return any_as_fmpq(other)
+        elif typecheck(other, fmpz):
+            return any_as_fmpq(other)
+        elif typecheck(other, fmpq):
+            res = fmpq.__new__(fmpq)
+            fmpq_set((<fmpq>res).val, (<fmpq>other).val)
+            return res
+        else:
+            return NotImplemented
+
+    def scalar_as_mpoly(self, other: fmpq):
+        # non-fmpq scalars should first be converted via self.any_as_scalar
+        return self.constant(<fmpq>other)
+
     def nvars(self):
         """
         Return the number of variables in the context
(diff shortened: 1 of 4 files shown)


Jake-Moss:
This is now rebased on master

src/flint/types/fmpq_mpoly.pyx
```diff
@@ -94,6 +94,22 @@ cdef class fmpq_mpoly_ctx(flint_mpoly_context):
...
 
+    def any_as_scalar(self, other):
+        if isinstance(other, int):
+            return any_as_fmpq(other)
+        elif typecheck(other, fmpz):
+            return any_as_fmpq(other)
+        elif typecheck(other, fmpq):
+            res = fmpq.__new__(fmpq)
+            fmpq_set((<fmpq>res).val, (<fmpq>other).val)
+            return res
```
oscarbenjamin, line 105: If it is an `fmpq` already can we not just return it?

Why does a copy need to be made?

Jake-Moss: I thought it best to be consistent with the other `any_as_*` methods, but since they are immutable on a `python-flint` level perhaps we can just return the same object.

`any_as_fmpq` is just not used there because no conversion is required

```python
>>> from flint import *
>>> x = fmpz(1)
>>> y = any_as_fmpz(x)
>>> x is y
False
```

(With `any_as_fmpz` modified to be a `cpdef`)

oscarbenjamin: Maybe `any_as_fmpz` should be changed to not make a copy. I'm not sure that is compatible with how it is currently used though like whether the object returned is ever mutated.

src/flint/types/fmpq_mpoly.pyx
```diff
@@ -336,205 +352,77 @@ cdef class fmpq_mpoly(flint_mpoly):
...
-            fmpq_mpoly_scalar_mul_fmpq(res.val, (<fmpq_mpoly>self).val, (<fmpq>other).val, res.ctx.val)
-            return res
-        return NotImplemented
+        res = create_fmpq_mpoly(self.ctx)
+        fmpq_mpoly_add_fmpq(res.val, self.val, other.val, self.ctx.val)
+        return res
 
-    def __pow__(self, other, modulus):
+    def _add_mpoly_(self, other: fmpq_mpoly):
         cdef fmpq_mpoly res
```
oscarbenjamin, line 363: Maybe these methods should all be `cdef`.

oscarbenjamin:
This looks good. Improvements can always be made later.

src/flint/types/fmpq_mpoly.pyx
Jake-Moss: I've just tested this locally and it's possible, just needed a little massaging due to C-methods not being visible at runtime. The non-commutative operands can't use runtime reflection after coercion anymore, it needs a explicit C reflected function e.g. `_rtruediv_mpoly_` which can just call the normal C `_truediv_mpoly_` after a cast +  argument swap. Will have this up this evening

Jake-Moss: Actually should just be able to use the normal dunder methods after the coercion rather than have a reflected cdef.

Jake-Moss: [PR opened](https://github.com/flintlib/python-flint/pull/192)

PR #199, opened 2024-08-27
Add DomainError positive char univariate polynomial rings

GiacomoPope:
Following the discussion in #62

### DIFF
--- a/src/flint/flint_base/flint_base.pyx
+++ b/src/flint/flint_base/flint_base.pyx
@@ -254,6 +254,9 @@ cdef class flint_poly(flint_elem):
                 roots.append((v, m))
         return roots
 
+    def real_roots(self):
+        raise AttributeError("Real roots are not supported for this polynomial")
+
     def complex_roots(self):
         raise AttributeError("Complex roots are not supported for this polynomial")
 
--- a/src/flint/flint_base/flint_base.pyx
+++ b/src/flint/flint_base/flint_base.pyx
@@ -255,10 +255,10 @@ cdef class flint_poly(flint_elem):
         return roots
 
     def real_roots(self):
-        raise AttributeError("Real roots are not supported for this polynomial")
+        raise NotImplementedError("Real roots are not supported for this polynomial")
 
     def complex_roots(self):
-        raise AttributeError("Complex roots are not supported for this polynomial")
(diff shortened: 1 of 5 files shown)


oscarbenjamin:
This looks good.

We don't necessarily need to do it here but to complete these we should add an implementation of real_roots for arb_poly, fmpz_poly and fmpq_poly.

For arb_poly it is:
```python
def real_roots(self):
    return self.roots()
```
For `fmpz_poly` we just need a square-free factorisation (all poly have this now after gh-189). Then convert to arb_poly to compute the roots:
```python
def real_roots(self):
    _, fac = self.factor_squarefree()
    roots = []
    for f, m in fac:
        roots.extend([(r, m) for r, m in arb_poly(f).roots()])
    # Sort the roots?
    return roots
```
For `fmpq_poly` the real roots are the real roots of the numerator:
```python
def real_roots(self):
    return self.numer().real_roots()
```
Oh, wait `arb_poly` does not have `roots`:
```python
In [4]: arb_poly([1, 0, 1]).roots()
...
NotImplementedError: Polynomial has no factor method, roots cannot be determined
```
I guess we need to add `real_roots` to `acb_poly` using `acb_poly_validate_real_roots` and then `arb_poly` can have:
```python
def roots(self):
    return acb_poly(self).real_roots()
```
Oh, and there is an interface inconsistency:
```python
In [12]: acb_poly([1, 0, -1]).roots()
Out[12]: [-1.00000000000000, 1.00000000000000]

In [13]: fmpz_poly([1, 0, -1]).roots()
Out[13]: [(1, 1), (-1, 1)]
```
I suppose that `acb_poly` and `arb_poly` could have a different interface for now...

oscarbenjamin:
Or maybe for `fmpz_poly.real_roots` it should just filter the roots from `fmpz_poly.complex_roots()`. There seems to be
`arb_fmpz_poly_complex_roots` but no separate method for real roots.

oscarbenjamin:
I assume that there is some advantage in using `arb_fmpz_poly_complex_roots` rather than `acb_poly_validate_real_roots` for the case where we have exact `fmpz` coefficients. Not sure, but isolating all complex roots is a vastly more expensive operation than isolating only real roots...

oscarbenjamin:
Okay, let's just merge this for now and consider proper `real_roots` support later.

PR #200, opened 2024-08-28
Include ruff

GiacomoPope:
Currently `ruff.toml` is the default settings, feel free to suggest changes. 

The current output of `ruff` statistics is:

```
Jack: python-flint % ruff check --statistics
29	F401	[ ] unused-import
27	F403	[ ] undefined-local-with-import-star
16	E731	[*] lambda-assignment
14	E711	[*] none-comparison
12	E712	[*] true-false-comparison
 9	E721	[ ] type-comparison
 7	E741	[ ] ambiguous-variable-name
 3	F841	[*] unused-variable
 1	E401	[*] multiple-imports-on-one-line
```

I will try and fix these one at a time in some commits.

### DIFF
--- a/src/flint/test/test_all.py
+++ b/src/flint/test/test_all.py
@@ -47,7 +47,7 @@ def test_pyflint():
         ctx.prec = oldprec
 
     assert ctx.dps == 15
-    olddps = ctx.dps
+
     try:
         f1 = flint.arb(2).sqrt()
         ctx.dps = 3
@@ -1530,7 +1530,6 @@ def test_nmod_mat():
     assert (M2.nrows(), M2.ncols()) == (3, 4)
     M3 = M(2,2,[1,2,3,4],17)
     assert M3[0,1] == G(2,17)
-    M3_copy = M(M3)
     M3[0,1] = -1
     assert M3[0,1] == G(-1,17)
     def set_bad(i,j):
--- a/src/flint/test/test_all.py
+++ b/src/flint/test/test_all.py
@@ -2803,7 +2803,7 @@ def test_polys():
         assert P([1, 2, 1]).derivative() == P([2, 2])
 
         p = P([1, 2, 1])
-        if is_field and type(p) is flint.fq_default_poly:
+        if is_field and type(p) is not flint.fq_default_poly:
             assert p.integral() == P([0, 1, 1, S(1)/3])
         if type(p) is flint.fq_default_poly:
             assert raises(lambda: p.integral(), NotImplementedError)
--- a/src/flint/test/test_all.py
+++ b/src/flint/test/test_all.py
@@ -197,8 +197,8 @@ def test_fmpz():
     assert abs(flint.fmpz(1)) == 1
     assert abs(flint.fmpz(-1)) == 1
 
-    assert bool(flint.fmpz(0)) == False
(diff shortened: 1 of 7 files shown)


GiacomoPope:
Many of the `F401` issues are because of: https://github.com/PyCQA/pyflakes/issues/471 -- I suppose we should decide about whether to follow best practices and define an `__all__`

GiacomoPope:
After the above fixes we're at

```py
Jack: python-flint % ruff check --statistics
21	F401	unused-import
```

Related to the comment above.

oscarbenjamin:
> we should decide about whether to follow best practices and define an `__all__`

Yes, I think that `__all__` should be defined but which files does it want it on?

Is it only the .pyx files or does it also want it in .pxd files?

Also we should remove all `import *`.

GiacomoPope:
The `__all__` seems to be only needed in the `*/__init__.py` files as far as I can tell. I have also reenabled F403 so we eventually remove the * imports.

```
Jack: python-flint % ruff check --statistics                     
27	F403	undefined-local-with-import-star
21	F401	unused-import
```

I will be busy now potentially until the end of the week. Feel free to merge / ignore this if you want and we can continue work whenever.

GiacomoPope:
actually, i think I have misunderstood ruff, the above has ONLY checked python files.

```
Jack: python-flint % ruff check src/flint/types/fq_default.pyx --statistics
203		syntax-error
```

This means I must have missed some flag for the cython files and there might be a LOT more work to do. I will have to look into this more after this week

src/flint/test/test_all.py
```diff
@@ -3678,10 +3671,8 @@ def test_matrices_eq():
         assert (A1 != A2) is False
         assert (A1 == B) is False
         assert (A1 != B) is True
-        assert (A1 == None) is False
-        assert (A1 != None) is True
-        assert (None == A1) is False
-        assert (None != A1) is True
+        assert (A1 is None) is False
+        assert (A1 is not None) is True
```
oscarbenjamin, line 3675: This is also the case for some of the other changes but the code here is supposed to be testing `__eq__` so it does need to be `==` rather than `is`. In cases like this we just need a `# ruff: ignore` or whatever the flag is.

GiacomoPope:
we should probably just close this, im not sure this PR really helps with anything currently.

GiacomoPope:
I also ran

```
Jack: python-flint % cython-lint src/flint/types/*.pyx
Jack: python-flint % cython-lint --version
0.16.0
```

on main and got no errors.

oscarbenjamin:
The actual changes here to the code are mostly not needed or not correct so I guess that the ruff configuration would need to be improved to be useful. I think it could be possible to improve that but it requires choosing which ruff rules to use and configuring them.

Obviously feel free to close if you don't want to work on it.

The cython-lint checks pass now but note that lots of checks are disabled:
https://github.com/flintlib/python-flint/blob/be0e41057339b71321d9539558bf054266540cea/pyproject.toml#L42-L45
We would probably not want to ignore all of those but enabling each one would likely then require making some changes somewhere in the codebase.

GiacomoPope:
OK -- I'll close this for now and if I feel like refactoring I'll look at removing elements from the ignore list and refactoring for each one potentially.

PR #203, opened 2024-08-30
Various simple fixes

fchapoton:
a sequence of commits fixing various things, mostly cosmetics

### DIFF
--- a/bin/rst_to_pxd.py
+++ b/bin/rst_to_pxd.py
@@ -76,7 +76,7 @@ def fill_import_dict(pyflintlibdir):
 
 def undecorate(str):
     """
-    remove variable name, const, *, etc. to just get types
+    remove variable name, const, ``*``, etc. to just get types
     """
     ret = str.strip()
     ret = ret[:ret.rfind(' ')]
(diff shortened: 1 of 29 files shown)


oscarbenjamin:
Thanks. Did you use a tool to find these?

There was some recent discussion about `cython-lint`: https://github.com/flintlib/python-flint/pull/191#issuecomment-2317341048.

It would be good to have some sort of linting for local use and for CI and I think that cython-lint at least picks up on some of the cases you have covered here. I just checked and this PR reduces the number of cython-lint complaints from 455 to 342.

There are no e.g. no W605 invalid escape sequence errors with the changes here so now it would be good to add a CI check that use cython-lint and checks for those.

fchapoton:
I indeed used cython-lint. I can try to help make a linter with cython-lint in some other pull request, sure.

oscarbenjamin:
> `sage --tox -e rst` to check the rst syntax

Is that just something in-house for Sage?

> I can try to help make a linter with cython-lint in some other pull request, sure.

Feel free to do that. I wasn't suggesting that you need to do it or that it needs to be part of this PR though. I was just observing that it is something that we should do generally speaking. First step of course is to fix existing errors as you have done here.

oscarbenjamin:
Looks good. Thanks!

PR #204, opened 2024-08-30
first sketch of cython-lint workflow

fchapoton:
let's try to add a cython-lint checker in a workflow


oscarbenjamin:
It might be that the workflow doesn't run until it is actually merged or something.

fchapoton:
yes, could be. I have to add many ignored codes too

.github/workflows/lint.yml
```diff
@@ -0,0 +1,26 @@
+name: Linting
+
+on: [push]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.12"]
```
oscarbenjamin, line 10: Not sure it matters but minimum version is 3.10 now.

Does the lint check depend on the Python version?

fchapoton: ok. I think the cython-lint does not, but maybe `ruff` checks will depend if they are added later. So let us keep a set of python versions

fchapoton: now the lint workflow is running ; probably it will fail, let's see

fchapoton:
now with large-scale removal of unused variables and imports. The former in particular may benefit from a careful check.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,26 @@
...
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install dependencies
+      run: |
+        pip install uv
+        uv pip install cython-lint --upgrade --system
+
+    - name: cython-lint
+      run: |
+        cython-lint --ignore=E114,E117,E127,E128,E129,E202,E221,E222,E231,E261,E262,E265,E302,E303,E306,E501,E701,E703,E711,E722,E731,E741,E743,W291,W293,W391,W605 src/
```
oscarbenjamin, line 26: Does it still need all of these?

I think it is fine if there are lots of ignore code for now because they can be disabled incrementally.

It would be better if the ignore codes are in pyproject.toml though so that it works the same when running locally as in CI.

oscarbenjamin: Or maybe it is better to leave these here for now until most errors are fixed since we don't intend to keep all of these ignores codes.

oscarbenjamin:
I've been through all the diff and it looks good although I'll let the CI finish.

.github/workflows/lint.yml
fchapoton: I think I only added codes that I have seen failing.

I have moved the config to pyproject.toml

fchapoton: Although my branch here is not based on my recent fix-ups, that in particular fix W605.

fchapoton:
I have rebased the branch on develop, squashed the first few commits and shortened the ignore list

oscarbenjamin:
Okay, looks good to me. I'll wait for CI to finish.

oscarbenjamin:
I'm going to add a development workflow page to the docs and I'll mention how to run cython-lint there. We should probably arrange it so that you can do e.g.
```console
$ spin lint
```
since `spin` is the developer frontend.

oscarbenjamin:
Okay this looks good. Thanks!

PR #206, opened 2024-08-31
fine tuning the cython-linter + fix W391 code

fchapoton:
- using just python 3.12
- removing E701 and E703, E302,E303,E306 from ignored, already fixed
- removing W391 too, fixed here


oscarbenjamin:
Looks good. Thanks!

oscarbenjamin:
Some of these could be fixed by the script mentioned https://github.com/flintlib/python-flint/pull/191#issuecomment-2317451322 which we should probably include as `bin/lint_fix.py` or something.

PR #208, opened 2024-09-01
Update docs for install, build and dev workflow

oscarbenjamin:
Update the docs for installing. The setup page is split into two parts where the simple page just explains how to `pip install python-flint` and which versions of Python have binaries available.

The other page gives more complicated installation instructions for building from source including proper instructions for different old versions and which versions of Flint etc are supported by which versions of python-flint.

Add a new workflow page that explains how to use `meson` and `spin` for development and how to build and run tests etc.

Also add version checks for Cython in the meson build so that it will check the version when building without build isolation.


oscarbenjamin:
Docs updated so the new pages are:

- install: https://python-flint.readthedocs.io/en/latest/install.html
- build: https://python-flint.readthedocs.io/en/latest/build.html
- development workflow: https://python-flint.readthedocs.io/en/latest/workflow.html

These three are linked from the main page:

https://python-flint.readthedocs.io/en/latest/

Let me know if anyone has any suggestions for improvement.

Issue #210, opened 2024-09-02
E743 ambiguous function definition


GiacomoPope:
```py
    def l(self, s):
        """
        Evaluates the Dirichlet L-function of this character at the given
        complex number s.

            >>> from flint import showgood
            >>> chi = dirichlet_char(1, 1)
            >>> showgood(lambda: chi.l(2), dps=25)
            1.644934066848226436472415
            >>> chi = dirichlet_char(7, 3)
            >>> showgood(lambda: chi.l(2+3j), dps=25)
            1.273313649440490751755284 - 0.07432329442559421607102118j

        """
        s = any_as_acb(s)
        cdef acb v
        v = acb.__new__(acb)
        acb_dirichlet_l((<acb>v).val, (<acb>s).val, self.G.val, self.val, getprec())
        return v
```

We currently have this function, which triggers the listing error `E741 ambiguous variable name 'l'`. I think we could probably call this `dirichlet_l` to match Flint, or `l_function`? I'm not sure what other people think so I've made this issue to discuss.

oscarbenjamin:
Does cython-lint have a way to mark it as ignored like `# lint: ignore`?

Having a different name does not seem unreasonable if we don't like the current name. The old name should probably still be kept as an alias though since there is no strong reason to break compatibility.

GiacomoPope:
yeah i think we can add a comment to ignore the warning, but if we allow `l()` here, should we just allow it wherever and maintain the ignore in the list?

I have no feelings about the current name really. I probably wouldnt call it `l(s)` but it's also probably fine.

PR #211, opened 2024-09-02
Small changes to please cython-lint

GiacomoPope:
Most changes are trivial.

See issue #210 about one remaining small change. The "last" thing I think we need to decide is a line length so we can fix E501 eventually.

### DIFF
--- a/src/flint/types/acb_poly.pyx
+++ b/src/flint/types/acb_poly.pyx
@@ -261,7 +261,7 @@ cdef class acb_poly(flint_poly):
         q = acb_poly.__new__(acb_poly)
         r = acb_poly.__new__(acb_poly)
         if acb_poly_divrem((<acb_poly>q).val, (<acb_poly>r).val,
-                (<acb_poly>s).val, (<acb_poly>t).val, getprec()):
+                           (<acb_poly>s).val, (<acb_poly>t).val, getprec()):
             return q
         else:
             raise ZeroDivisionError("acb_poly leading coefficient must be nonzero")
@@ -281,7 +281,7 @@ cdef class acb_poly(flint_poly):
         q = acb_poly.__new__(acb_poly)
         r = acb_poly.__new__(acb_poly)
         if acb_poly_divrem((<acb_poly>q).val, (<acb_poly>r).val,
-                (<acb_poly>s).val, (<acb_poly>t).val, getprec()):
+                           (<acb_poly>s).val, (<acb_poly>t).val, getprec()):
             return r
         else:
             raise ZeroDivisionError("acb_poly leading coefficient must be nonzero")
@@ -301,7 +301,7 @@ cdef class acb_poly(flint_poly):
         q = acb_poly.__new__(acb_poly)
         r = acb_poly.__new__(acb_poly)
         if acb_poly_divrem((<acb_poly>q).val, (<acb_poly>r).val,
-                (<acb_poly>s).val, (<acb_poly>t).val, getprec()):
+                           (<acb_poly>s).val, (<acb_poly>t).val, getprec()):
             return q, r
         else:
             raise ZeroDivisionError("acb_poly leading coefficient must be nonzero")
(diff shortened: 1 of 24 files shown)


GiacomoPope:
I don't understand the E202 error in the CI as I don't get this locally and the error itself seems to point to `f{x = }` in format strings, which *should* be allowed... seems weird that this lint is failing for the CI machine but not locally on mine?!

GiacomoPope:
Hmm it seems that people argue about this elsewhere: https://github.com/PyCQA/pycodestyle/issues/1201 and if my E202 doesn't fail but it does on the CI I'll just include it back into the ignore.

src/flint/types/acb_poly.pyx
```diff
@@ -197,7 +197,7 @@ cdef class acb_poly(flint_poly):
         return u
 
     def __pos__(self):
-        return self # ?
+        return self
```
oscarbenjamin, line 200: I think something should be kept in the comment here.

This is not how it works for `arb` and `acb` since `+a` rounds to context precision:
```
In [1]: from flint import *

In [2]: a = arb(10**50)

In [3]: a
Out[3]: 1.00000000000000e+50

In [4]: +a
Out[4]: [1.00000000000000e+50 +/- 3.40e+34]

In [5]: a.man_exp()
Out[5]: (88817841970012523233890533447265625, 50)

In [7]: (+a).mid().man_exp()
Out[7]: (4814824860968089, 114)
```
I assume that the comment here is a reminder that maybe this method should be changed to match by using `acb_poly_set_round`.

src/flint/types/fmpz.pyx
```diff
@@ -528,8 +528,8 @@ cdef class fmpz(flint_scalar):
         return u
 
     # This is the correct code when fmpz_or is fixed (in flint 3.0.0)
-    #
-    #def __or__(self, other):
+
+    # def __or__(self, other):
```
oscarbenjamin, line 532: Now that Flint 3.0 is the minimum version we could use this code instead.

GiacomoPope: Ahh ok. Shall I make this change in this PR or another

oscarbenjamin: It should work fine just to uncomment this as it is tested. There was a bug in Flint 2.7 at the time.

oscarbenjamin:
> seems weird that this lint is failing for the CI machine but not locally on mine?!

The two likely explanations for this are always:

- Different version of something e.g. different cython-lint or perhaps even different Python version (some lint tools behave differently for different Python versions). Obvious thing is to make sure you have the latest version of `cython-lint` since that is what CI will use.
- The PR branch is not based off of the tip of master i.e. there are further changes on master that are not in the PR branch. In this case a merge/rebase with master should mean that you see the same thing locally as in CI.

src/flint/types/fmpz.pyx
GiacomoPope: ok, this has been addressed

src/flint/types/acb_poly.pyx
GiacomoPope: I have reintroduced the comments

GiacomoPope:
At the moment the rule is flagging up the format strings (in the CI at least) which I disagree with (I think `f"{x = }"` produces a better output than `f"{x=}"`). Let me add some comments to the toml.

oscarbenjamin:
> At the moment the rule is flagging up the format strings (in the CI at least) which I disagree with (I think `f"{x = }"` produces a better output than `f"{x=}"`)

The question is whether the rule is actually useful in other situations. If it is then we can work around it in this case e.g. `f"x = {x}"`.

If the rule is generally not that important but a bit nice for formatting then I would be most inclined just to "fix" it throughout the codebase and then just leave it excluded in the lint config.

GiacomoPope:
Looks like E202 is now fixed by doing the (in my opinion worse) option of `x = {x}` considering `{x = }` is intended to be used... Maybe better to do this though than have the lint hack in place.

oscarbenjamin:
Okay, looks good.

PR #212, opened 2024-09-02
rename l_function method name

### LINKED ISSUE #210: E743 ambiguous function definition

GiacomoPope:
Fixes #210


src/flint/types/dirichlet.pyx
```diff
@@ -189,6 +189,9 @@ cdef class dirichlet_char(object):
         acb_dirichlet_l((<acb>v).val, (<acb>s).val, self.G.val, self.val, getprec())
         return v
 
+    # For backwards compatibility we allow self.l(s) see Issue #210
+    l = l_function # no-cython-lint
```
oscarbenjamin, line 193: There isn't any point in doing this unless the docstring is updated to suggest calling the new function name.

I would do it like:
```
def l_function(self, s):
    """
    >>> l_function()
    123
    """
    ...


def l(self, s):
    """
    Alias for :meth:`l_function`
    """
    return self.l_function(s)
```

oscarbenjamin:
I'm not sure it is worth changing the name here. I don't know much about this function and I don't know what it would normally be called. If the linter wasn't complaining then it would not have been suggested to change this so I would probably just leave it I think.

Putting "function" in the name seems sort of redundant when most methods are functions.

src/flint/types/dirichlet.pyx
GiacomoPope: oh sorry, i just missed this

GiacomoPope:
I think `l_function` makes sense, as it's a mathematical object known as a l-function (https://en.wikipedia.org/wiki/Dirichlet_L-function), obviously all the code is built with methods/functions but just calling it `l` doesnt seem very helpful to me. Happy to just close this though if you disagree

oscarbenjamin:
Okay, let's get it in.

Issue #195, opened 2024-08-21
Interface for making an mpoly context


oscarbenjamin:
CC @Jake-Moss 

The idea of higher-level interfaces is discussed in gh-53. I would imagine that ultimately the usual way to create a multivariate polynomial would look something like:
```python
from flint import ZZ
R, [x, y] = ZZ.mpoly_ring(['x', 'y'])
p = x**2 - y**2
p.factor()
```
Other options like monomial ordering could be keyword arguments.

In that sense we can think of the current interface as being sort of "low-level" but still I think it can be nicer than it is. This is currently the way to construct and use a multivariate polynomial context:
```python
ctx = fmpz_mpoly_ctx(3, Ordering.lex, ["x", "y", "z"])
x, y, z = ctx.gens()
(x**2 - y**2).factor()
```
I think it is reasonable that the generators are not returned by default but the constructor syntax is awkward. It is redundant to specify the number of generators if we are already supplying a list of names. Also most of the time using lex as the default ordering is fine. I doubt many people will ever care about the monomial ordering.

It would be better to change it so that you can do
```python
ctx = fmpz_mpoly_ctx(["x", "y", "z"])
```
and get a lex ordered ring by default. I think it would be reasonable to say that the ordering parameter could be a keyword-only argument like:
```python
ctx = fmpz_mpoly_ctx(["x", "y", "z"], order='degrevlex')
```
Another problem is that the monomial ordering has to be passed as an enum which is also awkward e.g.:
```python
import flint
ctx = flint.fmpz_mpoly_ctx(["x", "y", "z"], order=flint.Ordering.degrevlex)
```
It is reasonable to use these enums but it should be possible to just pass a string when calling the constructor. I think that a string should be displayed in the repr as well:
```python
>>> ctx
fmpz_mpoly_ctx(["x", "y", "z"], order="lex")
```
The reason that the first argument to the current constructor is an integer is because it is envisaged that you would do:
```python
>>> ctx = fmpz_mpoly_ctx(3, Ordering.lex, "x")
>>> ctx.gens()
(x1, x2, x3)
```
I think that a better way of handling that is to allow a symbol name to be a tuple of name and count e.g.:
```python
>>> ctx = fmpz_mpoly_ctx([("x", 3), "y"])
>>> ctx.gens()
(x1,x2,x3,y)
```
[...]

PR #216, opened 2024-09-04
Add mpoly docs, functions for discriminant, resultant, term_content, and deflation, and run mod_mpoly doctests.

Jake-Moss:
A small collection of changes I've just pushed. This PR:
- Adds `mpoly` docs, addresses https://github.com/flintlib/python-flint/issues/195#issuecomment-2323515445
- Adds `mod_mpoly`s to the doc test run list. Previously they were not being run, I've fixed that up and fixed the docs tests themselves. There doesn't appear to be any other modules missing from the list.
Ideally these would be discovered and run as part of pytest. There's a [pytest plugin for Cython modules](https://github.com/lgpage/pytest-cython), however it relies on in-tree builds. Cython conveniently provides the `__test__` dictionary in modules that contain function names + line no. to doc string for doc strings that contain `>>>`. I think is is all that is required so a pytest plugin could work with meson-python's out of tree builds. 
- Adds functions for discriminant, resultant, term_content, and deflation for all `mpoly` types.
- Adds a `primitive_part` function for `fmpz_mpoly`. 

I've added those functions as I intended to use them with SymPy's DMP_Flint, however these are the univariate functions adapted for multivariate polynomial types, not the multivariate variants that SymPy seems to expect. Regardless they're here now.


oscarbenjamin:
> * Previously they were not being run, I've fixed that up and fixed the docs tests themselves. There doesn't appear to be any other modules missing from the list.

We might want to have a test for this a bit like `test_all_tests` that checks the tests are all listed in `test_all.py`.

> * Ideally these would be discovered and run as part of pytest. There's a [pytest plugin for Cython modules](https://github.com/lgpage/pytest-cython), however it relies on in-tree builds. Cython conveniently provides the `__test__` dictionary in modules that contain function names + line no. to doc string for doc strings that contain `>>>`. I think is is all that is required so a pytest plugin could work with meson-python's out of tree builds.

I'm not quite sure I understand what you mean by this. Firstly I believe that the `__test__` thing is what causes doctest failures to be reported more than once sometimes. It doesn't happen consistently though so not sure how to reproduce right now...

Is there a simple way to adapt the current code so that instead of using a hard-coded list of modules it can walk the tree looking for `__test__`?

Secondly you seem to be saying that there is a plugin but it doesn't work. Or are you saying that we should make a new plugin or send a fix to the existing one?

I agree that it would be better if pytest could do this and if we could arrange it so that `spin test` would run the doctests by default. Writing a plugin is hard work though. I'm trying to upstream the coverage one.

It would still be nice even if pytest is used if there is a way to run all the doctests like `python -m flint.test` after install. I like the fact that users can run the tests:
```console
pip install python-flint
python -m flint.test
```

oscarbenjamin:
> * Adds functions for discriminant, resultant, term_content, and deflation for all `mpoly` types.

This is nice. I wonder for `deflation` if there should also be `deflate` and `inflate`. At least `inflate` is needed to be able to reverse `deflation` although I guess compose can do it:
```python
In [6]: from flint import *

In [7]: x = fmpz_poly([0, 1])

In [8]: p = x**4 + x**2 + 1

In [9]: p.deflation()
Out[9]: (x^2 + x + 1, 2)

In [10]: p2, n = p.deflation()

In [11]: p2(x**n)
Out[11]: x^4 + x^2 + 1
```
Using compose is more awkward for `mpoly` though.

Looking at this across other types it seems to be somewhat unfortunately designed that we have:

- `p.inflate(n)` - replace `x -> x^n`.
- `p.deflate(n)` - replace `x^n -> x`.
- `p.deflation()` - find `n` and replace `x^n -> x`.

So there is no method to get `n` without computing the deflated polynomial. The reason we might want `n` without the polynomial is that you might have more than one polynomial that you want to transform analogously e.g.:
```python
n1 = p1.get_deflation()
n2 = p2.get_deflation()
n = gcd(n1, n2)
p1p = p1.deflate(n)
p2p = p2.deflate(n)
```

oscarbenjamin:
> I've added those functions as I intended to use them with SymPy's DMP_Flint, however these are the univariate functions adapted for multivariate polynomial types, not the multivariate variants that SymPy seems to expect.

I'm not sure I understand here. Note that there are many different function in SymPy as well as two different implementations of multivariate polynomials and the fact that any kind of polynomial can have multivariate polynomial coefficients...

We have e.g.
```python
In [2]: Poly(2*x**2*y + 4*y).primitive()
Out[2]: (2, Poly(x**2*y + 2*y, x, y, domain='ZZ'))

In [3]: Poly(2*x**2*y + 4*y, x).primitive()
Out[3]: (2⋅y, Poly(x**2 + 2, x, domain='ZZ[y]'))
```
In the second case the coefficients would by python-flint types if PolyElement could wrap python-flint's mpoly types.

In the first case the `Poly` would internally hold a `DMP_FLint` which would internally hold an `fmpz_mpoly`. The `primitive_part` method here only computes the primitive part but not also the content. Of course you can get the content by dividing the polynomials but it would have been better to compute the content and divide the polynomial by the (scalar) content which is what `fmpz_poly_primitive_part` does except that it discards the content.

Suppose that we had a way to get the coefficients of an `fmpz_mpoly` as an `fmpz_vec` and also that `fmpz_vec` had a `gcd` method. Then in python-flint (or in SymPy) you could do:
```python
def content(p):
    content = p.coeffs_vec().gcd()

def content_primitive(p):
    primitive_part = p / p.content() # scalar div
    return content, primitive
```
Apart from making the leading coefficient positive that is exactly what `fmpz_poly_primitive_part` does. Otherwise we need:
```python
def content_primitive(p):
    primitive_part = p.primitive_part()
    content = p / primitive_part # poly div
    content = content.leading_coefficient()
    if content < 0:
        content = -content
        primitive_part = -primitive_part
    return content, primitive
```
That's fine but needing a polynomial division is unfortunate.

Note that SymPy would attach the sign to the polynomial rather than the content (not sure if that's actually a good idea but that is what it does):
```python
In [5]: primitive(-x)
Out[5]: (1, -x)
```

oscarbenjamin:
For resultant and discriminant we have these in SymPy:
```python
In [6]: Poly(x + y).resultant(Poly(x - y))
Out[6]: Poly(-2*y, y, domain='ZZ')
```
In this case we need to go from a ring with 2 generators to a ring with 1 generator. That is not what Flint's resultant etc functions do so we would need some way of converting the `fmpz_mpoly` from `R[x,y] -> R[y]`.

src/flint/test/test_all.py
```diff
@@ -3229,6 +3229,27 @@ def quick_poly():
...
+
+        assert (x0**2 - 1).resultant(x0**2 - 2*x0 + 1, 'x0') == 0
+        assert (x0**2 + x1**2 - 1).resultant(x0 - x1, 'x0') == 2*x1**2 - 1
+
+        assert (x0**3 - 6*x0**2 + 11*x0 - 6).discriminant('x0') == 4
+        assert (x0**2 + 4*x0*x1 + 4*x1**2 - 1).discriminant('x0') == 4
+
+        res, stride = (3*x0**2*x1**2 + 6*x0*x1**2 + 9*x1**2).deflation()
+        assert res == 3*x0**2 + 6*x0 + 9
+        assert stride.to_tuple() == (1, 0)
```
oscarbenjamin, line 3244: Is it any better to have a `stride.to_tuple()` method rather than just calling `tuple(stride)`?

src/flint/types/fmpz_mpoly.pyx
```diff
@@ -977,6 +1066,27 @@ cdef class fmpz_mpoly(flint_mpoly):
         fmpz_mpoly_reduction_primitive_part(res.val, self.val, (<fmpz_mpoly_vec>vec).val, self.ctx.val)
         return res
 
+    def deflation(self):
+        """
+        Compute the deflation of ``self``. See Flint documentation for
+        details. Returns deflated polynomial and the stride vector.
```
oscarbenjamin, line 1072: I went and looked at the Flint documentation which says

>    For each variable `v` let `S_v` be the set of exponents appearing on `v`.
>    Set ``shift[v]`` to `\operatorname{min}(S_v)` and set ``stride[v]`` to `\operatorname{gcd}(S-\operatorname{min}(S_v))`.
>
>    If *A* is zero, all shifts and strides are set to zero.

It is probably best not to refer to the very terse Flint documentation especially when the description there doesn't fully match what this function does. Although we should probably set up intersphinx and see if we can link to the flint docs when we want to refer to them...

Also though I am wondering if handling both shifts and strides is a better thing to do in the python-flint interface. What you have here matches with what existing `deflation` functions in python-flint do I suppose.

oscarbenjamin: Actually the shifts thing seems a little odd. I don't think it's needed.

oscarbenjamin:
I think that the changes here look good and could be merged if you are done with this.

oscarbenjamin:
Actually I don't think that `deflation` works quite right because of the shifts:
```python
In [30]: f = x**3 * y + x * y**4 + x * y

In [31]: f
Out[31]: x^3*y + x*y^4 + x*y

In [32]: f_deflated, strides = f.deflation()

In [33]: f_deflated
Out[33]: x + y + 1

In [34]: strides
Out[34]: fmpz_vec(['2', '3'], 2)

In [36]: f_deflated.compose(x**2, y**3)
Out[36]: x^2 + y^3 + 1

In [37]: f_deflated.compose(x**2, y**3) * x*y
Out[37]: x^3*y + x*y^4 + x*y
```
The output is affects by the shifts but they aren't returned. Either this should not use the shifts or it needs to return them.

src/flint/test/test_all.py
Jake-Moss: No I don't believe so, we should probably just provide a proper `iter` generator instead. I'll add that tonight

src/flint/types/fmpz_mpoly.pyx
Jake-Moss: > What you have here matches with what existing deflation functions in python-flint do I suppose.

Yep that was my reasoning. Perfectly happy to expose all three methods as well as a sort of "just do whole the deflation" method. Will do so tonight

> It is probably best not to refer to the very terse Flint documentation especially when the description there doesn't fully match what this function does.

That's reasonable. I'll make that change

Jake-Moss:
> I'm not quite sure I understand what you mean by this. Firstly I believe that the `__test__` thing is what causes doctest failures to be reported more than once sometimes. It doesn't happen consistently though so not sure how to reproduce right now...
> 
> Secondly you seem to be saying that there is a plugin but it doesn't work. Or are you saying that we should make a new plugin or send a fix to the existing one?

> I agree that it would be better if pytest could do this and if we could arrange it so that `spin test` would run the doctests by default. 

Sorry I should have clearer, there is a pytest plugin by the name of `pytest-cython` that provides doctest support for Cython code. I briefly tried and I couldn't get it to work with an out-of-tree build, the documentation says 

> It is assumed that the C extension modules have been build in place before running py.test and there is a matching Cython .pyx file

I've briefly looked over the code and it's a small single file plugin (<200LOC), from there I saw mention of the `__test__` attribute with this comment 
>```python
>        # If somehow the equivalent test does not already exist, we
>        # keep the __test__ test (maybe it is something else not
>        # generated by autotestdict)
>```
https://github.com/lgpage/pytest-cython/blob/bdd8266f3a9f58e7efd931ab8dbec42bc9e481f7/src/pytest_cython/plugin.py#L137-L140

It seems like the plugin could pick up the "missing" doctests from the modules alone but they are being skipped because

>```python
># only run test if matching .so and .pyx files exist
>```
https://github.com/lgpage/pytest-cython/blob/bdd8266f3a9f58e7efd931ab8dbec42bc9e481f7/src/pytest_cython/plugin.py#L45-L46

To me this sounds like the existing plugin could be modified to support running doctests from the modules alone, like how the existing `flint.test` does.

> Writing a plugin is hard work though. I'm trying to upstream the coverage one.

It certainly seems like it, thank you for that. I gave the pytest docs for it a skim but had to move on to other things. I might give patching pytest-cython a crack this weekend but it's low on my list of things to do.

> Is there a simple way to adapt the current code so that instead of using a hard-coded list of modules it can walk the tree looking for `__test__`?

I haven't experimented with this but I believe so.

> It would still be nice even if pytest is used if there is a way to run all the doctests like python -m flint.test after install. I like the fact that users can run the tests:

That is nice just as a sanity check.

Jake-Moss:
> I'm not sure I understand here. Note that there are many different function in SymPy as well as two different implementations of multivariate polynomials and the fact that any kind of polynomial can have multivariate polynomial coefficients...

Yeah I'm certainly realising the complexity of it. I'm first allowing a DMP_Flint to hold a fmpz_mpoly for the ZZ and QQ domains

> For resultant and discriminant we have these in SymPy:
> 
> ```python
> In [6]: Poly(x + y).resultant(Poly(x - y))
> Out[6]: Poly(-2*y, y, domain='ZZ')
> ```
> 
> In this case we need to go from a ring with 2 generators to a ring with 1 generator. That is not what Flint's resultant etc functions do so we would need some way of converting the `fmpz_mpoly` from `R[x,y] -> R[y]`.

These are the two functions I was referring to, in SymPy `DMP.resultants` and `DMP.subresultants` seem to operate without respect to a particular generator, while `PolyElement.subresultants` and `mpoly.resultant`s do. 


> Suppose that we had a way to get the coefficients of an `fmpz_mpoly` as an `fmpz_vec` and also that `fmpz_vec` had a `gcd` method. Then in python-flint (or in SymPy) you could do:
> 
> ```python
> def content(p):
>     content = p.coeffs_vec().gcd()
> ```

Is this not what `mpoly.term_content` computes?

> For resultant and discriminant we have these in SymPy:
> 
> ```python
> In [6]: Poly(x + y).resultant(Poly(x - y))
> Out[6]: Poly(-2*y, y, domain='ZZ')
> ```
> 
> In this case we need to go from a ring with 2 generators to a ring with 1 generator. That is not what Flint's resultant etc functions do so we would need some way of converting the `fmpz_mpoly` from `R[x,y] -> R[y]`.

```python
In [1]: from flint import *

In [5]: ctx = fmpz_mpoly_ctx.get_context(2, Ordering.lex, 'x')

In [6]: x, y = ctx.gens()\

In [7]: (x + y).resultant(x - y, 'x0')
Out[7]: -2*x1

In [8]: (x + y).resultant(x - y, 'x1')
Out[8]: 2*x0
```
This seems to align with what flint computes with respect to the first generator. Not sure if that is a coincidence. I've been having a little difficulty deciphering the DMP docs and implementations.

oscarbenjamin:
I don't know if you've read this:
https://docs.sympy.org/latest/modules/polys/domainsintro.html#dmp-representation

The DMP representation is a recursive representation of multivariate polynomials where a polynomial in say `QQ[x,y,z]` is actually represented like an element of `QQ[y,z][x]` or more explicitly `(QQ[y,z])[x]`. It is a univariate polynomial whose coefficients are polynomials in `QQ[y,z]`. Likewise the elements of `QQ[y,z]` are represented as elements of `QQ[z][y]` and so on.

What this means is that a `DMP` always has a leading variable so if you want to compute the resultant of two multivariate polynomials then it is always known which variable you want to use: the leading one or "main variable". Technically the Sylvester resultant is a concept that is *defined* for univariate polynomials and the resultant itself should be an element of the ring of coefficients e.g. resultant of two polynomials in `D[x]` is an element of `D` not of `D[x]`. It just happens to be *most useful* though to compute resultants of multivariate polynomials that are viewed as univariate by choosing a main variable:
https://en.wikipedia.org/wiki/Resultant

The DMP recursive representation makes it natural that an operation like `resultant` produces a polynomial in a different ring e.g. projecting from `QQ[x,y,z]` to `QQ[y,z]` (or from `QQ[y,z][x]` to `QQ[y,z]`):
```python
In [11]: p1 = Poly(x + y)

In [12]: p2 = Poly(x - y)

In [13]: p3 = p1.resultant(p2)

In [14]: p1
Out[14]: Poly(x + y, x, y, domain='ZZ')

In [15]: p2
Out[15]: Poly(x - y, x, y, domain='ZZ')

In [16]: p3
Out[16]: Poly(-2*y, y, domain='ZZ')
```
Notice that `p3` does not have `x` as a generator. The internal representation is different:
```python
In [19]: p1.rep.to_list()
Out[19]: [[1], [1, 0]]

In [20]: p2.rep.to_list()
Out[20]: [[1], [-1, 0]]

In [21]: p3.rep.to_list()
Out[21]: [-2, 0]
```
This is different from what FLINT's resultant function does because while it returns a polynomial free of `x` it is still represented as an element of the ring `ZZ[x,y,z]` i.e. it still has the same context. To implement what SymPy's resultant does there needs to be a way to convert a polynomial that is free of `x` from `ZZ[x,y,z]` to `ZZ[y,z]`.

`DMP` itself does not have any notion of what its generators are:
```python
In [22]: p1.rep
Out[22]: DMP_Python([[1], [1, 0]], ZZ)
```
The idea that the variables are "x" and "y" is maintained at a higher-level in the Poly type. What this means is that we need a general way to compute resultant of a polynomial wrt main variable and project down to the representation with the main variable removed from the context. We don't care what the variable names are and we always know which one is the main variable.

Note that I am not saying that what you have as a resultant function here is wrong for python-flint but just that it needs to be adapted somehow in SymPy and possibly something extra needs to be provided in python-flint to make that possible.

What I'm not sure about is how to hold the contexts for this sort of thing in memory. You could imagine a stack of contexts where the context for 3 variables holds a reference to the context for 2 variables etc. (I mean that this stack would be in SymPy, not in python-flint, although it could potentially make sense in python-flint as well).

oscarbenjamin:
> > Suppose that we had a way to get the coefficients of an `fmpz_mpoly` as an `fmpz_vec` and also that `fmpz_vec` had a `gcd` method. Then in python-flint (or in SymPy) you could do:
> > ```python
> > def content(p):
> >     content = p.coeffs_vec().gcd()
> > ```
> 
> Is this not what `mpoly.term_content` computes?

Almost but not quite:
```python
In [26]: (4*x + 2*x*y).term_content()
Out[26]: 2*x
```
It is the gcd of the terms rather than of the coefficients and so it also includes the gcd of the monomials. SymPy would give:
```python
In [31]: primitive(4*x + 2*x*y)
Out[31]: (2, x*y + 2*x)

In [32]: Poly(4*x + 2*x*y).primitive()
Out[32]: (2, Poly(x*y + 2*x, x, y, domain='ZZ'))
```
I suppose in python-flint it can be:
```python
In [34]: (4*x + 2*x*y).term_content().leading_coefficient()
Out[34]: 2
```
It is not ideal to compute the gcd of the monomials when it is not needed though. Note that computing the gcd of many things like `gcd(a, b, c, d, ...)` is often not as costly as you might imagine because usually the gcd turns out to be 1 fairly early and you can shortcut checking all of the arguments. If you take a random polynomial and multiply it by `x` though then it will be very likely that the gcd of the coefficients is 1 but the gcd of the terms will not be and so it will not be possible to shortcut in `term_content` but it would be in some analogous function for computing the gcd of the coefficients.

Internally FLINT computes the equivalent of `p.coeffs_vec().gcd()`:
https://github.com/flintlib/flint/blob/331171713cfb0c1b3e9dc4dbabb9849cbb3f7eb7/src/fmpz_mpoly/primitive_part.c#L33-L35
That operation is not exposed through a public function but we could make one in python-flint that does the same since it is straight-forward. Ideally we don't even need to make an `fmpz_vec` explicitly.

There are a few different natural notions of what "content" means when we are working with polynomials. For univariate polynomials there are two cases:

- We might mean gcd of coefficients (an element of `ZZ`)
- We might mean gcd of terms (a 1-term element of `ZZ[x,y,z]`).

Usually when someone says "content of a polynomial" they mean the gcd of coefficients:
https://en.wikipedia.org/wiki/Primitive_part_and_content

In a multivariate context we also need to define our interpretation of what is a coefficient i.e. whether some generators are viewed as being part of the coefficients or part of the monomials for the purpose of defining the content.

- We might choose a main variable and regard all other variables as being part of the coefficients rather than the terms e.g. we think of `ZZ[x,y,z]` as `ZZ[y,z][x]` and so the content is a polynomial in `ZZ[y,z]`
- We can also choose some subset of the variables to be generators and regard the remainder as being part of the coefficients e.g. we think of `ZZ[x,y,z,t]` as `ZZ[z,t][x,y]` and then the content is a polynomial in `ZZ[z,t]`.

The Flint functions related to this are:

- `fmpz_mpoly_term_content`. This treats all variables as generators but computes gcd of terms rather than coefficients. This is not usually what is meant by "content" and there is no analogue of this in SymPy.
- `fmpz_mpoly_content_vars`. This allows choosing a subset of variables to be generators and computing the content as a polynomial in the remaining generators. There is no direct analogue of this operation in SymPy because it would be done by converting to a representation that already distinguishes which variables are generators.
- `fmpz_mpoly_primitive_part`. This treats all variables as generators and uses the usual definition of content as just the gcd of the coefficients. This is directly analogous to what SymPy's content and primitive functions/methods do except that it does not return the content and also it handles sign normalisation differently.

There are several ways to adapt these functions so that they do what SymPy's equivalent functions would do:

- `c = p.term_content().leading_coefficient()` gives the usual content and then `prim = p / c` gives the primitive part. This is less efficient than calculating `c` directly.
- `c = p.content_vars([0,1,2,...,n-1])` also gives `c` but I think it is computed in a less efficient way by a more general algorithm that is designed to do something more complicated.
- `prim = p.primitive_part()` gives the usual primitive part and then `c = p / prim` gives the content but using a polynomial division here is less efficient.

None of these is quite optimal for just getting the normal idea of the "content" which is just the gcd of the integer coefficients. Generally `term_content` is probably more useful in many situations but still there are situations where we just want to factor out the gcd of the coefficients as an integer. I think that we just need to add a function that does this in python-flint using `_fmpz_vec_content` much like FLINT's `primitive_part` does.

What I'm not sure about is what names to give to methods that do these things. Probably there should be a method that is just called `content` for getting the gcd of the coefficients although we should check what existing names are used for any related poly or mpoly methods. In SymPy the analogous methods/functions are called:

- `c = content(p)` get the gcd of coefficients as an integer.
- `(c, prim) = primitive(p)` returns both content and primitive part such that `p == c*prim`.

It makes sense to have these two functions because it isn't possible to compute the primitive part without first computing the content so you might as well have a function called `primitive` that returns both.

Note also that we should not necessarily always follow exactly Flint's conventions for names of functions and methods etc. Flint itself is not even fully consistent across all types as in they may or may not have exactly the same methods which may or may not be defined to mean exactly the same things. Rather we should define methods that make sense consistently within python-flint from the perspective of someone using python-flint from Python. In some cases that means we need to piece together a few Flint functions to do something useful rather then just exposing precisely Flint's documented functions for each type.

We should also consider that in general it is useful if all types have the same methods e.g. if all polynomial types have primitive and content methods or even if those can be consistent across both univariate and multivariate polynomials. If we are going to have `term_content` for multivariate polynomials then we probably want it for univariate polynomials as well. It is easy to compute for univariate polynomials so I guess FLINT does not provide a function for it but it's nicer for users if the interface is as uniform as possible. Likewise if we have content and primitive for multivariate polynomials then we also want it for univariate polynomials and it should be defined consistently in both cases.

Note that in SymPy content and primitive are defined not just for polynomials with integer coefficients:
```python
In [5]: primitive(2*x/3 + 2)
Out[5]: (2/3, x + 3)

In [6]: primitive(2*x + 3, modulus=5)
Out[6]: (1, 2⋅x - 2)
```
The definition is potentially ambiguous over a field but it can still be useful to define some notion of "factor out a scalar".

oscarbenjamin:
> Note also that we should not necessarily always follow exactly Flint's conventions for names of functions and methods etc.

To summarise:

- `term_content` is a useful function with a reasonable name. It makes sense for all polynomial types so let's add that.
- We also want `content` though which gives the actual coefficient content and this should be for both `fmpz_poly` and `fmpz_mpoly` at least.
- `primitive_part` seems poorly designed to me. It should return the content instead of discarding it. I don't think we should copy that interface. It would be better to have a function `primitive` or `content_primitive` or something that returns both the content and the primitive part. We could also have a `primitive_part` method but it does not add much value compared to `content_primitive`.
- `deflation` is good but FLINT has defined it somewhat inconsistently for univariate and multivariate polynomials. I don't think that we should copy that inconsistency. We could have two versions like `deflation` and `shift_deflation` or something and then `deflation` for multivariate polynomials should be analogous to what it does with univariate polynomials.

oscarbenjamin:
> * We could have two versions like `deflation` and `shift_deflation` or something

Note that this is also sort of unnecessary because you can use `term_content` for mostly the same effect:
```python
In [1]: from flint import *

In [2]: ctx = fmpz_mpoly_ctx(2, Ordering.lex, ['x', 'y'])

In [3]: x, y = ctx.gens()

In [4]: f = x**3 * y + x * y**4 + x * y

In [5]: f
Out[5]: x^3*y + x*y^4 + x*y

In [6]: tc = f.term_content()

In [7]: tc
Out[7]: x*y

In [8]: fd, s = (f / tc).deflation()

In [9]: fd
Out[9]: x + y + 1

In [10]: s
Out[10]: fmpz_vec(['2', '3'], 2)

In [11]: tc * fd.compose(x**2, y**3)
Out[11]: x^3*y + x*y^4 + x*y

In [12]: tc * fd.compose(x**2, y**3) == f
Out[12]: True
```
Using `term_content` like this is not exactly equivalent because it also computes the gcd of the coefficients but in may cases that would be desired anyway. There could be a function (not sure of a good name) to compute the deflation like this:
```python
m, fd, s = f.shift_deflation()
assert m*fd.inflate(s) == f
```
Another point I realise now is that it would be fine to have `__call__` do the equivalent of what `compose` does. When we discussed this before the `__call__` function was doing lots of different things but I think now that `subs` is a separate function there is probably no reason that you shouldn't be able to do e.g.:
```python
In [13]: fd.compose(x**2, y**3)
Out[13]: x^2 + y^3 + 1

In [14]: fd(x**2, y**3) # maybe this could work
TypeError
```
That does work for univariate polynomials already:
```python
In [15]: x = fmpz_poly([0, 1])

In [16]: p = x**2 + 1

In [17]: p(p)
Out[17]: x^4 + 2*x^2 + 2
```

Jake-Moss:
> `deflation` is good but FLINT has defined it somewhat inconsistently for univariate and multivariate polynomials. I don't think that we should copy that inconsistency. We could have two versions like `deflation` and `shift_deflation` or something and then `deflation` for multivariate polynomials should be analogous to what it does with univariate polynomials.

Regarding this, what should these functions be? If I'm understanding correctly to make the interface consistent with the univariate polynomials the `shift` vector should optionally be zeros. So the `deflation` function could optionally accept the two vectors, `stride` and `shift` as arguments then if provided, use them, otherwise call `fmpz_mpoly_deflation` to compute the stride and shift, then call `fmpz_mpoly_deflate` with just the stride, ignoring the shift vector? 
To make the above interface useful we'd then also need to have another method to compute the stride and shift so they can be passed as arguments. On top of `deflation` we could then provide an `inflation` method that takes the stride as a required argument and optionally the shift to undo the deflation.

Jake-Moss:
> The DMP recursive representation makes it natural that an operation like `resultant` produces a polynomial in a different ring e.g. projecting from `QQ[x,y,z]` to `QQ[y,z]` (or from `QQ[y,z][x]` to `QQ[y,z]`):

That makes sense thank you. 

> This is different from what FLINT's resultant function does because while it returns a polynomial free of `x` it is still represented as an element of the ring `ZZ[x,y,z]` i.e. it still has the same context. To implement what SymPy's resultant does there needs to be a way to convert a polynomial that is free of `x` from `ZZ[x,y,z]` to `ZZ[y,z]`.

There should still be a context coercion function laying around somewhere. Here it is
```python

    # TODO: Rethink context conversions, particularly the proposed methods in #132
    def coerce_to_context(self, ctx):
        cdef:
            fmpz_mod_mpoly res
            slong *C
            slong i

        if not typecheck(ctx, fmpz_mod_mpoly_ctx):
            raise ValueError("provided context is not a fmpz_mod_mpoly_ctx")

        if self.ctx is ctx:
            return self

        C = <slong *> libc.stdlib.malloc(self.ctx.val.minfo.nvars * sizeof(slong *))
        if C is NULL:
            raise MemoryError("malloc returned a null pointer")
        res = create_fmpz_mod_mpoly(self.ctx)

        vars = {x: i for i, x in enumerate(ctx.py_names)}
        for i, var in enumerate(self.ctx.py_names):
            C[i] = <slong>vars[var]

        fmpz_mod_mpoly_compose_fmpz_mod_mpoly_gen(res.val, self.val, C, self.ctx.val, (<fmpz_mod_mpoly_ctx>ctx).val)

        libc.stdlib.free(C)
        return res
```
It works by composing the mpolys with the generators as the functions to convert between contexts. It's rather rough in this form but could be made to work. I'm not sure if Flint provides a better method than this.

> What I'm not sure about is how to hold the contexts for this sort of thing in memory. You could imagine a stack of contexts where the context for 3 variables holds a reference to the context for 2 variables etc. (I mean that this stack would be in SymPy, not in python-flint, although it could potentially make sense in python-flint as well).

I think context interop and coercion is something is missing from python-flint (and Flint). A tree of contexts and their variables could work but it would have to be lazy or risk blowing up for lots of variables. Either way to change domain we need a new context, I think creating one on the fly and coercing within SymPy is reasonable for now, just storing them in python-flints context cache. It would have to be an explicit operation that's only used either when required or user-side to prevent slow down.

oscarbenjamin:
> A tree of contexts and their variables could work but it would have to be lazy or risk blowing up for lots of variables

At least for DMP we don't need a tree. Remember that DMP does not care what the names of the variables are and only ever wants to compute the resultant wrt to the main variable. So it would be a singly linked list of contexts for different numbers of variables like the 100 variable context holds a reference to the 99 variable context and so on. I suppose that a `DMP_Flint` will need to have a reference to a `*_mpoly_ctx`. We could wrap the `_mpoly_ctx` in a Python class that could lazily generate and then cache a context for one fewer variable like:
```python
class Flint_mpoly_ctx:
    def __init__(self, ...):
        self.fctx = fmpz_mpoly_ctx(...)
        self.fctx_drop = None

   @property
   def drop_ctx(self):
       if self.fctx_drop is None:
            ...
      return self.fctx_drop
```
Alternatively we could also just `@lru_cache` the function that creates one of these contexts for DMP.

> I think context interop and coercion is something is missing from python-flint (and Flint).

It does look like `fmpz_mod_mpoly_compose_fmpz_mod_mpoly_gen` is the most direct function for this though. Internally it looks like it makes a matrix to perform the transformation but I would have expected some more direct way.

I think this is possibly something that the `univar` structs can be used for although I'm not sure:
https://flintlib.org/doc/fmpz_mpoly.html#univariate-functions

oscarbenjamin:
> So the `deflation` function could optionally accept the two vectors, `stride` and `shift` as arguments then if provided, use them, otherwise call `fmpz_mpoly_deflation` to compute the stride and shift, then call `fmpz_mpoly_deflate` with just the stride, ignoring the shift vector?

We already have the methods `inflate`, `deflate` and `deflation` for different univariate polynomial types although not all types have them. Obviously the multivariate ones need to return a list of integers for the deflation whereas the univariate ones return an int so there has to be that difference. Otherwise though we should try to keep them as consistent as possible.

I suggest the following:
```python
class foo_poly:

    def inflate(self, n: int) -> foo_poly:
        """q  s.t.  q(x) = p(x^n)"""

    def deflate(self, n: int) -> foo_poly:
        """q  s.t.  p(x) = q(x^n)"""

    def deflation(self) -> (foo_poly, int):
        """(q, n)  s.t.  p(x) = q(x^n)  for maximal n"""

    def deflation_monom(self) -> (foo_poly, int, foo_poly):
        """(q, n, m)  s.t.  p(x) = m * q(x^n)  for maximal n and monomial m"""

    def deflation_index(self) -> (foo_poly, int, int):
        """(n, i)  s.t.  p(x) = x^i * q(x^n)  for maximal n and i"""


class foo_mpoly:

    def inflate(self, N: list[int]) -> foo_mpoly:
        """q  s.t.  q(X) = p(X^N)"""

    def deflate(self, N: list[int]) -> foo_mpoly:
        """q  s.t.  p(X) = q(X^N)"""

    def deflation(self) -> (foo_mpoly, list[int]):
        """(q, N)  s.t.  p(X) = q(X^N)  for maximal N"""

    def deflation_monom(self) -> (foo_mpoly, list[int], foo_mpoly):
        """(q, N, m)  s.t.  p(X) = m * q(X^N)  for maximal N and monomial m"""

    def deflation_index(self) -> (foo_mpoly, list[int], list[int]):
        """(N, I)  s.t.  p(X) = X^I * q(X^N)  for maximal N and I"""
```
Then the invariants for both cases are like:
```python
assert p.inflate(n).deflate(n) == p

q, n = p.deflation()
assert q.inflate(n) == p

q, n, m = p.deflation_monom()
assert m * q.inflate(n) == p

n, i = p.deflation_index()
q = (p/x**i).deflate(n)
assert q.inflate(n) * x**i == p
```
The last case for `deflation_index` is a bit awkward with mpoly because you would need to somehow turn the list of indices `i` into a monomial but that is fine. Possibly we can do without `deflation_index` altogether or at least for now it is not necessary.

This keeps consistency with the different types in python-flint. The equivalent methods in SymPy correspond to `inflate` and `deflation` (called `deflate` in SymPy).

oscarbenjamin:
Apologies but there are merge conflicts after gh-217.

Jake-Moss:
I believe I've got that implemented for the fmpz polys. Here's a quick snippet
```python
In [1]: from flint import *

In [5]: ctx = fmpz_mpoly_ctx(2, Ordering.lex, ['x', 'y'])

In [6]: x, y = ctx.gens()

In [7]: f = x**3 * y + x * y**4 + x * y

In [8]: f.deflation_index()
Out[8]: (x + y + 1, [2, 3], [1, 1])

In [9]: f.deflation_monom()
Out[9]: (x + y + 1, [2, 3], x*y)

In [10]: q, N, m = f.deflation_monom()

In [11]: m * q.inflate(N)
Out[11]: x^3*y + x*y^4 + x*y

In [12]: m * q.inflate(N) == f
Out[12]: True
```

> The last case for `deflation_index` is a bit awkward with mpoly because you would need to somehow turn the list of indices `i` into a monomial but that is fine. Possibly we can do without `deflation_index` altogether or at least for now it is not necessary.

This wasn't bad with `fmpz_mpoly_push_term_ui_ffmpz` onto an empty fmpz_mpoly, though it would be nicer to not go fmpz_vec -> list -> fmpz_vec so I might duplicate the `deflation_index` function into `deflation_monom`.

> Apologies but there are merge conflicts after https://github.com/flintlib/python-flint/pull/217.

No worries, I'll handle those tomorrow as well as propagate the reset of the deflation/inflation changes and add docs for them.

> Another point I realise now is that it would be fine to have __call__ do the equivalent of what compose does. When we discussed this before the __call__ function was doing lots of different things but I think now that subs is a separate function there is probably no reason that you shouldn't be able to do e.g.:

I'm happy to bring this back to be consistent with fmpz_poly.

Jake-Moss:
I've just pushed a POC doctest runner that finds and executes the doc tests in a similar manner to the existing stuff. Easy to revert though.

I've flushed out the deflation/inflation functions again but I've run across some odd behaviour. Here's a MWE

```python
In [1]: from flint import *

In [5]: (x,) = fmpz_mpoly_ctx.get_context(1, Ordering.lex, 'x').gens()

In [6]: x.deflate([0])
Exception (fmpz_divexact). Division by zero.

Process Python aborted (core dumped)
```

This is occurring because of the division here https://github.com/flintlib/flint/blob/main/src/mpoly/monomials_deflate.c#L43
It is correct in that is it dividing by zero, however the docs say that 
> fmpz_mpoly_deflate
> Apply the function `e -> (e - shift[v])/stride[v]` to each exponent e corresponding to the variable v. If **any `stride[v]` is zero, the corresponding numerator `e - shift[v]` is assumed to be zero, and the quotient is defined as zero.** This allows the function to undo the operation performed by fmpz_mpoly_inflate() when possible.

The provided stride[0] is indeed `0`, so I thought that it would mean that `e - shift[v]` is assumed to be zero, and the division is not performed, instead defined as zero. However the `!fmpz_is_zero` check occurs in just the if-statement above https://github.com/flintlib/flint/blob/main/src/mpoly/monomials_deflate.c#L40 but it checks if `e - shift[v]` is zero or not, not `stride + j`, then promptly divides by zero. The comment doesn't make sense to me with the if-statement directly below.

I instead believe it should be something similar to
```patch
diff --git a/src/mpoly/monomials_deflate.c b/src/mpoly/monomials_deflate.c
index cc7b4809a..6de7b035d 100644
--- a/src/mpoly/monomials_deflate.c
+++ b/src/mpoly/monomials_deflate.c
@@ -37,11 +37,16 @@ void mpoly_monomials_deflate(ulong * Aexps, flint_bitcnt_t Abits,
         {
             fmpz_sub(exps + j, exps + j, shift + j);
             /* stride + j is allowed to be zero */
-            if (!fmpz_is_zero(exps + j))
+            if (!fmpz_is_zero(stride + j))
             {
                 FLINT_ASSERT(fmpz_divisible(exps + j, stride + j));
                 fmpz_divexact(exps + j, exps + j, stride + j);
             }
+            else
+            {
+                fmpz_zero(exps + j);
+            }
+
         }
         FLINT_ASSERT(Abits >= mpoly_exp_bits_required_ffmpz(exps, mctx));
         mpoly_set_monomial_ffmpz(Aexps + NA*i, exps, Abits, mctx);
```
but haven't had the case to test it thoroughly yet. But with that patch applied I get the output I expect
```python
In [1]: from flint import *

In [5]: (x,) = fmpz_mpoly_ctx.get_context(1, Ordering.lex, 'x').gens()

In [6]: x.deflate([0])
Out[6]: 1
```

Jake-Moss:
This is still missing the invariants as test cases and the merge conflicts to be resolved.

oscarbenjamin:
The doctest runner looks good. There is probably some way to make it so that pytest can discover the tests cases it collects. The simple way would just be to put them in a function called `test_something`. Perhaps it could be a parametrised test like:
```
@parametrised("module", all_modules)
def test_docstrings(mod):
   ...
```
Then pytest would show each docstring as a separate test case.

It would be good of `pytest --pyargs flint` worked to run the doctests as well as the main tests and also if `spin test` would do the doctests.

There are also one or two test cases in the .rst docs. Ideally there would be more but the docs are a bit sparse at the moment:
```console
$ pytest --doctest-glob='*.rst' doc/source/ -v
...
collected 1 item                                                                                    

doc/source/general.rst::general.rst PASSED                                                    [100%]

========================================= 1 passed in 0.64s =========================================
```

oscarbenjamin:
> I instead believe it should be something similar to

I think you are right and it is probably a Flint bug.

Jake-Moss:
> The doctest runner looks good. There is probably some way to make it so that pytest can discover the tests cases it collects. The simple way would just be to put them in a function called `test_something`. Perhaps it could be a parametrised test like:
> 
> ```
> @parametrised("module", all_modules)
> def test_docstrings(mod):
>    ...
> ```
> 
> Then pytest would show each docstring as a separate test case.

This is a great idea! I tried getting the pytest-cython plugin working but this is way easier and nicer. Here's some sample output. Only downside is there are a lot of dots, I'm sure we can live with that though.
```python
================================================ test session starts ================================================
platform linux -- Python 3.12.4, pytest-8.3.2, pluggy-1.5.0
rootdir: /home/jake/Uni/Honours/Thesis/python-flint
configfile: pyproject.toml
plugins: cov-5.0.0
collected 760 items                                                                                                 

flint/test/test_all.py ......................................................                                 [  7%]
flint/test/test_docstrings.py ............................................................................... [ 17%]
............................................................................................................. [ 31%]
............................................................................................................. [ 46%]
............................................................................................................. [ 60%]
............................................................................................................. [ 74%]
..............FF.....................................FF...................................................... [ 89%]
..................................................................................                            [100%]

===================================================== FAILURES ======================================================
__________________________________ test_docstrings[flint.types.fmpz_mpoly-test529] __________________________________
Failed: flint.types.fmpz_mpoly.deflate, line: 1125
Failed example:
	f.deflate([2, 3])
Expected:
	x + y + 1sdf
Got:
	x + y + 1
__________________________________ test_docstrings[flint.types.fmpz_mpoly-test530] __________________________________
Failed: flint.types.fmpz_mpoly.deflation, line: 1150
Failed example:
	q, N
Expected:
	(x + y + 1, [2, 3])sdf
Got:
	(x + y + 1, [2, 3])
__________________________________ test_docstrings[flint.types.fmpz_poly-test568] ___________________________________
Failed: flint.types.fmpz_poly.deflation_index, line: 657
Failed example:
	f.deflation()
Expected:
	(1, 1)
Got:
	(x + 1, 2)
__________________________________ test_docstrings[flint.types.fmpz_poly-test569] ___________________________________
Failed: flint.types.fmpz_poly.deflation_monom, line: 640
Failed example:
	f.deflation()
Expected:
	(1, x)
Got:
	(x + 1, 2)
============================================== short test summary info ==============================================
FAILED flint/test/test_docstrings.py::test_docstrings[flint.types.fmpz_mpoly-test529] - Failed: Failed: flint.types.fmpz_mpoly.deflate, line: 1125
FAILED flint/test/test_docstrings.py::test_docstrings[flint.types.fmpz_mpoly-test530] - Failed: Failed: flint.types.fmpz_mpoly.deflation, line: 1150
FAILED flint/test/test_docstrings.py::test_docstrings[flint.types.fmpz_poly-test568] - Failed: Failed: flint.types.fmpz_poly.deflation_index, line: 657
FAILED flint/test/test_docstrings.py::test_docstrings[flint.types.fmpz_poly-test569] - Failed: Failed: flint.types.fmpz_poly.deflation_monom, line: 640
========================================== 4 failed, 756 passed in 13.06s ===========================================
```

src/flint/test/test_all.py
```diff
@@ -3229,6 +3229,38 @@ def quick_poly():
...
+        assert res == x0**3 + 3*x0**2*x1 + x0**2 + 3*x0*x1**2 + 2*x0*x1 + x1**3 + x1**2 + 1
+        assert tuple(stride) == (2, 2)
+
+        for p in [f1, g1]:
+            n, m = p.deflation_monom()
+            assert m * p.deflate(n).inflate(n) == p
+
+            n, i = p.deflation_index()
+            m = ctx.term(exp_vec=i)
+            assert p.deflate(n).inflate(n) * m == p
```
oscarbenjamin, line 3258: These don't quite seem right. It should be:
```python
assert (p / m).deflate(n).inflate(n) * m = p
```
In other words `.deflate(n)` and `.inflate(n)` should be direct inverses.

In the first case I was expecting that `deflation_monom()` returns a deflated polynomial i.e.:
```python
pd, n, m = p.deflation_monom()
assert m * pd.inflate(n)
```
There is less value in having two functions for this if the only difference is whether `m` is a list of integers or a monomial since, as demonstrated here, it is not too difficult to convert between those forms.

oscarbenjamin:
It wouldn't be too bad to require pytest unconditionally. We could also add it to the requirements as an extra like `pip install python-flint[test]`.

One thing though is that when I am debugging a crash (core dump etc) that happens during the test suite the output from pytest is particularly unhelpful. Typically I have to add `print(1)`, `print(2)` etc throughout the code and then run with `python -m flint.test` so that I can actually see print output. Maybe there is a way to make that better with pytest.

oscarbenjamin:
> It wouldn't be too bad to require pytest unconditionally.

Although we need to ensure that it works after install with
```python
pytest --pyargs flint
```
and also we should ensure that the currently documented approach still works:
```
python -m flint.test
```

oscarbenjamin:
> Only downside is there are a lot of dots, I'm sure we can live with that though.

I think that the dots are a good thing. You can also pass `-v` to turn each into something that prints the filename. It is hardly needed for python-flint but with pytest-xdist the different tests can be run in parallel with `pytest -nauto`.

src/flint/test/test_all.py
Jake-Moss: > In other words .deflate(n) and .inflate(n) should be direct inverses.

There is something more wrong with the way I've implemented these functions then, but this doesn't seem possible by using a zero-shift vector. 

```python
In [47]: x.deflation()
Out[47]: (1, [0, 0])

In [48]: x.deflate([0, 0]).inflate([0, 0])
Out[48]: 1
```

When the exponent of a variable matches in all instances the shift vector then contains all the information, it's then discarded.

Jake-Moss: `deflation_monom` turns the shift vector into the monomial so I don't think it's an issue there. Let me re-review your pervious comments. I think I've misunderstood something

Jake-Moss: I think both the shift and stride vectors are strictly required. The multivariate deflation does more than it's univariate counter part because the univariate form does not allow mapping x^n -> x^0 while the multivariate form does.

oscarbenjamin: Can you not set the shifts to zero or is that the bug that you found?

I was imagining something like this:
```python
shifts, strides = foo_mpoly_deflation(...)

for i in range(n):
    g = gcd(shifts[i], strides[i])
    if g != 1: # maybe this check not needed?
        strides[i] = g
    shifts[i] = 0

p = foo_mpoly_deflate(shifts, strides, ...)
return p, strides
```

oscarbenjamin: We should also check what the `gr_poly` and `gr_mpoly` functions do for things like this but it seems that they don't define this yet...

Jake-Moss: > Can you not set the shifts to zero or is that the bug that you found?

I believe that was unrelated but might be the reason why that went undiscovered for 6 years. The current implementation sets the shift vector to zeros. But the returned stride vector is only correct when used with the shift vector.

> I was imagining something like this:
> 
> ```python
> shifts, strides = foo_mpoly_deflation(...)
> 
> for i in range(n):
>     g = gcd(shifts[i], strides[i])
>     if g != 1: # maybe this check not needed?
>         strides[i] = g
>     shifts[i] = 0
> 
> p = foo_mpoly_deflate(shifts, strides, ...)
> return p, strides
> ```

I've give that a shot

Jake-Moss:
> I think that the dots are a good thing. You can also pass `-v` to turn each into something that prints the filename. It is hardly needed for python-flint but with pytest-xdist the different tests can be run in parallel with `pytest -nauto`.

Previously they were identifiable in pytests verbose mode, I've just fixed that

Jake-Moss:
> > Only downside is there are a lot of dots, I'm sure we can live with that though.
> 
> I think that the dots are a good thing. You can also pass `-v` to turn each into something that prints the filename. It is hardly needed for python-flint but with pytest-xdist the different tests can be run in parallel with `pytest -nauto`.

These both work for me (when run under spin)

src/flint/test/test_all.py
Jake-Moss: This should be resolved now

src/flint/flint_base/flint_base.pyx
```diff
@@ -344,24 +344,24 @@ cdef class flint_mpoly_context(flint_elem):
         Create a key for the context cache via the number of variables, the ordering, and
         either a variable name string, or a tuple of variable names.
         """
-        # A type hint of `ordering: Ordering` results in the error "TypeError: an integer is required" if a Ordering
+        # A type hint of ``ordering: Ordering`` results in the error "TypeError: an integer is required" if a Ordering
         # object is not provided. This is pretty obtuse so we check its type ourselves
         if not isinstance(ordering, Ordering):
-            raise TypeError(f"`ordering` ('{ordering}') is not an instance of flint.Ordering")
+            raise TypeError(f"``ordering`` ('{ordering}') is not an instance of flint.Ordering")
```
oscarbenjamin, line 350: We shouldn't use single or double backticks in comments or error messages. They are only needed in docstrings or rst docs where they get rendered by sphinx.

src/flint/flint_base/flint_base.pxd
```diff
@@ -23,6 +23,7 @@ cdef class flint_mpoly(flint_elem):
     cdef _mul_mpoly_(self, other)
 
     cdef _divmod_mpoly_(self, other)
+    cdef _truediv_scalar_(self, other, assume_exact: bool)
```
oscarbenjamin, line 26: I would generally just make this two functions rather than having a bool flag so we have:
```
cdef _truediv_scalar_(self, other)
cdef _truediv_exact_scalar_(self, other)
```

src/flint/flintlib/functions/fmpz_vec.pxd
```diff
@@ -68,3 +68,6 @@ cdef extern from "flint/fmpz_vec.h":
     void _fmpz_vec_dot_general_naive(fmpz_t res, const fmpz_t initial, int subtract, const fmpz_struct * a, const fmpz_struct * b, int reverse, slong len)
     void _fmpz_vec_dot_general(fmpz_t res, const fmpz_t initial, int subtract, const fmpz_struct * a, const fmpz_struct * b, int reverse, slong len)
     void _fmpz_vec_dot(fmpz_t res, const fmpz_struct * vec1, const fmpz_struct * vec2, slong len2)
+
+    # Undocumented internal functions
+    void _fmpz_vec_content(fmpz_t res, const fmpz_struct * vec, slong len)
```
oscarbenjamin, line 73: We can't add these in the `functions/*.pxd` because these files are now all auto-generated by:
```console
$ bin/all_rst_to_pxd.sh ../flint/doc/source/
```
It will need to go in `flintlib/types` somewhere or otherwise would need to be special-cased in the `rst_to_pxd` script.

src/flint/flintlib/functions/nmod_mpoly.pxd
```diff
@@ -133,3 +133,8 @@ cdef extern from "flint/nmod_mpoly.h":
     void nmod_mpoly_div_monagan_pearce(nmod_mpoly_t polyq, const nmod_mpoly_t poly2, const nmod_mpoly_t poly3, const nmod_mpoly_ctx_t ctx)
     void nmod_mpoly_divrem_monagan_pearce(nmod_mpoly_t q, nmod_mpoly_t r, const nmod_mpoly_t poly2, const nmod_mpoly_t poly3, const nmod_mpoly_ctx_t ctx)
     void nmod_mpoly_divrem_ideal_monagan_pearce(nmod_mpoly_struct ** q, nmod_mpoly_t r, const nmod_mpoly_t poly2, nmod_mpoly_struct * const * poly3, slong len, const nmod_mpoly_ctx_t ctx)
+
+    # These functions seem to be undocumented but appear in the header.
+    void nmod_mpoly_deflation(fmpz_struct * shift, fmpz_struct * stride, const nmod_mpoly_t A, const nmod_mpoly_ctx_t ctx)
+    void nmod_mpoly_deflate(nmod_mpoly_t A, const nmod_mpoly_t B, const fmpz_struct * shift, const fmpz_struct * stride, const nmod_mpoly_ctx_t ctx)
+    void nmod_mpoly_inflate(nmod_mpoly_t A, const nmod_mpoly_t B, const fmpz_struct * shift, const fmpz_struct * stride, const nmod_mpoly_ctx_t ctx)
```
oscarbenjamin, line 140: These need to move as well.

Maybe we should have a `flintlib/types/undocumented.pxd`. Then we can get the docs fixed upstream.

src/flint/test/test_all.py
```diff
@@ -3229,6 +3229,43 @@ def quick_poly():
...
+            assert p.deflate(n).inflate(n) == p
+
+            pd, n, m = p.deflation_monom()
+            assert m * pd.inflate(n) == p
+
+            if not composite_characteristic:
+                n, i = p.deflation_index()
+                m = ctx.term(exp_vec=i)
+                assert (p / m).deflate(n).inflate(n) * m == p
+
```
oscarbenjamin, line 3264: This looks right.

src/flint/flintlib/functions/fmpz_vec.pxd
oscarbenjamin: See:
https://github.com/flintlib/python-flint/blob/14241cdd5c580b74eab2b9e010f8c0a0ae7e222c/src/flint/flintlib/README.md?plain=1#L1-L35

src/flint/types/fmpq_mpoly.pyx
```diff
@@ -914,6 +996,152 @@ cdef class fmpq_mpoly(flint_mpoly):
...
+            >>> f = x**3 * y + x * y**4 + x * y
+            >>> N, I = f.deflation_index()
+            >>> N, I
+            ([2, 3], [1, 1])
+            >>> f_deflated = f.deflate(N)
+            >>> f_deflated
+            x + y + 1
+            >>> m = ctx.term(exp_vec=I)
+            >>> m * f_deflated.inflate(N)
+            x^3*y + x*y^4 + x*y
```
oscarbenjamin, line 1135: I think it would be good to show what `m` is here because this function is a bit tricky to understand.

src/flint/types/fmpz_mpoly.pyx
```diff
@@ -750,10 +773,104 @@ cdef class fmpz_mpoly(flint_mpoly):
...
+
+            >>> from flint import Ordering
+            >>> ctx = fmpz_mpoly_ctx.get_context(2, Ordering.lex, 'x')
+            >>> x, y = ctx.gens()
+            >>> f = 4*x + 2*x*y
+            >>> f.primitive()
+            (2, x0*x1 + 2*x0)
+        """
+        cdef fmpz content = self.content()
+        return content, self._truediv_scalar_(content, True)
```
oscarbenjamin, line 868: I think this is the only place that uses the second argument assume_exact. This could just be changed to call
```
self._truediv_exact_scalar_(content)
```
Or Perhaps we can just call it:
```
self._divexact_scalar_(content)
```
Possibly good to line up with what FLINT calls it...

src/flint/types/fmpz_poly.pyx
```diff
@@ -593,18 +595,98 @@ cdef class fmpz_poly(flint_poly):
...
+
+    def deflation_monom(self) -> tuple[fmpz_poly, int, fmpz_poly]:
+        """
+        Compute the exponent vector ``n`` and monomial ``m`` such that ``p(x^(1/n))
+        = m * q(x^n)`` for maximal n. Importantly the deflation itself is not computed
+        here. The returned monomial allows the undo-ing of the deflation.
+
+            >>> f = fmpz_poly([1, 0, 1])
+            >>> f.deflation_monom()
+            (x^2 + 1, 1, x)
```
oscarbenjamin, line 653: In this case it is just an "exponent" rather than an "exponent vector".

oscarbenjamin:
I left a few comments. The only critical ones are about adding hand-written definitions in `flintlib/functions`. That can't work after gh-217. In gh-219 I have added a README in flintlib that explains this.

Otherwise this looks good. I think that the inflate etc functions are doing the right thing now and it is good that they are consistent across types.

src/flint/flintlib/functions/fmpz_vec.pxd
Jake-Moss: Makes sense. Looks like this function was picked up by the recent auto-generation so I've just removed it instead.

src/flint/flintlib/functions/nmod_mpoly.pxd
Jake-Moss: Sounds like a good idea. I've done just that

src/flint/types/fmpq_mpoly.pyx
```diff
@@ -914,6 +999,154 @@ cdef class fmpq_mpoly(flint_mpoly):
...
+        fmpz_mpoly_deflate(res.val.zpoly, self.val.zpoly, shift.val, stride.val, self.ctx.val.zctx)
+        fmpq_set(res.val.content, self.val.content)
+
+        return res, list(stride)
+
+    def deflation_monom(self) -> tuple[fmpq_mpoly, list[int], fmpq_mpoly]:
+        """
+        Compute the exponent vector ``N`` and monomial ``m`` such that ``p(X^(1/N))
+        = m * q(X^N)`` for maximal N. Importantly the deflation itself is not computed
+        here. The returned monomial allows the undo-ing of the deflation.
```
oscarbenjamin, line 1092: The statement that the deflation is not computed is not correct for this function.

src/flint/types/fmpz_poly.pyx
```diff
@@ -648,3 +730,14 @@ cdef class fmpz_poly(flint_poly):
...
+    def content(self):
+        """
+        Return the GCD of the coefficients of ``self``.
+
+            >>> fmpz_poly([3, 6, 0]).content()
+            3
+        """
+        cdef fmpz res = fmpz()
+        _fmpz_vec_content(res.val, self.val.coeffs, self.val.length)
+        return res
```
oscarbenjamin, line 743: We may need to think a bit about the sign convention for `primitive` and `content`. Typically the primitive part is defined to be a canonical representative in some sense which may mean that the content should absorb the sign although there are different ways of defining this.

oscarbenjamin:
Okay, this looks good.

Also the test runner changes are nice.

Let's get this in.

Thanks!

src/flint/types/fmpq_mpoly.pyx
Jake-Moss: Quick PR here https://github.com/flintlib/python-flint/pull/220

Issue #221, opened 2024-09-09
fq_default_poly test fails sometimes


oscarbenjamin:
I just tried running the test suite in parallel with pytest-xdist like:
```
$ spin test -- -n4
Invoking `build` prior to running tests:
$ meson compile -C build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /home/oscar/.pyenv/versions/python-flint-3.12/bin/ninja -C /home/oscar/current/active/python-flint/build
ninja: Entering directory `/home/oscar/current/active/python-flint/build'
ninja: no work to do.
$ meson install --only-changed -C build --destdir ../build-install
$ export PYTHONPATH="/home/oscar/current/active/python-flint/build-install/usr/lib/python3.12/site-packages"
$ /home/oscar/.pyenv/versions/3.12.0/envs/python-flint-3.12/bin/python3.12 -P -c 'import flint'
$ cd /home/oscar/current/active/python-flint/build-install/usr/lib/python3.12/site-packages
$ /home/oscar/.pyenv/versions/3.12.0/envs/python-flint-3.12/bin/python3.12 -P -m pytest --import-mode=importlib --pyargs -n4
======================================== test session starts ========================================
platform linux -- Python 3.12.0, pytest-8.3.2, pluggy-1.5.0
rootdir: /home/oscar/current/active/python-flint
configfile: pyproject.toml
plugins: doctestplus-1.2.1, xdoctest-1.2.0, hypothesis-6.88.4, xdist-3.4.0, cov-5.0.0
4 workers [760 items]   
............................................................................................. [ 12%]
......F...................................................................................... [ 24%]
............................................................................................. [ 36%]
............................................................................................. [ 48%]
............................................................................................. [ 61%]
............................................................................................. [ 73%]
............................................................................................. [ 85%]
............................................................................................. [ 97%]
................                                                                              [100%]
============================================= FAILURES ==============================================
_______________________________________ test_fq_default_poly ________________________________________
...
            # series
            f_non_square = R_test([nqr, 1, 1, 1])
            f_zero = R_test([0, 1, 1, 1])
            assert raises(lambda: f_non_square.sqrt_trunc(1), ValueError)
            assert raises(lambda: f_zero.sqrt_trunc(1), ZeroDivisionError)
            assert raises(lambda: f_non_square.inv_sqrt_trunc(1), ValueError)
            assert raises(lambda: f_zero.inv_sqrt_trunc(1), ZeroDivisionError)
[...]

PR #222, opened 2024-09-09
fix faulty test in fq_default_poly

GiacomoPope:
In response to #221 I have amended the test for the error when trying to invert polynomials where the constant coefficient is zero (only case where it is non-invertible)


oscarbenjamin:
Thanks

PR #251, opened 2025-01-27
Add PrecisionManager to flint context

Tom-Hubrecht:
This allows temporarily changing the working precision using python context managers.

Related: #5

### DIFF
--- a/src/flint/flint_base/flint_context.pyx
+++ b/src/flint/flint_base/flint_context.pyx
@@ -57,6 +57,18 @@ cdef class FlintContext:
         assert num >= 1 and num <= 64
         flint_set_num_threads(num)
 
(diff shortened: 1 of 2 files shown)


src/flint/flint_base/flint_context.pyx
```diff
@@ -69,4 +153,46 @@ cdef class FlintContext:
     def cleanup(self):
         flint_cleanup()
 
+
+class PrecisionManager:
+    def __init__(self, ctx, eprec=None, edps=None):
```
oscarbenjamin, line 158: I think this should be a cdef class for speed.

oscarbenjamin:
Looks good. Thanks

PR #254, opened 2025-01-28
gr: Improve conversions

Tom-Hubrecht:
C.f. https://flintlib.org/doc/gr.html#assignment-and-conversions

We can use more appropriate functions for conversions than roundtripping through a string

### DIFF
--- a/src/flint/types/_gr.pyx
+++ b/src/flint/types/_gr.pyx
@@ -224,11 +224,23 @@ cdef class gr_ctx(flint_ctx):
     def __call__(self, arg) -> gr:
         """Create a new element of the domain.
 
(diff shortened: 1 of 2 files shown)


src/flint/types/_gr.pyx
```diff
@@ -229,6 +229,13 @@ cdef class gr_ctx(flint_ctx):
         >>> ctx(2)
         2
         """
+        if isinstance(arg, gr):
+            return self.from_other(arg)
+        if arg is int:
+            return self.from_si(arg)
+        if arg is float:
+            return self.from_d(arg)
```
oscarbenjamin, line 237: Should this not be something like `if type(arg) is int`?

oscarbenjamin:
Looks good

PR #260, opened 2025-02-07
test: reenable setuptools coverage build

oscarbenjamin:
See also https://github.com/cython/cython/issues/6658


oscarbenjamin:
So the situation seems to be that the coverage build with setuptools works apart from a Python 3.13 problem due to https://github.com/cython/cython/issues/6658 which introduced the change somewhere in between cython 3.0.11 (latest release) and 3.1.0a1.

That same bug applies to the meson coverage build. Unfortunately the meson coverage build already requires https://github.com/cython/cython/pull/6341 so 3.0.11 is not usable. Because of another Cython bug the first usable commit is https://github.com/cython/cython/commit/fdbca9960dc484dd21b580dac72074659c5d6047 which also precedes https://github.com/cython/cython/issues/6658. 

However even with that Cython commit there is still another Python 3.13 related bug. The problem that I fixed in https://github.com/cython/cython/pull/6341 seems to have now reemerged but as a Python 3.13 only problem where the tracer outputs basenames for some files. There must be some 3.13-only paths in Cython that have reintroduced the basename bug.

All in all then the situation is that:

- the meson coverage build works only for Python < 3.13 and can work with 3.1.0a1 as the earliest working release.
- The setuptools coverage build works for all Python versions with Cython 3.0.11 but does not work for Python 3.13 as of Cython 3.1.0a1.

PR #262, opened 2025-02-27
WASM build

### LINKED ISSUE #234: Nightly wheels and WASM/pyodide builds
CC @agriyakhetarpal

See also:
https://github.com/scientific-python/upload-nightly-action/issues/111

We should upload nightly wheels after each merge to main to the scientific-python nightly wheels index:
https://anaconda.org/scientific-python-nightly-wheels

That is useful for downstream testing e.g. in SymPy. It would also make it possible to have the latest main branches of python-flint and SymPy available for testing in SymPy Live:
https://live.sympy.org/

We should also build and test python-flint in WASM with pyodide in CI. There already are WASM builds of python-flint in pyodide:
https://pyodide.org/en/stable/usage/packages-in-pyodide.html

Those can be used in the browser e.g. at:
https://pyodide.org/en/stable/console.html
```python
>>> import flint
>>> x = flint.fmpz_poly([0, 1])
>>> p = x**2 + 2*x + 1
>>> p
x^2 + 2*x + 1
>>> p.factor()
(1, [(x + 1, 2)])
```
It looks like cibuildwheel could build these with `--platform pyodide`:
https://cibuildwheel.pypa.io/en/stable/options/

We would first need emscripten builds of GMP, MPFR and FLINT. The build recipes for those can be found in the pyodide repo e.g.:
https://github.com/pyodide/pyodide/tree/main/packages/flint

Probably a separate CI workflow should be used for making and testing pyodide wheels because we can't upload them to PyPI and the build commands and CI setup would be a bit different. We can upload the WASM wheels to the nightly wheels index though and the final release versions are already built in pyodide.

agriyakhetarpal:
Closes #234

- [x] Build
	- [x] libgmp
	- [x] libmpfr
	- [x] flint
	- [x] python-flint
- [x] Test


agriyakhetarpal:
I've been testing it a bit through a PR in my fork, and it all builds well, up to one point in the Cython sources where it fails in step 107/114: https://github.com/agriyakhetarpal/python-flint/actions/runs/13576857716/job/37954995494?pr=1

with the following error trace:

```
FAILED: src/flint/types/_gr.cpython-312-wasm32-emscripten.so.p/meson-generated_src_flint_types__gr.pyx.c.o 
/tmp/tmp8t79rr74/cc -Isrc/flint/types/_gr.cpython-312-wasm32-emscripten.so.p -Isrc/flint/types -I../src/flint/types -I/opt/hostedtoolcache/Python/3.12.9/x64/include/python3.12 -I/home/runner/work/python-flint/python-flint/wasm-library-dir/include -fvisibility=hidden -fdiagnostics-color=always -DNDEBUG -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -O3 -fPIC -MD -MQ src/flint/types/_gr.cpython-312-wasm32-emscripten.so.p/meson-generated_src_flint_types__gr.pyx.c.o -MF src/flint/types/_gr.cpython-312-wasm32-emscripten.so.p/meson-generated_src_flint_types__gr.pyx.c.o.d -o src/flint/types/_gr.cpython-312-wasm32-emscripten.so.p/meson-generated_src_flint_types__gr.pyx.c.o -c src/flint/types/_gr.cpython-312-wasm32-emscripten.so.p/src/flint/types/_gr.pyx.c
src/flint/types/_gr.cpython-312-wasm32-emscripten.so.p/src/flint/types/_gr.pyx.c:19051:65: error: incompatible integer to pointer conversion passing 'mp_limb_t' (aka 'unsigned long') to parameter of type 'const fmpz *' (aka 'const long *') [-Wint-conversion]
 19051 |   gr_ctx_init_fq_nmod(__pyx_v_ctx->__pyx_base.__pyx_base.ctx_t, __pyx_v_p, __pyx_v_d, __pyx_v_name);
       |                                                                 ^~~~~~~~~
/home/runner/work/python-flint/python-flint/wasm-library-dir/include/flint/gr.h:1366:53: note: passing argument to parameter 'p' here
 1366 | void gr_ctx_init_fq_nmod(gr_ctx_t ctx, const fmpz_t p, slong d, const char * var);
      |                                                     ^
src/flint/types/_gr.cpython-312-wasm32-emscripten.so.p/src/flint/types/_gr.pyx.c:19150:65: error: incompatible integer to pointer conversion passing 'mp_limb_t' (aka 'unsigned long') to parameter of type 'const fmpz *' (aka 'const long *') [-Wint-conversion]
 19150 |   gr_ctx_init_fq_zech(__pyx_v_ctx->__pyx_base.__pyx_base.ctx_t, __pyx_v_p, __pyx_v_d, __pyx_v_name);
       |                                                                 ^~~~~~~~~
/home/runner/work/python-flint/python-flint/wasm-library-dir/include/flint/gr.h:1367:53: note: passing argument to parameter 'p' here
 1367 | void gr_ctx_init_fq_zech(gr_ctx_t ctx, const fmpz_t p, slong d, const char * var);
      |                                                     ^
2 errors generated.
```

agriyakhetarpal:
FWIW; I hardly have any Cython development experience, so I will have to wrap my head around this a bit, unless someone has an idea for a fix. The basis for the failure here is that WASM is stricter about type conversions than compilation for other platforms, which is why this mismatch is causing errors.

The cause of the error, more specifically, is that `gr_fq_nmod_ctx` and `gr_fq_zech_ctx` pass a `ulong` parameter `p` 

https://github.com/flintlib/python-flint/blob/34dbb5d0a943576a6e7131aa69cc90a1f827bf9e/src/flint/types/_gr.pxd#L1189-L1202

but `flint/gr.h` has `fmpz_t` in version 3.0.1: https://github.com/flintlib/flint/blob/c05007be863b1aae7221f11a9c135d673d968638/src/gr.h#L1366-L1367

and this seems to have been fixed in version 3.1.0. I'll try updating to it.

agriyakhetarpal:
FWIW; I hardly have any Cython development experience, so I will have to wrap my head around this a bit, unless someone has an idea for a fix. The basis for the failure here is that WASM is stricter about type conversions than compilation for other platforms, which is why this mismatch is causing errors.

agriyakhetarpal:
Okay – the cause of the error, more specifically, is that `gr_fq_nmod_ctx` and `gr_fq_zech_ctx` pass a `ulong` parameter `p` 

https://github.com/flintlib/python-flint/blob/34dbb5d0a943576a6e7131aa69cc90a1f827bf9e/src/flint/types/_gr.pxd#L1189-L1202

but `flint/gr.h` has `fmpz_t` in version 3.0.1: https://github.com/flintlib/flint/blob/c05007be863b1aae7221f11a9c135d673d968638/src/gr.h#L1366-L1367

and this has apparently been fixed in version v3.1.0. I'll try updating `flint` to v3.1.0 (before proceeding to try out v3.1.2, if I do)

agriyakhetarpal:
All building: https://github.com/agriyakhetarpal/python-flint/actions/runs/13577635441?pr=1 ✨

The next task is to run tests.

agriyakhetarpal:
Successful build: https://github.com/agriyakhetarpal/python-flint/actions/runs/13577635441?pr=1 ✨

The next task is to run tests, which I can take a look at later in the day.

oscarbenjamin:
I haven't looked through this yet but thanks!

oscarbenjamin:
> but `flint/gr.h` has `fmpz_t` in version 3.0.1:

Best to keep with latest FLINT version for the pyodide build. There is no need to support a range of versions in pyodide. Generally python-flint tracks the latest version but has some `#ifdef` type stuff for older versions so that someone could build against them. No one is going to build against old versions in WASM though.

agriyakhetarpal:
Sounds good to me! I think the tests will have a bunch of function signature mismatches, though 😬 based on what I noticed through my fork's run. Also, building itself takes ~20 minutes in total – we could get to half of that if we have prebuilt WASM binaries. I guess they could be hosted somewhere in a GitHub release when things are ready.

agriyakhetarpal:
This can take a bit of time. I'm disabling tests and running it in on my fork to see how widespread the signature mismatches are, and I managed to get `test_all.py` to run till the end – now we have some more in `test_docstrings.py`. My next tasks will be to debug these mismatches, which I haven't had fun doing, as, i. compilation of these libraries to WASM is messy on a macOS and ii. a test here contains several assert statements, instead of several tests containing a few of them. Please bear with me as I proceed with this slowly and steadily. We could also wholly ignore `test_docstrings.py` if you're not too keen on that being tested. (I think it's good to test, though).

oscarbenjamin:
I don't quite understand what the problem with the tests is. Is it just caused by the Cython/C signatures not matching?

There is now FLINT 3.2.0-rc1. We could perhaps update all the Cython declarations to that version and use that for the WASM build. I'm not sure when 3.2.0 final will be released but we would want to update to that straight away when it is.

agriyakhetarpal:
> I don't quite understand what the problem with the tests is. Is it just caused by the Cython/C signatures not matching?

Yes, majorly. If there's even a slight mismatch between the upstream signatures for Flint and the ones we have here, WASM's type safety guarantees that it will terminate the program, which makes Pyodide raise a fatal error. These can happen at the time of linking (like when it happened with version 3.0.1 above), or at runtime, like we are facing now. I have to drop these completely instead of xfailing them as the running Pyodide interpreter is no longer possible to be used after it has encountered fatal errors.

There are so many of these with SciPy, especially from its wrapping of Fortran libraries that we can't compile to WASM directly without having to `f2c` them over. I hope the new Flang from LLVM-19 can do this: https://github.com/pyodide/pyodide/issues/5268

> There is now FLINT 3.2.0-rc1. We could perhaps update all the Cython declarations to that version and use that for the WASM build. I'm not sure when 3.2.0 final will be released but we would want to update to that straight away when it is.

Yes, this sounds good to me, and I hope they've fixed a few of these upstream. I could start another PR for this when I have enough time to do so; I assume it will help me learn Cython a bit. :)

oscarbenjamin:
I've opened gh-264 to bump the FLINT version to 3.2.0-rc1 and update all of the Cython declarations.

The Cython declarations are set from the FLINT docs by running the `bin/all_rst_to_pxd.py` script. There can often be mismatches resulting from parsing the signatures from the docs rather than the actual C code e.g. the docs may be inaccurate but usually in the Cython -> C -> compiled extension modules the differences even out largely because a lot of FLINT's functions are not actually called directly by python-flint.

agriyakhetarpal:
Thank you! Let's see how this goes; I've improved the caching for the CI job, though I've kept the failing tests skipped. I have been facing trouble building GMP locally with my installed Emscripten toolchain, let alone the rest of the libraries, so CI is my testbed for now.

oscarbenjamin:
It is expected that the ARM Linux jobs will fail after gh-264.

I think that the "Test coverage meson build" job is failing here because of this line that should be deleted:
https://github.com/flintlib/python-flint/blob/242c48e45f52bbeb72ae80687cde5adce681a2cc/coverage_plugin.py#L131
Somehow the changes here to `test_all.py` are causing that `print` to happen at a different time.

agriyakhetarpal:
The WASM job says

`configure: error: Cannot determine label suffix`

coming from

`emconfigure: error: './configure --disable-dependency-tracking --disable-shared --prefix=/home/runner/work/python-flint/python-flint/wasm-library-dir --with-gmp=/home/runner/work/python-flint/python-flint/wasm-library-dir --with-mpfr=/home/runner/work/python-flint/python-flint/wasm-library-dir' failed (returned 1)`

.github/workflows/ci-emscripten.yml
```diff
@@ -0,0 +1,153 @@
...
+          tar -xf flint-3.2.0-rc1.tar.xz
+
+          cd flint-3.2.0-rc1
+
+          emconfigure ./configure \
+            --disable-dependency-tracking \
+            --disable-shared \
+            --prefix=${{ env.WASM_LIBRARY_DIR }} \
+            --with-gmp=${{ env.WASM_LIBRARY_DIR }} \
+            --with-mpfr=${{ env.WASM_LIBRARY_DIR }}
```
oscarbenjamin, line 112: I think you need to set something like `--disable-assembly` or `--host=...` here.

oscarbenjamin: I don't know what the host triple is for WASM

oscarbenjamin:
The CMake build is not recommended except for Windows. I don't know if it would work.

oscarbenjamin:
> The WASM job says
> 
> `configure: error: Cannot determine label suffix`

The label suffix is used by the macro preprocessor that runs on the assembly code. I don't think that there is any assembly code for WASM. This is why I suggested `--disable-assembly`.

agriyakhetarpal:
Ah, wait, missed that comment

agriyakhetarpal:
I think it wouldn't help, as we have a warning as follows:

```
configure: WARNING: Currently no assembly available for 32-bit systems. Disabling assembly...
```

so `--disable-assembly` might be redundant. I can manually inspect the diff between the Makefiles for both versions if that brings something, or downgrade to version 3.1.3. Building version 3.2.0rc1 would be ideal, though.

agriyakhetarpal:
The most rudimentary reason for this is because the compiler is misconfigured – or it's probably coming from https://github.com/flintlib/flint/commit/76c90d53477fd4817d30cf2d681bb5588bdf1512.

albinahlback:
> This was working before with FLINT 3.1 but now fails with 3.2.0-rc1.

> As I understand it determining the label suffix is not needed when building for WASM since assembly will be disabled anyway. Maybe that has changed since I last looked at it though...

It is needed for `MPN_IORD_U`.

albinahlback:
Disable assembly does not mean disabling assembly completely, it just means disable compiling assembly files.  Perhaps this could be made more clear, but that's the meaning behind it.

oscarbenjamin:
> It is needed for `MPN_IORD_U`.

From what I see it looks like this is only defined for `__amd64__` or `__aarch64__` so for WASM this MACRO doesn't exist.

What should we do in the WASM case?

albinahlback:
I'll fix this before releasing 3.2.0.  Please open up an issue at FLINT.

oscarbenjamin:
I've opened https://github.com/flintlib/flint/issues/2247 for this.

oscarbenjamin:
We should test FLINT main after https://github.com/flintlib/flint/pull/2245

.github/workflows/ci-emscripten.yml
```diff
@@ -0,0 +1,155 @@
...
+      - name: Build flint
+        if: steps.cache-wasm-library-dir.outputs.cache-hit != 'true'
+        env:
+          CFLAGS: "-fPIC"
+          WASM_LIBRARY_DIR: ${{ github.workspace }}/wasm-library-dir
+        run: |
+          curl -L https://github.com/flintlib/flint/releases/download/v3.2.0-rc1/flint-3.2.0-rc1.tar.xz -o flint-3.2.0-rc1.tar.xz
+          tar -xf flint-3.2.0-rc1.tar.xz
+
+          cd flint-3.2.0-rc1
```
oscarbenjamin, line 105: We can checkout FLINT main here,

.github/workflows/ci-emscripten.yml
```diff
@@ -0,0 +1,155 @@
...
+
+          echo "PKG_CONFIG_PATH=${PKG_CONFIG_PATH}"
+          echo "CFLAGS=${CFLAGS}"
+          echo "LDFLAGS=${LDFLAGS}"
+
+          pkg-config --modversion python3
+          pkg-config --modversion mpfr
+          pkg-config --modversion flint
+
+          pyodide build
```
oscarbenjamin, line 140: I'm not sure what arguments need to be passed here but we need to tell meson to allow other flint versions with `-Dflint_version_check=false`.

.github/workflows/ci-emscripten.yml
agriyakhetarpal: Addressed in 21f6560e1b694f08388747ff2a31428b296fd411

.github/workflows/ci-emscripten.yml
agriyakhetarpal: `pyodide build` can pass config settings to pass on to the build backend. I added it in 382cdecda85f81b07a71488fbacdea08621b49ad.

agriyakhetarpal:
Thanks, let's see!

oscarbenjamin:
It sat no file `./configure`. We need `./bootstrap.sh` first.

agriyakhetarpal:
I'm confused as to why this wasn't the case for the release tarballs, as they are also from the GitHub source. Either way, pushed.

.github/workflows/ci-emscripten.yml
```diff
@@ -0,0 +1,158 @@
...
+        run: |
+          ./bootstrap.sh
+          emconfigure ./configure \
+            --disable-dependency-tracking \
+            --disable-shared \
+            --prefix=${{ env.WASM_LIBRARY_DIR }} \
+            --with-gmp=${{ env.WASM_LIBRARY_DIR }} \
+            --with-mpfr=${{ env.WASM_LIBRARY_DIR }}
+            --host=wasm32-unknown-emscripten
+            --disable-assembly
```
oscarbenjamin, line 117:
```suggestion
          emconfigure ./configure \
            --disable-dependency-tracking \
            --disable-shared \
            --prefix=${{ env.WASM_LIBRARY_DIR }} \
            --with-gmp=${{ env.WASM_LIBRARY_DIR }} \
            --with-mpfr=${{ env.WASM_LIBRARY_DIR }} \
            --host=wasm32-unknown-emscripten \
            --disable-assembly
```
Maybe this is ignoring the `--host` argument because it does not run the whole command?

oscarbenjamin: Yes, this was the problem.

.github/workflows/ci-emscripten.yml
```diff
@@ -0,0 +1,158 @@
...
+          emconfigure ./configure \
+            --disable-dependency-tracking \
+            --disable-shared \
+            --prefix=${{ env.WASM_LIBRARY_DIR }} \
+            --with-gmp=${{ env.WASM_LIBRARY_DIR }} \
+            --with-mpfr=${{ env.WASM_LIBRARY_DIR }} \
+            --host=wasm32-unknown-emscripten \
+            --disable-assembly
+          emmake make -j $(nproc)
+          emmake make install
```
oscarbenjamin, line 119:
```suggestion
      - name: Build flint
        if: steps.cache-wasm-library-dir.outputs.cache-hit != 'true'
        env:
          CFLAGS: "-fPIC"
          WASM_LIBRARY_DIR: ${{ github.workspace }}/wasm-library-dir
        working-directory: flint
        run: |
          curl -L https://github.com/flintlib/flint/releases/download/v3.2.0-rc1/flint-3.2.0-rc1.tar.xz -o flint-3.2.0-rc1.tar.xz
          tar -xf flint-3.2.0-rc1.tar.xz
          cd flint-3.2.0-rc1

          emconfigure ./configure \
            --disable-dependency-tracking \
            --disable-shared \
            --prefix=${{ env.WASM_LIBRARY_DIR }} \
            --with-gmp=${{ env.WASM_LIBRARY_DIR }} \
            --with-mpfr=${{ env.WASM_LIBRARY_DIR }} \
            --host=wasm32-unknown-emscripten \
            --disable-assembly
          emmake make -j $(nproc)
          emmake make install
```

.github/workflows/ci-emscripten.yml
```diff
@@ -0,0 +1,154 @@
...
+            --prefix=${{ env.WASM_LIBRARY_DIR }}
+          emmake make -j $(nproc)
+          emmake make install
+
+      - name: Build flint
+        if: steps.cache-wasm-library-dir.outputs.cache-hit != 'true'
+        env:
+          CFLAGS: "-fPIC"
+          WASM_LIBRARY_DIR: ${{ github.workspace }}/wasm-library-dir
+        working-directory: flint
```
oscarbenjamin, line 100:
```suggestion
```

oscarbenjamin:
> > We should test FLINT main after [flintlib/flint#2245](https://github.com/flintlib/flint/pull/2245)
> 
> I'm unsure now whether this was actually necessary because the error was caused by not having a line continuation character in the configure command.

I've just tested with FLINT 3.2.0-rc1 and the label problem came back. Looks like the fix in flintlib/flint#2245 is necessary.

oscarbenjamin:
Okay this builds again!

It fails when running the tests with:
```
Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
The cause of the fatal error was:
Error: Dynamic linking error: cannot resolve symbol pthread_getaffinity_np
```
Does wasm/emscripten have threads?

Maybe we need to `--disable-pthread`.

oscarbenjamin:
> Maybe we need to `--disable-pthread`.

Okay some tests are passing now. It fails in `test_fmpz_mod` with:
```
Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
The cause of the fatal error was:
RuntimeError: null function or function signature mismatch
```
Unfortunately there is nothing to indicate the name of the function...

agriyakhetarpal:
> Does wasm/emscripten have threads?
> 
> Maybe we need to `--disable-pthread`.

Yes, Emscripten supports pthreads, but when in conjunction with dynamic linking, not that well. Pyodide does not support them at the moment (xref https://github.com/pyodide/pyodide/issues/237). `--disable-pthread` should be the way to go.

agriyakhetarpal:
> It fails in `test_fmpz_mod` with:
> 
> ```
> Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
> The cause of the fatal error was:
> RuntimeError: null function or function signature mismatch
> ```
> 
> Unfortunately there is nothing to indicate the name of the function...

We'll need to sit with a debug build of Pyodide and FLINT and try it out in a browser. I haven't been able to try that out in a while, but I can spend some time doing it today and share my observations.

We also need to fix `test_polys`, `test_fq_default`, and `test_factor_poly_mpoly` – I agree it's a bit messy especially when it terminates the program fully...

oscarbenjamin:
> We also need to fix `test_polys`, `test_fq_default`, and `test_factor_poly_mpoly`

These tests are awkward because they have so many calls to different functions. I don't know how to debug any of this but if commenting out test code to bisect the problem then those tests are not great candidates. The `test_polys` and `test_factor_poly_mpoly` tests are designed to use as many types and functions as possible.

Right now this PR uses FLINT main to get the fix from https://github.com/flintlib/flint/pull/2245 but the declarations were updated to FLINT 3.2.0-rc1 in gh-264. There might be incompatible signatures because of that difference (main is ahead of the 3.2 branch).

oscarbenjamin:
Why don't we see all these signature mismatch problems with the version of python-flint that is currently in pyodide?

agriyakhetarpal:
We don't run the full test suite inside Pyodide; just a few functions that are supposed to act as sanity checks. We consider it the responsibility of the developers/recipe maintainers to ensure that the entire test suite runs and passes inside WASM with out of tree builds such as this PR. This can probably change with https://github.com/pyodide/pyodide-recipes when we have that ready, as it eliminates the CI time concern.

agriyakhetarpal:
Should we update the declarations against `flint` `main` here and try them out?

oscarbenjamin:
> Should we update the declarations against `flint` `main` here and try them out?

I'm not yet convinced that the difference matters. It is more likely that the differences are to do with something else. So far I don't have a clear idea though of what sort of difference matters because in this PR large parts of the test suite are commented out. I think it would be better to comment out tests like in 29fb4d87835036d3f51a403ca1f5291055966170 until we have something working and then try to hone in on one example of what is a signature mismatch and what it takes to fix it.

One thing that worries me is that there is a sort of basic type mismatch that is deep in python-flint which is that the fmpz type in FLINT is not something that is expressible in normal C. It is either an inline fixed width integer or a pointer to a multiprecision integer (like a tagged pointer). We have to lie to Cython about what this is and that works with most C compilers but I am a bit worried that pyodide's stricter runtime checking might be a problem with this.

oscarbenjamin:
Many tests are commented but otherwise the tests now pass.

oscarbenjamin:
There might just be a single fmpz_mod function that causes almost all the problems.

oscarbenjamin:
It seems to crash on the third or fourth of these lines:
https://github.com/flintlib/python-flint/blob/242c48e45f52bbeb72ae80687cde5adce681a2cc/src/flint/test/test_all.py#L1645-L1648
That is just constructing the fmpz_mod context. It fails for the larger modulus but the same functions (in terms of FLINT's C API) are called for all moduli. Looking at it though I see:
https://github.com/flintlib/python-flint/blob/242c48e45f52bbeb72ae80687cde5adce681a2cc/src/flint/types/fmpz_mod.pyx#L35-L49
I think that the call to `fmpz_mod_discrete_log_pohlig_hellman_clear` is wrong. It should be `fmpz_mod_discrete_log_pohlig_hellman_init`. That line was added in gh-95 which was apparently to prevent a Windows crash seen at the time. That code doesn't use the modulus though so it should not matter whether the modulus is large or small...

The modulus comes in here:
https://github.com/flintlib/python-flint/blob/242c48e45f52bbeb72ae80687cde5adce681a2cc/src/flint/types/fmpz_mod.pyx#L55-L73
I don't think that any of those functions would have a signature mismatch.

I'm a bit confused about exactly what is happening but the only thing I can immediately see that seems wrong is that call to `fmpz_mod_discrete_log_pohlig_hellman_clear`.

oscarbenjamin:
Depending on the modulus `fmpz_mod_ctx_init` assigns different function pointers into the context struct:
https://github.com/flintlib/flint/blob/277980c37e22978f05f64e35a317f21ba9d7bcb5/src/fmpz_mod/ctx.c#L16-L75
If that is what trips pyodide's type checking then it would make sense that this fails for a particular modulus. So far the crashes seem consistent with saying that it crashes for medium modulus but not small modulus. I'm testing now whether it also crashes for large modulus.

oscarbenjamin:
I'm not sure why but apparently this line causes the main problem:
https://github.com/flintlib/python-flint/blob/242c48e45f52bbeb72ae80687cde5adce681a2cc/src/flint/types/fmpz_mod.pyx#L73
At least if it is commented out then we don't see the crash... (although other tests fail)

src/flint/types/fmpz_mod.pyx
```diff
@@ -32,6 +32,40 @@ cimport libc.stdlib
...
+    b = fmpz(2)
+    c = fmpz(0)
+    fmpz_mod_add(c.val, a.val, b.val, ctx)
+    return c
+
+
+def use_fmpz_is_probabprime():
+    cdef fmpz p
+    p = fmpz(2**127 - 1)
+    return fmpz_is_probabprime(p.val)
```
oscarbenjamin, line 66: Just calling this function is enough to cause the crash:
```
Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
The cause of the fatal error was:
RuntimeError: null function or function signature mismatch
    at wasm://wasm/02e45b4a:wasm-function[1550]:0x1de118
../.venv-pyodide/lib/python3.12/site-packages/flint/test/test_all.py::test_use_fmpz_is_probabprime Stack (most recent call first):
    at wasm://wasm/02e45b4a:wasm-function[1547]:0x1de04f
  File "/home/runner/work/python-flint/python-flint/.venv-pyodide/lib/python3.12/site-packages/flint/test/test_all.py", line 4682 in test_use_fmpz_is_probabprime
  File "/home/runner/work/python-flint/python-flint/.venv-pyodide/lib/python3.12/site-packages/_pytest/python.py", line 159 in pytest_pyfunc_call
    at wasm://wasm/02e45b4a:wasm-function[1531]:0x1dcd97
    at wasm://wasm/02e45b4a:wasm-function[1530]:0x1dcba1
    at wasm://wasm/02e45b4a:wasm-function[1521]:0x1dae4b
    at wasm://wasm/02e45b4a:wasm-function[1522]:0x1db7dd
    at wasm://wasm/02e45b4a:wasm-function[313]:0xa924e
    at wasm://wasm/02e45b4a:wasm-function[220]:0x9dce6
    at wasm://wasm/0268a37a:wasm-function[1160]:0x1c4cf8
    at wasm://wasm/0268a37a:wasm-function[3428]:0x2a2a6d {
  pyodide_fatal_error: true
}
```

oscarbenjamin:
It would be great if we could see what the functions in the stacktrace are but their names are all replaced with numbers in the minified wasm build. What I don't understand is if the mismatch occurs exactly when calling `fmpz_is_probabprime` or if it is during some internal call between FLINT functions.

Note that the crash occurs when the `fmpz` has a larger value. This is important because the FLINT `fmpz` type can be either an inline integer or a pointer to a multiprecision integer. In python-flint the C type is declared as:
https://github.com/flintlib/python-flint/blob/242c48e45f52bbeb72ae80687cde5adce681a2cc/src/flint/flintlib/types/flint.pxd#L39-L40
This is specifying that an `fmpz_t` is an array of one `slong` which means an unsigned long that is 32 bits in a wasm32 bit build. The type of `p.val` is declared as `fmpz_t`:
https://github.com/flintlib/python-flint/blob/242c48e45f52bbeb72ae80687cde5adce681a2cc/src/flint/types/fmpz.pxd#L28-L36
Ultimately the C code generated for this struct says that val is of type `slong[1]`.

When we call `fmpz_is_probabprime(p.val)` for large `p` the bytes at `.val` really represent a pointer to a multiprecision integer rather than an inline slong so somewhere there needs to be a cast from an integer type to a pointer type. I wonder if that is what is causing the signature mismatch here but I don't really understand how wasm's runtime signature checking works. This doesn't crash for small modulus which suggests that the signature is okay when calling `fmpz_is_probabprime` with the integer value so maybe the problematic cast/call is actually within FLINT rather than when python-flint calls into the FLINT shared library?

Note that in the browser at [sympy live](https://live.sympy.org/) you can use  python-flint 0.6.0 along with FLINT 3.1 and this works fine there:
```
>>> import flint
>>> flint.fmpz(2**127 - 1).is_probable_prime()
1
```
Also there the whole test suite passes at least if the doctests are skipped (if they are not skipped then it seems to hang and not run them):
```
from flint.test.__main__ import main
main('-t')
```
I'm at a bit of a loss here. Maybe better debug output would help.

oscarbenjamin:
This should go from FLINT main to rc2: https://github.com/flintlib/flint/issues/2247#issuecomment-2715719885

agriyakhetarpal:
I'm going to try diving deeper into `fmpz_is_probabprime` with a debug Pyodide build; I will post the results in a few hours.

oscarbenjamin:
> I will post the results in a few hours.

I'm guessing the results got lost somewhere :)

In the mean time not much else has changed in FLINT/python-flint land. 

Where we are stuck as I see it as still the unhelpful debug output...

Maybe it is worth attempting the build in the pyodide distribution rather than here to see if bumping the version of either FLINT or python-flint breaks everything there.

agriyakhetarpal:
> > I will post the results in a few hours.
> 
> I'm guessing the results got lost somewhere :)
> 
> In the mean time not much else has changed in FLINT/python-flint land.
> 
> Where we are stuck as I see it as still the unhelpful debug output...

Argh, I'm so sorry about that! Yes, the results are indeed lost by now 😅 Attempting it again.

> Maybe it is worth attempting the build in the pyodide distribution rather than here to see if bumping the version of either FLINT or python-flint breaks everything there.

In Pyodide, we'll migrate to https://github.com/pyodide/pyodide-recipes/ pretty soon, and surely within this release cycle. I had bumped the FLINT and python-flint versions to 3.2.2 and 0.7.1 quite recently in https://github.com/pyodide/pyodide-recipes/pull/91 to accommodate Cython>=3.1 support and fix the failing pipelines...

I don't think it is useful to try out Emscripten 3.1.58 here, as we now have Emscripten 4.0.9 in Pyodide's `main` branch, so I'll bump it here.

agriyakhetarpal:
Let's see how it goes with the Emscripten version bump, hopefully no compile-time failures 🤞🏻

agriyakhetarpal:
Okay, there are no compilation failures, but in https://github.com/agriyakhetarpal/python-flint/actions/runs/15234126937/job/42845748809?pr=1, we have a fatal error with `test_properties_poly_mpoly`. I'll now build `python-flint` with the Pyodide repo with a debug build of Pyodide.

oscarbenjamin:
> Does that help track any problems in some way?

The critical question here is really whether the difference between this PR and pyodide is to do with this PR or to do with other changes in FLINT or python-flint.

If I go to [sympy live](https://live.sympy.org/) then I can run:
```
from flint.test.__main__ import main
main("-vt")
```
That runs the tests but not doctests and passes fine. Without `-vt` the doctests seem to hang. That website is using python-flint 0.6.0 which has FLINT 3.0 bundled in.

Note that the doctests did actually used to be quite slow until some tests were skipped etc. In 0.7.1 the whole test suite including doctests takes a few seconds on a native CPU so it should not in any case hang even in pyodide.

To get from there to here though we have changed a few things:

- FLINT goes from 3.0 to 3.2
- python-flint goes from 0.6.0 to 0.7.1
- This PR builds everything possibly in a different way from the pyodide distribution.

My worry is that it might actually be the first one that is the problem here and that FLINT 3.2 is doing something that memory safe emscripten does not like. I'm not sure how to isolate that from everything else though.

One way is to confirm if `main("-vt")` succeeds in the pyodide distribution with FLINT 3.2 and python-flint 0.7.1. If that is the case then it suggests that the issue here is the last point: this is somehow not building things the same way as for pyodide.

On the other hand if `main("-vt")` fails in pyodide with FLINT 3.2 and python-flint 0.7.1 then the issue is not really to do with this PR and something needs to be changed hopefully only in python-flint but possibly also in FLINT as well.

What I would like to see is the names of the functions in the traceback that I showed above:
```
Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
The cause of the fatal error was:
RuntimeError: null function or function signature mismatch
    at wasm://wasm/02e45b4a:wasm-function[1550]:0x1de118
../.venv-pyodide/lib/python3.12/site-packages/flint/test/test_all.py::test_use_fmpz_is_probabprime Stack (most recent call first):
    at wasm://wasm/02e45b4a:wasm-function[1547]:0x1de04f
  File "/home/runner/work/python-flint/python-flint/.venv-pyodide/lib/python3.12/site-packages/flint/test/test_all.py", line 4682 in test_use_fmpz_is_probabprime
  File "/home/runner/work/python-flint/python-flint/.venv-pyodide/lib/python3.12/site-packages/_pytest/python.py", line 159 in pytest_pyfunc_call
    at wasm://wasm/02e45b4a:wasm-function[1531]:0x1dcd97
    at wasm://wasm/02e45b4a:wasm-function[1530]:0x1dcba1
    at wasm://wasm/02e45b4a:wasm-function[1521]:0x1dae4b
    at wasm://wasm/02e45b4a:wasm-function[1522]:0x1db7dd
    at wasm://wasm/02e45b4a:wasm-function[313]:0xa924e
    at wasm://wasm/02e45b4a:wasm-function[220]:0x9dce6
    at wasm://wasm/0268a37a:wasm-function[1160]:0x1c4cf8
    at wasm://wasm/0268a37a:wasm-function[3428]:0x2a2a6d {
  pyodide_fatal_error: true
}
```
My previous testing (here in CI) indicated that this failure occurs if `fmpz_is_probabprime` is called but only when called with a larger integer. The largeness matters because FLINT's fmpz type uses tagged pointers. The representation is something like:

- an `fmpz` smaller than `2**62` is a left shifted 64 bit integer with low bit set to 1.
- an `fmpz` larger than `2**62` is a pointer to a multiprecision integer struct having low bits equal to zero.

There are therefore functions that convert pointers to integers and back. That seems like precisely the kind of thing that might upset emscripten although it seems to have been okay for FLINT 3.0.

Also though there are code paths with branches for smaller and larger integers. It is possible that the large integer just triggers a different path and then there really is a signature mismatch because the declared types for some function signature are wrong in python-flint compared to FLINT's C code. This is very possible because the declared types here are all parsed from FLINT's docs and there can be bugs where the docs don't match the code.

oscarbenjamin:
> One way is to confirm if `main("-vt")` succeeds in the pyodide distribution with FLINT 3.2 and python-flint 0.7.1

Note also that python-flint 0.7.1 is compatible with FLINT 3.0, 3.1 or 3.2. That means that it is possible to test first changing python-flint from 0.6.0 to 0.7.1 while keeping FLINT at 3.0. Then it is possible to increase the FLINT version as a separate step.

agriyakhetarpal:
Thanks for all the information and for providing a summary!

> One way is to confirm if `main("-vt")` succeeds in the pyodide distribution with FLINT 3.2 and python-flint 0.7.1. If that is the case then it suggests that the issue here is the last point: this is somehow not building things the same way as for pyodide.
> 
> On the other hand if `main("-vt")` fails in pyodide with FLINT 3.2 and python-flint 0.7.1 then the issue is not really to do with this PR and something needs to be changed hopefully only in python-flint but possibly also in FLINT as well.

Tests with `main("-vt")` in the main Pyodide repository are running with FLINT 3.0 and python-flint 0.6.0, where they have been passing (except on Firefox, which is not really a problem). Tests with `main("-vt")` in the pyodide-recipes repository (which just uses a nightly build for Pyodide based on Pyodide's main branch) are not passing and were xfailed. There is hardly any difference in the testing environments there, except for the fact that pyodide-recipes builds them out-of-tree, similar to how we're doing it in this PR, rather than in-tree, and this doesn't make a difference in how they're built. Tests are pyodide-recipes are roughly similar to those in the Pyodide distribution and both can be perceived as one. I am confident that our approach with this PR should not be wrong.

I got `libgmp` to build locally again, so I can have some results from `fmpz_is_probabprime` hopefully soon 🤞🏻

agriyakhetarpal:
Okay, I got everything to build, but what's odd is that

```python
def test_use_fmpz_is_probabprime():
    from flint.types.fmpz_mod import use_fmpz_is_probabprime
    assert use_fmpz_is_probabprime() == 1
```

does not raise the function signature mismatch we've seen in CI, but rather only fails with an import failure:

```
  File "<console>", line 1, in <module>
ImportError: cannot import name 'use_fmpz_is_probabprime' from 'flint.types.fmpz_mod'
 (/lib/python3.13/site-packages/flint/types/fmpz_mod.cpython-313-wasm32-emscripten.so
)
```

and there is no traceback that we can see for further debugging.

Then, I tried with `test_properties_poly_mpoly` for another test and got a fatal error, but since then `_all_polys_mpolys` is raising an `_IncompleteInputError` and so forth I've been unable to rerun this test after setting my breakpoint in the WASM debugger :/

agriyakhetarpal:
Okay, I tried `main("-vt")` instead as I wasn't able to get a mismatch any other way. Here is a minimal reproducer to get a function signature mismatch:

```python
from flint import fmpz_mod_ctx

p_med = 2**127 - 1
F_med = fmpz_mod_ctx(p_med)
```

I was able to get this from `test_fmpz_mod`, where it failed in the browser. However, we can see that `test_fmpz_mod` passes in the Pyodide CI job.

I'll now try to find the code pointer where this happened.

oscarbenjamin:
> we can see that `test_fmpz_mod` passes in the Pyodide CI job.

Where is that? Is that with FLINT 3.0 and python-flint 0.7.1?

src/flint/test/test_all.py
```diff
@@ -2021,8 +2023,8 @@ def test_fmpz_mod_poly():
 
     # Arithmetic
     p_sml = 163
-    p_med = 2**127 - 1
-    p_big = 2**255 - 19
+    p_med = 167
+    p_big = 173
 
     F_sml = fmpz_mod_ctx(p_sml)
     F_med = fmpz_mod_ctx(p_med)
```
oscarbenjamin, line 2030: This PR modifies `test_fmpz_mod` to avoid the large integers.

agriyakhetarpal: Would you like me to try with `p_big = 2**255 - 19`?

oscarbenjamin:
Maybe there is a 32 bit vs 64 bit mismatch somewhere. I forgot that WASM is 32 bit.

oscarbenjamin:
The signature in the C code is:
```
void mpn_mod_ctx_clear(gr_ctx_t ctx)
```
Your signature shows
```
call_indirect (param i32) (result i32)
```
I assume that result refers to the return type and is not void...

oscarbenjamin:
I need to pop out. Will take another look later.

agriyakhetarpal:
I've quite naïvely attempted to patch FLINT in e78c89146eb563ba14d0e5d71f4acea24845d4cd, let's see if it helps with anything :)

oscarbenjamin:
Okay, thanks for debugging this. Just knowing that the problem is connected to `mpn_mod_ctx_clear` already answers a lot of questions.

Firstly [mpn_mod](https://fredrikj.net/blog/2024/04/faster-crypto-sized-residue-ring-arithmetic-in-flint/) is a new FLINT feature and in particular it is new in FLINT 3.2. That suggests that a simple workaround for now would just be to use FLINT 3.1 (although there might be other problems with that).

Secondly `mpn_mod` is for intermediate-sized integers (2-16 words) so we see it with `2**127 - 1` which needs 4 words in 32 bit WASM. The issue is not connected to tagged pointers or anything. We have just used an integer of the right size to hit the `mpn_mod` code path.

Thirdly mpn_mod uses FLINT's generics system which means that there are lots of function pointers that are being casted to different types. I can easily imagine that there might be something happening there that is fine from the perspective of a normal C compiler but that emscripten might have trouble with.

Now the question is then why emscripten is confused about the signature of this function. I think this must either be a bug in emscripten or the WASM runtime or in FLINT.

In `src/gr_types.h` we have:
```c
typedef int (*gr_funcptr)(void);
```
https://github.com/flintlib/flint/blob/6b62c1c3e2b858517e30fa36d7a7b26d1c9e68ad/src/gr_types.h#L47
I think this says that `gr_funcptr` should be a function of that takes no arguments and returns an int.

This is also how `gr_funcptr` is documented:
https://github.com/flintlib/flint/blob/6b62c1c3e2b858517e30fa36d7a7b26d1c9e68ad/doc/source/gr_implementing.rst?plain=1#L70-L73

Here we can see where `mpn_mod_ctx_clear` is assigned to the method table with:
```c
    {GR_METHOD_CTX_CLEAR,       (gr_funcptr) mpn_mod_ctx_clear},
```
https://github.com/flintlib/flint/blob/6b62c1c3e2b858517e30fa36d7a7b26d1c9e68ad/src/mpn_mod/ctx.c#L26

That is casting `mpn_mod_ctx_clear` to `gr_funcptr`. However the signature of `mpn_mod_ctx_clear` does not match this:
```c
void
mpn_mod_ctx_clear(gr_ctx_t ctx)
```
https://github.com/flintlib/flint/blob/6b62c1c3e2b858517e30fa36d7a7b26d1c9e68ad/src/mpn_mod/ring.c#L33-L34

The various function pointers stored in this table are casted back to particular function types when used:
https://github.com/flintlib/flint/blob/6b62c1c3e2b858517e30fa36d7a7b26d1c9e68ad/src/gr.h#L738-L974

In particular `gr_clear` calls `GR_CTX_OP`:
https://github.com/flintlib/flint/blob/6b62c1c3e2b858517e30fa36d7a7b26d1c9e68ad/src/gr.h#L936
Then `GR_CTX_OP` casts to `gr_method_ctx`:
https://github.com/flintlib/flint/blob/6b62c1c3e2b858517e30fa36d7a7b26d1c9e68ad/src/gr.h#L838
And `gr_method_ctx` is defined as returning `int`:
https://github.com/flintlib/flint/blob/6b62c1c3e2b858517e30fa36d7a7b26d1c9e68ad/src/gr.h#L742
However that is not consistent with the signature of `mpn_mod_ctx_clear`.

If I have understood this correctly then I think that the bug is that `GR_CTX_OP` should not cast to `gr_method_ctx` because the clear functions don't return ints. There needs to be a new typedef like:
```c
typedef void ((*gr_clear_method_ctx)(gr_ctx_ptr));
```
Then `GR_CTX_OP` should cast to `gr_clear_method_ctx`.

If that is correct then the fix is something that needs to happen in FLINT although it is possible to patch it here. However I would suggest to first just try using FLINT 3.1. If we can get things working here with FLINT 3.1 then we can pursue the upstream fix separately.

agriyakhetarpal:
> > I don't know why the Pyodide CI hasn't kicked off here (GitHub is rate limiting itself nowadays? :P),
> 
> It is because you are a "first-time contributor". Every time you push I have to approve the CI workflows.

Ah, I'm aware of that setting; but no, @fredrik-johansson had approved the workflows for me on that occasion and it had failed in the setup stage; see https://github.com/flintlib/python-flint/actions/runs/15250259416/job/42885562521?pr=262

If this helps, I've created a small PR at https://github.com/flintlib/python-flint/pull/285 just now so that the workflows here will run without approval from next time. Thanks!

agriyakhetarpal:
This time, the mismatch is in a new test: `test_docstrings.py::test_docstrings[flint.types._gr.new5]` – the previous one, before the patch, was happening in `test_docstrings.py::test_docstrings[flint.types._gr.new11]`. How may I know which one of the `new()` declarations in `src/flint/types/_gr.pyx` it matches to?

If it is the fifth usage, then it should be

https://github.com/flintlib/python-flint/blob/a5a8adccfb9a1155169ca96ba2741c4bd2e0904e/src/flint/types/_gr.pyx#L344-L358

but I could be wrong.

oscarbenjamin:
We have a new cast failure:
```
Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
The cause of the fatal error was:
RuntimeError: null function or function signature mismatch
    at wasm://wasm/02fbbb56:wasm-function[420]:0xc3c08
    at wasm://wasm/02fbbb56:wasm-function[628]:0xed05f
    at wasm://wasm/020eeec2:wasm-function[2306]:0x1b8b1d
    at wasm://wasm/02fbbb56:wasm-function[683]:0xf1c10
    at wasm://wasm/020eeec2:wasm-function[2306]:0x1b8b1d
    at wasm://wasm/020eeec2:wasm-function[1962]:0x1a0d0a
    at wasm://wasm/020eeec2:wasm-function[1949]:0x1a01c4
    at wasm://wasm/020eeec2:wasm-function[2288]:0x1b80ba
    at wasm://wasm/020eeec2:wasm-function[2289]:0x1b829c
    at wasm://wasm/020eeec2:wasm-function[2277]:0x1b723d {
  pyodide_fatal_error: true
}
../.venv-pyodide/lib/python3.13/site-packages/flint/test/test_docstrings.py::test_docstrings[flint.types._gr.new5] Stack (most recent call first):
```
Having manually inserted something into each docstring I find that new5 seems to mean this one:
https://github.com/flintlib/python-flint/blob/40edeb6edb2dd07bcabed9893b719cdf151b3b69/src/flint/types/_gr.pyx#L1398-L1416

oscarbenjamin:
This FLINT patch should do it:
```diff
diff --git a/src/gr/ca.c b/src/gr/ca.c
index 1efb1da36..416456764 100644
--- a/src/gr/ca.c
+++ b/src/gr/ca.c
@@ -1600,12 +1600,11 @@ _gr_ca_mat_det(ca_t res, const ca_mat_t x, gr_ctx_t ctx)
     return GR_SUCCESS;
 }
 
-int
+void
 _gr_ca_ctx_clear(gr_ctx_t ctx)
 {
     ca_ctx_clear(GR_CA_CTX(ctx));
     flint_free(GR_CA_CTX(ctx));
-    return GR_SUCCESS;
 }
 
 truth_t
```

agriyakhetarpal:
Thanks! From the latest run, "memory access out of bounds" errors are usually more tricky to debug, because they can either be flaky, be a sign of an underlying problem in the code, or just that the stack size is too small to accommodate. In the browser's memory, there's roughly 300 MiB for a tab, and the maximum memory that can be acquired is 4 GiB. In practice, it's usually less than that, of course. I would say it should be fine to debug this if possible, but if we can skip the tests that cause, that is fine, too. If you could point me to which docstring this is, I can run it with a debug build and see if it points to something. For example, SciPy also skips a few tests like this because the underlying array is too big to operate on, and it doesn't make sense to do so from the browser's perspective. I don't know if there is a million x million matrix being used in python-flint's code, though. :)

oscarbenjamin:
> I don't know if there is a million x million matrix being used in python-flint's code, though. :)

Not in the test suite. The test suite does not do anything particularly taxing. In python-flint 0.6.0 there was one particularly slow doctest that took 100 times longer than the entire rest of the test suite but that one is skipped now.

fredrik-johansson:
> The latest run has
> 
> ```
> Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
> The cause of the fatal error was:
> RuntimeError: memory access out of bounds
>     at wasm://wasm/020eeec2:wasm-function[9770]:0x4883e2
>     at wasm://wasm/02dcd732:wasm-function[15594]:0x927095
>     at wasm://wasm/02dcd732:wasm-function[15578]:0x926b22
>     at wasm://wasm/02dcd732:wasm-function[14733]:0x8d1562
>     at wasm://wasm/02fbbb8e:wasm-function[420]:0xc3c08
>     at wasm://wasm/020eeec2:wasm-function[2306]:0x1b8b1d
>     at wasm://wasm/02fbbb8e:wasm-function[683]:0xf1c10
>     at wasm://wasm/020eeec2:wasm-function[2306]:0x1b8b1d
>     at wasm://wasm/020eeec2:wasm-function[1962]:0x1a0d0a
>     at wasm://wasm/020eeec2:wasm-function[1949]:0x1a01c4 {
>   pyodide_fatal_error: true
> }
> ../.venv-pyodide/lib/python3.13/site-packages/flint/test/test_docstrings.py::test_docstrings[flint.types._gr.new13] Stack (most recent call first):
> ```
> 
> I think that new13 is:
> 
> https://github.com/flintlib/python-flint/blob/590c7d6ea57530e7dfb4c687e53bfe7e3f2e7eb7/src/flint/types/_gr.pyx#L1243-L1262

That ``gr_nf_fmpz_poly_ctx`` class in python-flint looks wrong. The element type of ``nf`` is ``nf_elem_t``, not ``fmpz_poly_t``.

oscarbenjamin:
I commented out the `gr_nf_fmpz_poly_ctx` doctests and now we have another signature mismatch:
```
Pyodide has suffered a fatal error. Please report this to the Pyodide maintainers.
The cause of the fatal error was:
RuntimeError: null function or function signature mismatch
    at wasm://wasm/02e4ce0a:wasm-function[371]:0xb0b6a
    at wasm://wasm/02e4ce0a:wasm-function[7414]:0x4a17c1
    at wasm://wasm/02e4ce0a:wasm-function[7418]:0x4a1ec1
    at wasm://wasm/02e4ce0a:wasm-function[345]:0xacad0
    at wasm://wasm/02e4ce0a:wasm-function[245]:0xa13c5
    at wasm://wasm/020eeec2:wasm-function[1123]:0x167165
    at wasm://wasm/020eeec2:wasm-function[3522]:0x26be2a
    at wasm://wasm/020eeec2:wasm-function[3520]:0x25d66b
    at wasm://wasm/020eeec2:wasm-function[3485]:0x25aa9f
    at wasm://wasm/020eeec2:wasm-function[3522]:0x260f21 {
  pyodide_fatal_error: true
}
../.venv-pyodide/lib/python3.13/site-packages/flint/test/test_docstrings.py::test_docstrings[flint.types.fmpz_poly.hilbert_class_poly] Stack (most recent call first):
```
That one is `acb_modular_hilbert_class_poly`:
https://github.com/flintlib/python-flint/blob/590c7d6ea57530e7dfb4c687e53bfe7e3f2e7eb7/src/flint/types/fmpz_poly.pyx#L616-L636

oscarbenjamin:
> I assume it is getting into trouble with function pointers in `flint_parallel_binary_splitting` here:

The arguments are supposed to be:
```c
type:: void (* bsplit_basecase_func_t)(void *, slong, slong, void *)
type:: void (* bsplit_merge_func_t)(void *, void *, void *, void *)
type:: void (* bsplit_init_func_t)(void *, void *)
type:: void (* bsplit_clear_func_t)(void *, void *)
```
The last two are being given as `arb_poly_init` and `arb_poly_clear` which only take one argument.

oscarbenjamin:
Since this is finding more issues in FLINT more than in python-flint probably the best thing to do would be to run FLINT's own test suite (`make test`) under emscripten.

@agriyakhetarpal what is a reasonable way to get set up to run FLINT's `make test` on say Ubuntu and can you describe the steps needed to build FLINT in tree and run its tests?

oscarbenjamin:
Okay the test suite completed without crashing. Most of the test failures are because of disabling the prime check in `fmpz_mod` which also causes the tests to fail in a non-emscripten build so I've added that back.

Now I'll undo most of the changes to tests etc here to get back to the minimum changes needed.

agriyakhetarpal:
> > Building/testing FLINT in tree would be the same as we're doing it in the `pyodide_build_flint.sh` script, but wth the additional `make tests` and `make check` after that.
> 
> Does `emmake check` already work? If so then that is all that is needed.

It has all compiled well, and I'm running `emmake make tests && emmake make check` right now – will post the results once I have them.

oscarbenjamin:
> This is what I got from `emmake make tests`, it didn't succeed:

Maybe this can fix the immediate blocker:
```diff
diff --git a/src/mag/test/t-d_log_lower_bound.c b/src/mag/test/t-d_log_lower_bound.c
index 27c8bd4d3..94a8f2175 100644
--- a/src/mag/test/t-d_log_lower_bound.c
+++ b/src/mag/test/t-d_log_lower_bound.c
@@ -16,8 +16,8 @@
 
 /* Defined in t-d_log_lower_bound.c, t-d_log_upper_bound.c, t-set_d_2exp_fmpz.c
  * and t-set_d.c */
-#ifndef d_randtest2
-#define d_randtest2 d_randtest2
+#ifndef _d_randtest2
+#define _d_randtest2
 /* XXX: d_randtest is not good enough */
 
 #define EXP_MINUS_32 2.3283064365386962891e-10
diff --git a/src/mag/test/t-d_log_upper_bound.c b/src/mag/test/t-d_log_upper_bound.c
index bf01748f5..4f5126742 100644
--- a/src/mag/test/t-d_log_upper_bound.c
+++ b/src/mag/test/t-d_log_upper_bound.c
@@ -16,8 +16,8 @@
 
 /* Defined in t-d_log_lower_bound.c, t-d_log_upper_bound.c, t-set_d_2exp_fmpz.c
  * and t-set_d.c */
-#ifndef d_randtest2
-#define d_randtest2 d_randtest2
+#ifndef _d_randtest2
+#define _d_randtest2
 /* XXX: d_randtest is not good enough */
 
 #define EXP_MINUS_32 2.3283064365386962891e-10
diff --git a/src/mag/test/t-set_d.c b/src/mag/test/t-set_d.c
index f112ccc27..0a560a0af 100644
--- a/src/mag/test/t-set_d.c
+++ b/src/mag/test/t-set_d.c
@@ -16,8 +16,8 @@
 
 /* Defined in t-d_log_lower_bound.c, t-d_log_upper_bound.c, t-set_d_2exp_fmpz.c
  * and t-set_d.c */
-#ifndef d_randtest2
-#define d_randtest2 d_randtest2
+#ifndef _d_randtest2
+#define _d_randtest2
 /* XXX: d_randtest is not good enough */
 
 #define EXP_MINUS_32 2.3283064365386962891e-10
diff --git a/src/mag/test/t-set_d_2exp_fmpz.c b/src/mag/test/t-set_d_2exp_fmpz.c
index ff402e8e1..14622f477 100644
--- a/src/mag/test/t-set_d_2exp_fmpz.c
+++ b/src/mag/test/t-set_d_2exp_fmpz.c
@@ -16,8 +16,8 @@
 
 /* Defined in t-d_log_lower_bound.c, t-d_log_upper_bound.c, t-set_d_2exp_fmpz.c
  * and t-set_d.c */
-#ifndef d_randtest2
-#define d_randtest2 d_randtest2
+#ifndef _d_randtest2
+#define _d_randtest2
 /* XXX: d_randtest is not good enough */
 
 #define EXP_MINUS_32 2.3283064365386962891e-10
```

oscarbenjamin:
> For the remaining test failures then we have this one in `test_fq_default_poly`:
> 
> https://github.com/flintlib/python-flint/blob/590c7d6ea57530e7dfb4c687e53bfe7e3f2e7eb7/src/flint/test/test_all.py#L4731
> 
> 
> That implies that this function is not raising `ValueError` which is worrying:

After 60c291eb0d61e3af55f4de060264c37fcf1910bb I can reproduce the `test_fq_default` failure locally (in a non-WASM build) as now shown in CI but only with when running `python -m flint.test` and not via `pytest` or `spin test`.

I think that this is something else, not directly related to emscripten/WASM and it is probably to do with caching and comparison of context objects...

agriyakhetarpal:
Yes, I think not compiling the files is fine for now ;)

Some more mismatches popped up, if this helps:

```bash
make: make tests
  CC  mag/test/main.c
/Users/agriyakhetarpal/Desktop/flint/src/mag/test/main.c:78:5: error: use of undeclared identifier 'test_mag_d_log_lower_bound'; did you mean 'mag_d_log_lower_bound'?
   78 |     TEST_FUNCTION(mag_d_log_lower_bound),
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/test_helpers.h:59:32: note: expanded from macro 'TEST_FUNCTION'
   59 | #define TEST_FUNCTION(label) { CAT(test, label), #label }
      |                                ^~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/templates.h:15:18: note: expanded from macro 'CAT'
   15 | #define CAT(X,Y) X##_##Y
      |                  ^~~~~~~
<scratch space>:168:1: note: expanded from here
  168 | test_mag_d_log_lower_bound
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/mag.h:538:8: note: 'mag_d_log_lower_bound' declared here
  538 | double mag_d_log_lower_bound(double x);
      |        ^
/Users/agriyakhetarpal/Desktop/flint/src/mag/test/main.c:79:5: error: use of undeclared identifier 'test_mag_d_log_upper_bound'; did you mean 'mag_d_log_upper_bound'?
   79 |     TEST_FUNCTION(mag_d_log_upper_bound),
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/test_helpers.h:59:32: note: expanded from macro 'TEST_FUNCTION'
   59 | #define TEST_FUNCTION(label) { CAT(test, label), #label }
      |                                ^~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/templates.h:15:18: note: expanded from macro 'CAT'
   15 | #define CAT(X,Y) X##_##Y
      |                  ^~~~~~~
<scratch space>:171:1: note: expanded from here
  171 | test_mag_d_log_upper_bound
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/mag.h:537:8: note: 'mag_d_log_upper_bound' declared here
  537 | double mag_d_log_upper_bound(double x);
      |        ^
/Users/agriyakhetarpal/Desktop/flint/src/mag/test/main.c:108:5: error: use of undeclared identifier 'test_mag_set_d_2exp_fmpz'; did you mean 'test_mag_add_2exp_fmpz'?
  108 |     TEST_FUNCTION(mag_set_d_2exp_fmpz),
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/test_helpers.h:59:32: note: expanded from macro 'TEST_FUNCTION'
   59 | #define TEST_FUNCTION(label) { CAT(test, label), #label }
      |                                ^~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/templates.h:15:18: note: expanded from macro 'CAT'
   15 | #define CAT(X,Y) X##_##Y
      |                  ^~~~~~~
<scratch space>:258:1: note: expanded from here
  258 | test_mag_set_d_2exp_fmpz
      | ^~~~~~~~~~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/mag/test/t-add_2exp_fmpz.c:16:1: note: 'test_mag_add_2exp_fmpz' declared here
   16 | TEST_FUNCTION_START(mag_add_2exp_fmpz, state)
      | ^
/Users/agriyakhetarpal/Desktop/flint/src/test_helpers.h:62:5: note: expanded from macro 'TEST_FUNCTION_START'
   62 | int CAT(test, label)(void)                              \
      |     ^
/Users/agriyakhetarpal/Desktop/flint/src/templates.h:15:18: note: expanded from macro 'CAT'
   15 | #define CAT(X,Y) X##_##Y
      |                  ^
<scratch space>:272:1: note: expanded from here
  272 | test_mag_add_2exp_fmpz
      | ^
/Users/agriyakhetarpal/Desktop/flint/src/mag/test/main.c:109:5: error: use of undeclared identifier 'test_mag_set_d'; did you mean 'test_mag_get_d'?
  109 |     TEST_FUNCTION(mag_set_d),
      |     ^~~~~~~~~~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/test_helpers.h:59:32: note: expanded from macro 'TEST_FUNCTION'
   59 | #define TEST_FUNCTION(label) { CAT(test, label), #label }
      |                                ^~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/templates.h:15:18: note: expanded from macro 'CAT'
   15 | #define CAT(X,Y) X##_##Y
      |                  ^~~~~~~
<scratch space>:261:1: note: expanded from here
  261 | test_mag_set_d
      | ^~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/mag/test/t-get_d.c:17:1: note: 'test_mag_get_d' declared here
   17 | TEST_FUNCTION_START(mag_get_d, state)
      | ^
/Users/agriyakhetarpal/Desktop/flint/src/test_helpers.h:62:5: note: expanded from macro 'TEST_FUNCTION_START'
   62 | int CAT(test, label)(void)                              \
      |     ^
/Users/agriyakhetarpal/Desktop/flint/src/templates.h:15:18: note: expanded from macro 'CAT'
   15 | #define CAT(X,Y) X##_##Y
      |                  ^
<scratch space>:17:1: note: expanded from here
   17 | test_mag_get_d
      | ^
/Users/agriyakhetarpal/Desktop/flint/src/mag/test/main.c:78:5: error: incompatible function pointer types initializing 'int (*)(void)' with an expression of type 'double (double)' [-Wincompatible-function-pointer-types]
   78 |     TEST_FUNCTION(mag_d_log_lower_bound),
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/test_helpers.h:59:32: note: expanded from macro 'TEST_FUNCTION'
   59 | #define TEST_FUNCTION(label) { CAT(test, label), #label }
      |                                ^~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/templates.h:15:18: note: expanded from macro 'CAT'
   15 | #define CAT(X,Y) X##_##Y
      |                  ^~~~~~~
<scratch space>:168:1: note: expanded from here
  168 | test_mag_d_log_lower_bound
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/mag/test/main.c:79:5: error: incompatible function pointer types initializing 'int (*)(void)' with an expression of type 'double (double)' [-Wincompatible-function-pointer-types]
   79 |     TEST_FUNCTION(mag_d_log_upper_bound),
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/test_helpers.h:59:32: note: expanded from macro 'TEST_FUNCTION'
   59 | #define TEST_FUNCTION(label) { CAT(test, label), #label }
      |                                ^~~~~~~~~~~~~~~~
/Users/agriyakhetarpal/Desktop/flint/src/templates.h:15:18: note: expanded from macro 'CAT'
   15 | #define CAT(X,Y) X##_##Y
      |                  ^~~~~~~
<scratch space>:171:1: note: expanded from here
  171 | test_mag_d_log_upper_bound
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~
6 errors generated.
emcc: error: '/Users/agriyakhetarpal/emsdk/upstream/bin/clang -target wasm32-unknown-emscripten -fignore-exceptions -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr --sysroot=/Users/agriyakhetarpal/emsdk/upstream/emscripten/cache/sysroot -DEMSCRIPTEN -Xclang -iwithsysroot/include/fakesdl -Xclang -iwithsysroot/include/compat -Wall -Werror=newline-eof -Werror=implicit-function-declaration -std=c11 -pedantic -O3 -g3 -I/Users/agriyakhetarpal/Desktop/flint/src -Isrc -I/Users/agriyakhetarpal/Desktop/flint/wasm_libs/include -I/Users/agriyakhetarpal/Desktop/flint/wasm_libs/include -DBUILDING_FLINT -MMD -MP -MF build/mag/test/main.d -c /Users/agriyakhetarpal/Desktop/flint/src/mag/test/main.c -o /var/folders/b3/2bq1m1_50bs4c7305j8vxcqr0000gn/T/emscripten_temp_2bwp0ahc/main_0.o' failed (returned 1)
make: *** [build/mag/test/main] Error 1
emmake: error: 'make tests' failed (returned 2)
```

Otherwise, I shall build/test from a clean state again.

agriyakhetarpal:
Ah, yes, this time it compiles. Here is the output from `emmake make check`:

```bash
emmake make check   
make: make check
add_ssaaaa...
add_ssaaaa                                        0.00   (PASS)
add_sssaaaaaa...
add_sssaaaaaa                                     0.09   (PASS)
flint_clz...
flint_clz                                         0.00   (PASS)
flint_ctz...
flint_ctz                                         0.00   (PASS)
flint_fprintf...
flint_fprintf                                     0.00   (FAIL)
Could not open temporary file "tmp"
Aborted()
/Users/agriyakhetarpal/Desktop/flint/build/test/main:380
  var e = new WebAssembly.RuntimeError(what);
          ^

RuntimeError: Aborted(). Build with -sASSERTIONS for more info.
    at abort (/Users/agriyakhetarpal/Desktop/flint/build/test/main:380:11)
    at __abort_js (/Users/agriyakhetarpal/Desktop/flint/build/test/main:3388:7)
    at main.wasm.abort (wasm://wasm/main.wasm-083fa0d6:wasm-function[8641]:0x4647b4)
    at main.wasm.flint_abort (wasm://wasm/main.wasm-083fa0d6:wasm-function[40]:0x90a0)
    at main.wasm.main (wasm://wasm/main.wasm-083fa0d6:wasm-function[38]:0x9010)
    at Module._main (/Users/agriyakhetarpal/Desktop/flint/build/test/main:3728:102)
    at callMain (/Users/agriyakhetarpal/Desktop/flint/build/test/main:3754:15)
    at doRun (/Users/agriyakhetarpal/Desktop/flint/build/test/main:3793:24)
    at run (/Users/agriyakhetarpal/Desktop/flint/build/test/main:3806:5)
    at removeRunDependency (/Users/agriyakhetarpal/Desktop/flint/build/test/main:348:7)

Node.js v20.18.0
make: *** [build/test/main_TEST_RUN] Error 1
emmake: error: 'make check' failed (returned 2)
```

The failing test is fine; we can't access the file system without mounting a `NodeFS` or a similar one. I don't think we should dwell on it.

Note that I had to manually modify the generated `Makefile` as it did not interpret that `build/test/main` was a JavaScript file, so I had to prepend `node` before the `%_TEST_RUN: %` rules. It should build `main.js` if the `HOST_OS` matches `emscripten` and use Node.js in that case by default – but that was just some misconfiguration that we can deal with. At least from this, we can see that most tests pass and WASM compilation works.

agriyakhetarpal:
I added [`NODERAWFS=1`](https://emscripten.org/docs/api_reference/Filesystem-API.html#noderawfs) to my `TESTCFLAGS` for allowing access to my file system, and this has let the tests go forward much more:

```bash
make: make check
add_ssaaaa...
add_ssaaaa                                        0.00   (PASS)
add_sssaaaaaa...
add_sssaaaaaa                                     0.04   (PASS)
flint_clz...
flint_clz                                         0.00   (PASS)
flint_ctz...
flint_ctz                                         0.00   (PASS)
flint_fprintf...
flint_fprintf                                     0.01   (PASS)
flint_printf...
flint_printf                                          (SKIPPED)
memory_manager...
memory_manager                                    0.00   (PASS)
sdiv_qrnnd...
sdiv_qrnnd                                        0.01   (PASS)
smul_ppmm...
smul_ppmm                                         0.01   (PASS)
flint_sort...
flint_sort                                        0.00   (PASS)
sub_dddmmmsss...
sub_dddmmmsss                                     0.05   (PASS)
sub_ddmmss...
sub_ddmmss                                        0.00   (PASS)
udiv_qrnnd...
udiv_qrnnd                                        0.01   (PASS)
udiv_qrnnd_preinv...
udiv_qrnnd_preinv                                 0.00   (PASS)
umul_ppmm...
umul_ppmm                                         0.01   (PASS)
thread_pool...
thread_pool                                       0.10   (PASS)
thread_support_parallel_binary_splitting...
thread_support_parallel_binary_splitting          0.00   (PASS)
thread_support_parallel_do...
thread_support_parallel_do                        0.00   (PASS)
n_addmod...
n_addmod                                          0.00   (PASS)
n_cbrt_binary_search...
n_cbrt_binary_search                              0.00   (PASS)
n_cbrt...
n_cbrt                                            0.01   (PASS)
n_cbrt_chebyshev_approx...
n_cbrt_chebyshev_approx                           0.00   (PASS)
n_cbrtrem...
n_cbrtrem                                         0.01   (PASS)
n_clog_2exp...
n_clog_2exp                                       0.00   (PASS)
n_clog...
n_clog                                            0.00   (PASS)
compute_primes...
Aborted(OOM)
/Users/agriyakhetarpal/Desktop/flint/build/ulong_extras/test/main:380
  var e = new WebAssembly.RuntimeError(what);
          ^

RuntimeError: Aborted(OOM). Build with -sASSERTIONS for more info.
    at abort (/Users/agriyakhetarpal/Desktop/flint/build/ulong_extras/test/main:380:11)
    at abortOnCannotGrowMemory (/Users/agriyakhetarpal/Desktop/flint/build/ulong_extras/test/main:3949:7)
    at _emscripten_resize_heap (/Users/agriyakhetarpal/Desktop/flint/build/ulong_extras/test/main:3955:7)
    at main.wasm.sbrk (wasm://wasm/main.wasm-084fa6f2:wasm-function[8876]:0x4812bd)
    at main.wasm.emscripten_builtin_malloc (wasm://wasm/main.wasm-084fa6f2:wasm-function[8868]:0x47f4c6)
    at main.wasm._flint_malloc (wasm://wasm/main.wasm-084fa6f2:wasm-function[155]:0x186f6)
    at main.wasm.flint_malloc (wasm://wasm/main.wasm-084fa6f2:wasm-function[147]:0x1850f)
    at main.wasm.n_compute_primes (wasm://wasm/main.wasm-084fa6f2:wasm-function[189]:0x1a0b1)
    at main.wasm.n_primes_arr_readonly (wasm://wasm/main.wasm-084fa6f2:wasm-function[271]:0x1fe54)
    at main.wasm.test_compute_primes (wasm://wasm/main.wasm-084fa6f2:wasm-function[27]:0x50f5)

Node.js v20.18.0
make: *** [build/ulong_extras/test/main_TEST_RUN] Error 1
emmake: error: 'make check' failed (returned 2)
```

The relevant section of the `main` file that Emscripten has set up looks like this (lines 3944–3956), but I don't think this helps much:

```
  var getHeapMax = () =>
      HEAPU8.length;
  var _emscripten_get_heap_max = () => getHeapMax();

  var abortOnCannotGrowMemory = (requestedSize) => {
      abort('OOM');
    };
  var _emscripten_resize_heap = (requestedSize) => {
      var oldSize = HEAPU8.length;
      // With CAN_ADDRESS_2GB or MEMORY64, pointers are already unsigned.
      requestedSize >>>= 0;
      abortOnCannotGrowMemory(requestedSize);
    };
```

And from the CI jobs for FLINT, I see that none of the `nmod_vec_`, `flint_mpn`, `fmpz`, `fmpq`, `arb`, `gr_` tests ran; this isn't the full test suite either, so it would be nice if there were a `pytest`-like way to skip tests that we know are going to abort the program without having to recompile everything, or to allow a selective portion of the tests of our choice to run similar to the `-k` option.

oscarbenjamin:
> it would be nice if there were a `pytest`-like way to skip tests that we know are going to abort the program without having to recompile everything, or to allow a selective portion of the tests of our choice to run, similar to the `-k` option.

I'm not sure what options there are but you can just run the individual test files like:
```
$ build/acb/test/main
acb_acos...
acb_acos                                          0.00   (PASS)
acb_acosh...
acb_acosh                                         0.00   (PASS)
acb_agm1...
acb_agm1                                          0.15   (PASS)
acb_agm...
acb_agm                                           0.09   (PASS)
...
```
You can list the hundred or so test executables and maybe put them in a script and comment out the ones you don't want:
```console
$ ls build/*/test/main
build/acb_calc/test/main               build/fq_default_mat/test/main
build/acb_dft/test/main                build/fq_default_poly_factor/test/main
build/acb_dirichlet/test/main          build/fq_default_poly/test/main
build/acb_elliptic/test/main           build/fq_default/test/main
...
```
Given the things that currently don't work in this PR I would be most interested in `acb_modular`, `nf`, `nf_elem`, and the `gr_*` tests.

oscarbenjamin:
Okay, so the CI job passes now. I think that a good milestone here would be to get something merged so that we have emscripten running in CI and nightly wheels being uploaded even if there are some outstanding known issues.

I suggest we:

- Preserve this PR with its many commits and long discussion intact for posterity.
- Open a new PR that squashes the commits and does any remaining clean ups.
- Then I will follow up after with something to clean up the test changes and come up with a better solution for the doctests that are skipped or sorting of polynomial factors.

Lastly what remains is to see if we can identify and fix the problems with `acb_modular_hilbert_class_poly` and `gr_nf` but unless we can quickly identify the issue there I think it is better to leave that for later.

oscarbenjamin:
I think for now we can track FLINT's main branch for the WASM build so I think the FLINT release doesn't matter. At least the current state in this PR worked good enough with FLINT main and so if FLINT 3.3 is released now then it should be possible to get at least as good as we have here by pinning to FLINT 3.3.

How long it would take to iron these things out is unknown so it is probably best not to wait on it.

oscarbenjamin:
I think we can close this one now. Much progress was made elsewhere. There are still some FLINT/emscripten hickups but I don't think this PR is useful for those any more.

PR #264, opened 2025-03-03
Bump FLINT to 3.2.0-rc1

oscarbenjamin:


oscarbenjamin:
The ARM Linux job is failing but that is because of a FLINT bug: https://github.com/flintlib/flint/issues/2243.

I'll merge this now and then CI should get fixed when bumping the FLINT version from 3.2.0-rc1.

PR #271, opened 2025-03-16
fixing some minor details in test_all.py

fchapoton:
notably fixing a wrong restoration of attribute


oscarbenjamin:
Looks good. Thanks!

PR #274, opened 2025-03-20
Add resultant methods to fmpz_poly and fmpq_poly

rdbliss:
This adds resultant methods to fmpz_poly (resultant and resultant_modular) and fmpq_poly (resultant), which were previously only exposed for fmpz_mod_poly, fmpq_mpoly, fmpz_mpoly, nmod_mpoly, and fmpz_mod_mpoly.

I included resultant_modular because it's in flint, but in a quick performance test I didn't see much difference between it and fmpz_poly.resultant, or even fmpq_poly.resultant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,18 @@ CHANGELOG
 Next release (0.8.0)...
 -----------------------
 
+Contributors
+
+- Robert Dougherty-Bliss (RDB)
+
+Changes
+
+- [gh-274](https://github.com/flintlib/python-flint/pull/274),
+  Add resultant methods to `fmpz_poly`, `fmpq_poly` and
+  `nmod_poly`. Now all univariate and polynomial types have the
+  resultant method except for `fq_default_poly`. (RDB)
+
+
 0.7.0
 -----
 
--- a/src/flint/test/test_all.py
+++ b/src/flint/test/test_all.py
@@ -2834,6 +2834,36 @@ def test_polys():
         if type(p) == flint.fq_default_poly:
             assert raises(lambda: p.integral(), NotImplementedError)
 
+        # resultant checks.
+
+        if is_field and characteristic == 0:
+            # Check that the resultant of two cyclotomic polynomials is right.
+            # See Dresden's 2012 "Resultants of Cyclotomic Polynomials"
+            for m in range(1, 50):
+                for n in range(m + 1, 50):
+                    a = flint.fmpz_poly.cyclotomic(m)
+                    b = flint.fmpz_poly.cyclotomic(n)
+                    q, r = divmod(flint.fmpz(n), flint.fmpz(m))
+                    fs = q.factor()
+                    if r != 0 or len(fs) > 1:
+                        assert a.resultant(b) == 1
+                    else:
+                        prime = fs[0][0]
+                        tot = flint.fmpz(m).euler_phi()
+                        assert a.resultant(b) == prime**tot
+
+        x = P([0, 1])
+        # Flint does not implement resultants over GF(q) for nonprime q, so we
+        # there's nothing for us to check.
+        if composite_characteristic or type(x) == flint.fq_default_poly:
+            pass
+        else:
+            assert x.resultant(x) == 0
+            assert x.resultant(x**2 + x - x) == 0
+
+            for k in range(-10, 10):
+                assert x.resultant(x + S(k)) == S(k)
+                assert x.resultant(x**10 - x**5 + 1) == S(1)
 
 def _all_mpolys():
     return [
@@ -2868,7 +2898,6 @@ def _all_mpolys():
             flint.fmpz(100),
         ),
     ]
-
 
 def test_mpolys():
     for P, get_context, S, is_field, characteristic in _all_mpolys():
--- a/src/flint/test/test_all.py
+++ b/src/flint/test/test_all.py
@@ -2853,17 +2853,25 @@ def test_polys():
                         assert a.resultant(b) == prime**tot
 
         x = P([0, 1])
-        # Flint does not implement resultants over GF(q) for nonprime q, so we
-        # there's nothing for us to check.
-        if composite_characteristic or type(x) == flint.fq_default_poly:
+
+        if composite_characteristic and type(x) == flint.fmpz_mod_poly:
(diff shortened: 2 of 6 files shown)


oscarbenjamin:
Thanks for this.

Can you add a test sort of like this one:
https://github.com/flintlib/python-flint/blob/b87c681b8e4b8f34c286b4a565afcd1acb49bd40/src/flint/test/test_all.py#L3500-L3501
When possible it is best to have generic tests so that we can be sure that all types implement all methods consistently. The tests can make exceptions for particular types though e.g. it looks like there are no resultant functions for fq polynomials.

I suppose that the univariate polynomials have a different signature for resultant because they return a scalar.

rdbliss:
@oscarbenjamin tests added!

By the way, I discovered that flint's fmpz_mod_poly does not like it when the modulus is composite. I think a comment mentioning this exists in the tests already. I skipped that case since it seems like an upstream problem. However, it is very easy to cause a crash with this behavior:

```python
>>> from flint import fmpz_mod_poly, fmpz_mod_poly_ctx
>>> x = fmpz_mod_poly([0, 1], fmpz_mod_poly_ctx(164))
>>> x.resultant(x - 2)
    Exception in fmpz_mod_inv: Cannot invert.
Aborted (core dumped)
```

Would it be worth raising exceptions when using fmpz_mod_poly with composite modulus?

oscarbenjamin:
> Would it be worth raising exceptions when using fmpz_mod_poly with composite modulus?

Yes. There are lots of cases like this where FLINT aborts but in python-flint we want to raise an exception instead. For `fmpz_mod_poly` the context stores whether or not the modulus is prime. For `nmod_poly` there is no context so we don't have a good solution yet.

rdbliss:
@oscarbenjamin thanks for the advice! I added an exception for composite moduli in fmpz_mod_poly, and separated that case out in testing. It seems that nmod_poly does not have this problem, so I didn't do anything special for it.

src/flint/test/test_all.py
```diff
@@ -2834,6 +2834,44 @@ def setbad(obj, i, val):
...
 
+        # resultant checks.
+
+        if is_field and characteristic == 0:
+            # Check that the resultant of two cyclotomic polynomials is right.
+            # See Dresden's 2012 "Resultants of Cyclotomic Polynomials"
+            for m in range(1, 50):
+                for n in range(m + 1, 50):
+                    a = flint.fmpz_poly.cyclotomic(m)
+                    b = flint.fmpz_poly.cyclotomic(n)
```
oscarbenjamin, line 2845: These tests are running in a loop over all polynomial types but don't depend on the type.

I think it would be better just to move all of this out to a separate `test_poly_resultants` test function but still use `_all_polys` for the part that loops over the types.

oscarbenjamin: In general it would be better not to have such large test functions as this one.

src/flint/test/test_all.py
```diff
@@ -2834,6 +2834,44 @@ def setbad(obj, i, val):
...
+            if type(x) == flint.fmpz_mod_poly:
+                assert raises(lambda: x.resultant(x + divisor), ValueError)
+        elif type(x) == flint.fq_default_poly:
+            # Flint does not implement resultants over GF(q) for nonprime q, so
+            # there's nothing for us to check.
+            pass
+        else:
+            assert x.resultant(x) == 0
+            assert x.resultant(x**2 + x - x) == 0
+            assert x.resultant(x**10 - x**5 + 1) == S(1)
```
oscarbenjamin, line 2870: I'm surprised that nmod_poly doesn't crash here. I guess it sometimes does and sometimes doesn't:
```python
In [3]: a = flint.nmod_poly([1, 2, 3], 12)

In [4]: b = flint.nmod_poly([1, 2, 3], 12)

In [5]: a.resultant(b)
Out[5]: 0

In [6]: b = flint.nmod_poly([1, 2, 3, 4], 12)

In [7]: a.resultant(b)
    Cannot invert modulo 3*4
Aborted (core dumped)
```

src/flint/test/test_all.py
oscarbenjamin: I mean that test_all_polys is too long.

oscarbenjamin: Ideally the test suite would have reasonably short test functions like `test_resultants` or something but where each test loops over all possible polynomial types. When a new polynomial type is added it should be possible to just add it to to the list in `_all_polys` and then update each test or add the methods that are needed for consistency with other types.

src/flint/test/test_all.py
rdbliss: Good to know! I added a similar check for nmod_poly that makes this example throw an exception.

src/flint/types/fmpz_poly.pyx
```diff
@@ -397,6 +397,59 @@ cdef class fmpz_poly(flint_poly):
...
+            raise TypeError("cannot convert input to fmpz_poly")
+
+        res = fmpz.__new__(fmpz)
+        fmpz_poly_resultant(res.val, self.val, (<fmpz_poly>other).val)
+        return res
+
+    def resultant_modular(self, other):
+        """
+        Returns the resultant of *self* and *other* using Collins' 1971 modular
+        algorithm.
```
oscarbenjamin, line 429: I think it is better not to include this. There are lots of situations where FLINT provides variant functions for an operation like this and we don't generally wrap all of them in python-flint. It looks like the resultant method already switches between either using `fmpz_poly_resultant_modular` or `fmpz_poly_resultant_euclidean` depending on the degree and bitsize.

README.md
```diff
@@ -162,6 +162,18 @@ CHANGELOG
...
+Contributors
+
+- Robert Dougherty-Bliss (RDB)
+
+Changes
+
+- [gh-274](https://github.com/flintlib/python-flint/pull/274),
+  Add resultant methods to `fmpz_poly`, `fmpq_poly` and
+  `nmod_poly`. Now all univariate and polynomial types have the
+  resultant method except for `fq_default_poly`. (RDB)
```
oscarbenjamin, line 174: Let me know if you want to record a different name here.

oscarbenjamin:
Thanks for this.

PR #282, opened 2025-05-24
Add bindings for type qfb_t

remyoudompheng:
This is a proposal to add bindings for type `qfb_t` as a new class `qfb`.

This implementation defines operators `*` and `**` allowing computation in class groups (similar to the interface of `cypari2` type `Qfb` and SageMath `BinaryQF`) and a few convenience constructors such as `prime_form`.

Let me know if a specific convention should be followed for test cases and documentation.

### DIFF
--- a/bin/all_rst_to_pxd.sh
+++ b/bin/all_rst_to_pxd.sh
@@ -72,7 +72,7 @@ modules=(
     # "dlog"
     # "bool_mat"
     # "perm"
-    # "qfb"
+    "qfb"
     # "nf"
     # "nf_elem"
     # "fmpzi"
--- a/src/flint/__init__.py
+++ b/src/flint/__init__.py
@@ -26,6 +26,8 @@ from .types.fmpz_mod_mat import fmpz_mod_mat
 
 from .types.fmpq_mpoly import fmpq_mpoly_ctx, fmpq_mpoly, fmpq_mpoly_vec
 
+from .types.qfb import *
+
 from .types.fq_default import *
 from .types.fq_default_poly import *
 
--- a/src/flint/flintlib/functions/qfb.pxd
+++ b/src/flint/flintlib/functions/qfb.pxd
@@ -1,0 +1,35 @@
+from flint.flintlib.types.flint cimport fmpz_t, slong, ulong
+from flint.flintlib.types.qfb cimport qfb_t
+
+# unknown type qfb
+# unknown type qfb_hash_t
+
+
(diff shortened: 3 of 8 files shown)


oscarbenjamin:
Thanks for this. I am sure that it is good to add this but I will have to take some time to learn what qfb is...

Are there any other group types in FLINT that could in future be exposed in python-flint?

I'm wondering if it makes sense to have any kind of abstract superclass like the ones in `src/flint/flint_base/flint_base.pxd`. At least it probably makes sense for this to subclass `flint_elem`.

Would it make sense at all for these to have a context (parent) or is that not really needed?

Most other types are going to use the generic interface in future and will end up having contexts in future sort of like:
```
F = Zmod(5)
print(F(2) / F(3)
```
Does the generics part of flint include anything related to qfb?

Apart from the linting job the CI failures are unrelated. I think that they are all caused by a new Cython release. I will fix that.

fredrik-johansson:
> Are there any other group types in FLINT that could in future be exposed in python-flint?

Other than Dirichlet characters (which are already wrapped): permutation groups and the modular group, which both have generics wrappers. There's no qfb generics wrapper yet.

I must admit that I'm not familiar enough with binary quadratic forms to say exactly what the interface here should be, but my rough understanding is that you might want two different types:

1) A type representing any binary quadratic form, i.e. a qfb_t without further restrictions. (These don't form a nice algebraic structure, as far as I know.)

2) A type representing an element of a class group with fixed discriminant D by means of a qfb_t. For this, it would make more sense to store D in a context object defining the group rather than adjoining it to the element. You would also presumably want to enforce reducedness (or maybe have an option flag in the context object for whether to allow unreduced representations).

I guess 2) is more convenient for mathematical applications (and would be logical to do via generics), unless you have a specific use case for unreduced forms requiring 1)?

oscarbenjamin:
I'm going to reopen to rerun the tests after fixing the cython issue in gh-284.

oscarbenjamin:
There are two CI failures. One is the lint issue and the other is that the setuptools build fails. The qfb module needs to be added somewhere in here:
https://github.com/flintlib/python-flint/blob/2ac9797edd58af6dfe68b502a545ec261d04bb72/setup.py#L77-L81
The meson coverage job shows that some lines are not covered by the tests:
```
Name                 Stmts   Miss  Cover   Missing
--------------------------------------------------
flint/types/qfb.pyx     80     13    84%   35, 39, 43, 48, 53, 67, 70, 92-97
```

PR #290, opened 2025-06-23
Bump from windows-2019 to windows-2022 in CI

oscarbenjamin:
The windows-2019 image is no longer supported by GitHub Actions.

In gh-289 I found that the Windows build failed because of a new version of the setup-msys2 action. Looking into that I discovered that the GitHub actions windows-2019 runners will not be available by the end of the month (in a week!). Bumping to windows-2025 didn't work so we'll try windows-2022.

If this works then the msys2 issue still remains but we have another few years before needing to bump the Windows version.

### DIFF
--- a/.github/workflows/buildwheel.yml
+++ b/.github/workflows/buildwheel.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
Closing after gh-294

PR #291, opened 2025-06-24
CI: test all available OS in CI

oscarbenjamin:
Following up from gh-290.

It seems that the wheels built in CI on windows-2019 runners fail on windows-2022 runners. This PR just makes sure that the built wheels are tested on all versions of Windows, MacOS and Ubuntu available as standard GitHub Actions runners. I expect then that the windows-2022 and windows-2025 jobs here will fail.

Afterwards I'll test the effect of trying to build on a different Windows version.

### DIFF
--- a/.github/workflows/buildwheel.yml
+++ b/.github/workflows/buildwheel.yml
@@ -83,7 +83,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-24.04, ubuntu-24.04-arm, windows-2019, macos-13, macos-14]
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
Unfortunately windows-2019 is not usable today:
```
This is a scheduled Windows Server 2019 brownout.
```
I think after today there is a week left in which windows-2019 is still available so will make sure to put out a release in that time.

oscarbenjamin:
https://github.com/actions/runner-images/issues/12045

PR #294, opened 2025-06-26
Try building on windows-2022

oscarbenjamin:
Follows gh-291

### DIFF
--- a/src/flint/types/fmpz.pyx
+++ b/src/flint/types/fmpz.pyx
@@ -5,6 +5,7 @@ from flint.utils.conversion cimport str_from_chars, _str_trunc
 cimport libc.stdlib
 
 from flint.flintlib.types.flint cimport FMPZ_REF, FMPZ_TMP, FMPZ_UNKNOWN, COEFF_IS_MPZ
+from flint.flintlib.functions.flint cimport flint_free
 from flint.flintlib.functions.fmpz cimport *
 from flint.flintlib.types.fmpz cimport fmpz_factor_expand
 from flint.flintlib.functions.fmpz_factor cimport *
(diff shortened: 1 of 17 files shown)


src/flint/types/fmpz.pyx
```diff
@@ -21,7 +21,7 @@ cdef fmpz_get_intlong(fmpz_t x):
     cdef char * s
     if COEFF_IS_MPZ(x[0]):
         s = fmpz_get_str(NULL, 16, x)
-        v = int(str_from_chars(s), 16)
+        v = int(bytes(s).decode('ascii'), 16)
         libc.stdlib.free(s)
         return v
```
oscarbenjamin, line 26: I'm not sure why this change seems to have fixed the problem. The function is defined as:
```
   cdef inline str_from_chars(s):
         return bytes(s).decode('ascii')
```
However actually inlining the function here prevents the crash that is only seen on Windows when building with windows-2022 or windows-2025 but not if built on windows-2019... (The crash happens when running on any Windows version but not if the binaries were built on windows-2019.)

oscarbenjamin: No, I was tricked. This change does not fix the problem. I fell for the classic printing things makes the segfault go away confusion.

oscarbenjamin: Maybe this should use `flint_free`?

oscarbenjamin: All the observed problems were fixed by using flint_free when using `fmpz_get_str`. I'm not sure why this issue appears now and not before.

It is probably worth reviewing all other uses of `free` to see if they should be `flint_free`...

PR #310, opened 2025-08-24
Add shift and truncate methods to nmod_poly

remyoudompheng:
This PR adds a few methods to `nmod_poly` to make its interface closer to `fmpz_mod_poly`. Currently there is no workaround to shift a `nmod_poly` with acceptable performance.


oscarbenjamin:
> they are now added

Thanks

> They are currently not covered at all in `test_all.py` so I chose to keep it that way

Fair enough. Yes, they do need to be tested but I think as you say the awkwardness of the comparisons is probably why they are not.

oscarbenjamin:
The linter cython-lint is complaining:
```console
$ cython-lint src/
/home/runner/work/python-flint/python-flint/src/flint/types/arb_poly.pyx:351:5: E303 too many blank lines (2)
/home/runner/work/python-flint/python-flint/src/flint/types/acb_poly.pyx:353:5: E303 too many blank lines (2)
```

remyoudompheng:
It should be all good now.

oscarbenjamin:
Looks good. Thanks!

PR #312, opened 2025-08-31
Add discriminant method to fmpz_poly, fmpq_poly, nmod_poly

remyoudompheng:
This PR adds a `discriminant` method to all polynomial classes which already have the `resultant` method.
It modifies fmpz_poly, fmpq_poly and nmod_poly (fmpz_mod_poly already has a discriminant method).

Note that for some reason the FLINT C library does not define a `fmpq_poly_discriminant` method.


oscarbenjamin:
Looks good. Could you add a release note for 0.9.0 in the README?

Issue #317, opened 2025-09-03
Test failures on i686


tornaria:
With clean 0.8.0 sources from pypi
See: https://github.com/void-linux/void-packages/actions/runs/17418979707/job/49453428886?pr=56979

To summarize, 1 test failure is:
```
test_fq_default_poly...Traceback (most recent call last):
  File ".../flint/test/__main__.py", line 37, in run_tests
    test()
    ~~~~^^
  File ".../flint/test/test_all.py", line 5030, in test_fq_default_poly
    assert raises(lambda: f.inverse_mod(2*f), ValueError)
           ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
Error in test_fq_default_poly
```

The 8 doctest failures:
```
**********************************************************************
File ".../flint/types/acb.cpython-313-i386-linux-gnu.so", line 529, in flint.types.acb.acb.pow
Failed example:
    acb.integral(lambda z, a: z.pow(acb("1/3"), analytic=a), -5-1j, -5+1j)
Expected:
    [+/- 2.66e-14] + [1.8108516218463 +/- 3.58e-14]j
Got:
    [+/- 2.71e-14] + [1.8108516218463 +/- 3.58e-14]j
**********************************************************************
File ".../flint/types/acb_mat.cpython-313-i386-linux-gnu.so", line 688, in flint.types.acb_mat.acb_mat.eig
Failed example:
    for c in A.eig(algorithm="rump"): print(c)
Expected:
    [1.10529963495745 +/- 4.71e-15] + [+/- 2.92e-15]j
    [-1.91702762744092 +/- 8.45e-15] + [+/- 3.86e-15]j
    [36.8117279924835 +/- 4.72e-14] + [+/- 9.07e-15]j
Got:
    [1.10529963495745 +/- 4.71e-15] + [+/- 2.92e-15]j
    [-1.91702762744092 +/- 8.46e-15] + [+/- 3.87e-15]j
    [36.8117279924835 +/- 4.72e-14] + [+/- 9.11e-15]j
**********************************************************************
File ".../flint/types/acb_mat.cpython-313-i386-linux-gnu.so", line 731, in flint.types.acb_mat.acb_mat.eig
Failed example:
    acb_mat.dft(4).eig(multiple=True)
Expected:
[...]

PR #322, opened 2025-09-06
Extend mul_low and pow_trunc methods to fmpz_poly, fmpq_poly, nmod_poly

remyoudompheng:
These methods already exist on fmpz_mod_poly and fq_default_poly.

Additionally, the existing docstrings and doctests are updated to assume the upper bound of slong is 2^63. Let me know if this is not fine (current all configured Github runners are 64-bit platforms). I am unsure whether 64-bit specific tests are fine.

### DIFF
--- a/src/flint/types/nmod_poly.pyx
+++ b/src/flint/types/nmod_poly.pyx
@@ -724,6 +724,60 @@ cdef class nmod_poly(flint_poly):
             )
         return res
 
+    def mul_low(self, other, slong n):
+        r"""
+        Returns the lowest ``n`` coefficients of the multiplication of ``self`` with ``other``
+
+        Equivalent to computing `f(x) \cdot g(x) \mod x^n`
+
+            >>> f = nmod_poly([2,3,5,7,11], 163)
+            >>> g = nmod_poly([1,2,4,8,16], 163)
+            >>> f.mul_low(g, 5)
+            101*x^4 + 45*x^3 + 19*x^2 + 7*x + 2
+            >>> f.mul_low(g, 3)
+            19*x^2 + 7*x + 2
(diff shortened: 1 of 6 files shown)


oscarbenjamin:
> I am unsure whether 64-bit specific tests are fine.

We want the tests to still pass on 32 bit (gh-317).

> all configured Github runners are 64-bit platforms

Since gh-318 there is the pyodide job which runs on wasm32 although it would still be good to have a proper i686 job to test things like this.

oscarbenjamin:
Looks like arb_poly and acb_poly have mullow but not pow_trunc.

If at least all the exact poly types have these methods now then they can be added to this typing protocol:
https://github.com/flintlib/python-flint/blob/d5c40a4c7ae088caba2ffab5ad79cdd4f41ad505/src/flint/typing.py#L139-L145

src/flint/types/nmod_poly.pyx
```diff
@@ -724,6 +724,60 @@ cdef class nmod_poly(flint_poly):
...
+        nmod_poly_mullow(res.val, self.val, (<nmod_poly>other).val, n)
+        return res
+
+    def pow_trunc(self, slong e, slong n):
+        r"""
+        Returns ``self`` raised to the power ``e`` modulo `x^n`:
+        :math:`f^e \mod x^n`/
+
+        Note: For exponents larger that 2^63 (which do not fit inside a slong) use the
+        method :meth:`~.pow_mod` with the explicit modulus `x^n`.
```
oscarbenjamin, line 761: Maybe the code here could test for overflow and call pow_mod.

oscarbenjamin:
Looks good to me. Thanks

Issue #159, opened 2024-07-04
More efficient conversion from Python int to fmpz


oscarbenjamin:
It looks like Python 3.13 will provide public C API for efficiently converting to and from a `PyLong`. These should be compatible with `mpz_import/mpz_export` rather than converting between hex strings:

https://github.com/python/cpython/pull/121339

https://github.com/flintlib/python-flint/blob/430b8e50f4d8e697d4f5c7666d7b75f3e440a260/src/flint/types/fmpz.pxd#L15-L23
https://github.com/flintlib/python-flint/blob/430b8e50f4d8e697d4f5c7666d7b75f3e440a260/src/flint/types/fmpz.pyx#L16-L27

skirpichev:
> It looks like Python 3.13 will provide public C API

Nit: 3.14.  Maybe.

PR #324, opened 2025-09-07
Faster fmpz/int conversion using from_bytes/to_bytes

remyoudompheng:
This is a proposal to handle issue #159 for Python versions below 3.14 (for example Ubuntu 24.04 LTS uses Python 3.12).
The idea is to use standard int methods (`from_bytes`, `to_bytes`) to obtain a binary serialization, then cast the byte array to ulong.

Special care is needed for big-endian platforms, however I am unable to test whether the patch proposal is correct for big-endian platforms. Let me know what would be the preferred approach.

Benchmarks done using `%timeit` in iPython with Python 3.13
AFAIK gmpy2 uses private Python stuff already so it might be difficult to do better.

| Number | gmpy2 int(mpz) | gmpy2 mpz(int) | int(fmpz) 0.8.0 | fmpz(int) 0.8.0 | int(fmpz) PR | fmpz(int) PR |
| - | - | - | - | - | - | - |
| 3^100 | 45.7 ns | 53.6 ns | 186 ns | 234 ns | 138 ns | 166 ns |
| 3^10000 | 1.49 µs | 1.23 µs | 6.63 µs | 6.52 µs | 1.15 µs | 1.06 µs |
| -3^10000 | 1.50 µs | 1.23 µs | 6.63 µs | 6.60 µs | 1.71 µs | 1.25 µs |


remyoudompheng:
The function `ulong_from_little_endian` is endian-agnostic and was tested by forcing `is_big_endian=1` on a little-endian system.

oscarbenjamin:
The PEP 757 interface is provided as CPython's best effort to give something efficient for this using public API in Python 3.14 onwards. I think that should be the baseline before trying any other approach.

Issue #335, opened 2025-10-30
arb.neg() does nothing


Maegereg:
As currently implemented, `arb.neg()` returns a copy of the original `arb`, rather than its negation. You can see this pretty clearly:

```
from flint import arb

arb(1, 2) in arb(1, 2) # Returns true, as it should.
arb(1, 2).neg() in arb(1, 2) # Returns true, but should return false.
arb(1, 2).neg() in arb(-1, 2) # Returns false, should return true.
arb(1, 2).neg().mid() # Returns 1, should return -1
```

It looks like this is a pretty simple typo: [here](https://github.com/flintlib/python-flint/blob/main/src/flint/types/arb.pyx#L570) `neg()` calls `arb_set()` or `arb_set_round()` rather than `arb_neg()` or `arb_neg_round()`. `__neg__()` is implemented correctly, so if you replace `arb(...).neg()` with `-arb(...)` in the above, it will work.

PR #336, opened 2025-10-30
Fix arb.neg()

Maegereg:
Fixes https://github.com/flintlib/python-flint/issues/335, and adds tests for `arb.neg()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ Contributors (0.9.0):
 - Rémy Oudompheng (RO)
 - Agriya Khetarpal (AK)
 - Oscar Benjamin (OB)
+- Daniel Simmons-Marengo (DSM)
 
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
It looks like you have run a formatter such as black or ruff on the test code. Those formatting changes should be reverted and if we do want to use a formatter then that should be done in a separate PR.

Maegereg:
> Can you add a release note with your name here in the README:

Done!

oscarbenjamin:
Looks good. Thanks!

oscarbenjamin:
For testing purposes now that this PR is merged you can install the nightly wheel with these changes from here:
https://anaconda.org/scientific-python-nightly-wheels/python-flint

If you want to add something in CI to test with these then that is a good option.

PR #340, opened 2025-10-31
Add a much more detailed suite of tests for arb

Maegereg:
Adapted from [Tumult Core's tests](https://github.com/opendp/tumult-core/blob/main/test/unit/utils/test_arb.py) for our own arb wrapper. I've modified these to work with the more bare-bones test setup `python-flint` uses since `pytest` or `unittest` don't seem to be idiomatic.

This does require adding a dev dependency on scipy - let me know if that's a problem. It makes the `erf`, `erfc`, and `erfinv` tests much easier to write, but it's not strictly necessary.

Tumult's arb wrapper supports hashing, so we have tests for it. I am inclined to leave them in and add hashing in a future PR, but let me know if that's not a feature we want.

### DIFF
--- a/src/flint/test/test_arb.py
+++ b/src/flint/test/test_arb.py
@@ -319,7 +319,14 @@ def test_arb_sum():
     true_interval = arb(6, 1.5)  # [4.5, 7.5]
     assert true_interval in actual
 
+def test_no_tests_missing():
+    """Make sure all arb tests are included in all_tests."""
+    test_funcs = {f for name, f in globals().items() if name.startswith("test_")}
+    untested = test_funcs - set(all_tests)
+    assert not untested, f"Untested functions: {untested}"
+
 all_tests = [
+    test_no_tests_missing,
     test_from_int,
     test_from_float,
     test_from_float_inf,


src/flint/test/test_arb.py
```diff
@@ -0,0 +1,354 @@
...
+    """`arb.__sum__` works as expected."""
+    arb1 = arb(1, 0.5)  # [0.5,1.5]
+    arb2 = arb(2, 0.5)  # [1.5,2.5]
+    arb3 = arb(3, 0.5)  # [2.5,3.5]
+    with ctx.workprec(100):
+        actual = arb1 + arb2 + arb3
+    true_interval = arb(6, 1.5)  # [4.5, 7.5]
+    assert true_interval in actual
+
+all_tests = [
```
oscarbenjamin: There should be a test that all the tests in this file are in this list or otherwise someone will end up adding tests here without adding them to the list. There is a test function that does this in `test_all.py` which could be adapted to do check for this file as well.

Maegereg:
> Arb uses __richcmp__ so it does define == but it uses arb_eq which is defined as returning False unless both arbs are exact and exactly equal.

Ah, got it, that all makes sense now. In an ideal world you'd really want to use `arb_equal` for dictionary comparisons rather than `arb_eq`, but we only get one `__eq__` method. Still, I think hashing is situationally useful (we use it for a lru_cache where the arbs will all be exact), so I plan to add it unless someone objects.

src/flint/test/test_arb.py
oscarbenjamin: We could require pytest but I don't think it would particularly make writing tests easier. Either way you mostly just write a lot of functions called `test_foo`. The potential advantages of pytest are things like selecting particular tests or running tests in parallel but the test suite is fast enough not to need those. One thing pytest is not good at is doctests which make up a good fraction of the test suite.

oscarbenjamin:
> Still, I think hashing is situationally useful (we use it for a lru_cache where the arbs will all be exact),

Arbs are immutable so hashing makes sense. It obviously needs to match equality though so as it stands it would only be useful for exact arbs.

PR #341, opened 2025-11-01
Add hash function and re-enable hash tests

Maegereg:
This has function does not match the behavior of `__eq__` - non exact arbs with the same midpoint and radius will hash to the same value even though they're not mathematically equal. I considered having non-exact arbs return random hash values, but that feels wrong (though I do note that the python documentation doesn't explicitly forbid it), and it wouldn't even always work (hash collisions do exist). I also considered hashing the object id for non-exact arbs, but non-exact arbs aren't equal to themselves, so that doesn't quite work correctly either. Overall, this seemed like the simplest solution, even if it does have the possibility to create poor performance if you try to use non-exact arbs in a hashing data structure.

### DIFF
--- a/src/flint/test/test_arb.py
+++ b/src/flint/test/test_arb.py
@@ -126,20 +126,19 @@ def test_hash():
         (arb(10), arb(10), True),
         (arb(10), arb(11), False),
         (arb(10.0), arb(10), True),
-        (
-            arb(mid=10, rad=2),
-            arb(mid=10, rad=2),
-            True,
-        ),
-        (
-            arb(mid=10, rad=2),
-            arb(mid=10, rad=3),
-            False,
-        ),
-        (arb_pi(100), arb_pi(100), True),
-        (arb_pi(100), arb_pi(1000), False),
     ]:
         assert (hash(x) == hash(y)) == expected
+
+    for x in [
+        arb(mid=10, rad=2),
+        arb_pi(100),
+    ]:
+        try:
+            hash(x)
+        except ValueError:
+            pass
+        else:
+            assert False, f"Expected {x} to raise an error if hashed, but succeeded."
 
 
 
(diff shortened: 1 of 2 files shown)


Maegereg:
> Given that arbs don't compare equal to themselves it probably doesn't make sense to use them in a hashing data structure. The behaviour of list and dict is that they check for (truncated) hash equality first, then identity, then equality. This means that with this PR an inexact arb behaves like a nan:

Oh, that is rough. I agree that isn't desirable.

> Ideally Python would have some separate idea of equality like a === b which would be the kind of equality that is used for hash tables like sets and dicts and that uses a different method from a == b.

If I were designing `arb` from scratch, I suspect I'd have `__eq__` be structural equality and have mathematical equality be another method. But at this point that would be quite the breaking change, not to mention a bit counterintuitive.

> What is the actual use case here though? Is it just caching?

For my purposes, yes. We have a method that takes (exact) arbs as parameters, and we want to stick a `@lru_cache` on it for performance reasons. I can speculate that someone might want to hash arbs in other contexts, but I don't have a concrete use case.

I don't think a `.hash_key()` method works for that. Maybe you could do something similar though - I'm imagining a wrapper or subclass that has a hash method defined, and equality defined as structural equality. You could then make it very easy to convert back and forth between the two types. Not perfect, because you would have to convert back and forth, but it gives you reasonable behavior with hashing types.

Maegereg:
Fair point. One other thought: we could keep this hash behavior, but have it raise an error for inexact arbs.

If the three options are:
1. Sensible hash for exact arbs, error for inexact arbs.
2. Sensible hash for exact arbs, nonsensical hash for inexact arbs (current version).
3. Wrapper class that compares and hashes sensibly.

Which do you prefer?

I might lean slightly towards 1, since it's still relatively simple, but doesn't have weird edge cases. But I could be persuaded otherwise.

oscarbenjamin:
Looks good. Thanks.

PR #349, opened 2025-11-10
Add benchmarks for some of the basic operations on some of the basic types

Maegereg:
Covers addition, multiplication, and equality (contains, for arbs) for acb, arb, fmpq, and fmpz.

The primary goal is to use these to measure the performance effect of using the stable API (https://github.com/flintlib/python-flint/issues/338), but they could be useful for other things in the future.

I'm particularly looking for feedback on whether this should include additional types or operations.


Maegereg:
I went ahead wrote up a version that uses pyperf.

PR #367, opened 2026-02-14
Full test and typing coverage

oscarbenjamin:
- mixed-mode arithmetic for arf with float/int/fmpz (not fmpq).
- full type annotations for arf, arb and ctx.
- full test coverage for arf and arb.


oscarbenjamin:
This now gives full type and test coverage for everything except _gr.

Issue #375, opened 2026-03-10
Release that is compatible with FLINT 3.4.0


mkoeppe:

oscarbenjamin:
I opened gh-376 for this but the pyodide build failed.

oscarbenjamin:
I'll keep this open until the release actually happens.

oscarbenjamin:
There is a snag with FLINT 3.4.0 that needs investigating. See gh-383.

PR #376, opened 2026-03-10
Bump FLINT to 3.4.0

oscarbenjamin:
See gh-375


oscarbenjamin:
The CI job here that runs with FLINT's main branch failed with illegal instruction in `test_acb_rmul` which is just
https://github.com/flintlib/python-flint/blob/2da28fe90e27c0f9378f863ff4333ed9e1d13068/src/flint/test/test_acb.py#L136-L140
This is pretty much just calling `acb_mul`:
https://github.com/flintlib/python-flint/blob/2da28fe90e27c0f9378f863ff4333ed9e1d13068/src/flint/types/acb.pyx#L451-L461
That is possibly a bug in FLINT.

PR #388, opened 2026-04-29
check E261 using cython-lint

fchapoton:
about 2 spaces before comments, so very harmless


oscarbenjamin:
I don't see any changes here. Were they undone by the merge?

fchapoton:
It seems that they were done already..

PR #392, opened 2026-05-13
acb_theta_jet

SamSchiavone:
Adds interface to acb_theta_jet to compute derivatives of the Riemann theta function.


src/flint/types/acb_theta.pyx
```diff
@@ -95,3 +95,84 @@ def acb_theta(acb_mat z, acb_mat tau, ulong square=False):
...
+    acb_theta_jet(theta, zvec, nb_in, tau.val, ord, ab, all, square, getprec())
+
+    # Copy the output into a structured format
+    # We return a list of lists: res[char_idx] = [coeff_0, coeff_1, ...]
+    res = []
+    cdef acb r
+    for i in range(nb):
+        char_jet = []
+        for j in range(nj):
+            r = acb.__new__(acb)
```
oscarbenjamin: Rather than creating Python-level acb objects and then using those to create an acb_mat. it would be more efficient to create an acb_mat and use loop over acb_set/acb_mat_entry.

src/flint/test/test_acb_theta.py
```diff
@@ -45,3 +45,22 @@ def test_acb_theta_shape_assertions() -> None:
 
     assert raises(lambda: acb_theta(object(), tau), TypeError)  # type: ignore[arg-type]
     assert raises(lambda: acb_theta(z, object()), TypeError)  # type: ignore[arg-type]
+
+def test_acb_theta_jets_basic() -> None:
+  if not _has_acb_theta():
+      return
```
oscarbenjamin: The indentation here is off.

src/flint/types/acb_theta.pyx
SamSchiavone: Thanks for the tip! I think we fixed this in the latest commit.

oscarbenjamin:
This will need to be added to the docs:
https://github.com/flintlib/python-flint/blob/870524f89f217afb7f24f65989b696af50d06f12/doc/source/acb_theta.rst?plain=1#L4

src/flint/test/test_acb_theta.py
```diff
@@ -45,3 +45,20 @@ def test_acb_theta_shape_assertions() -> None:
 
     assert raises(lambda: acb_theta(object(), tau), TypeError)  # type: ignore[arg-type]
     assert raises(lambda: acb_theta(z, object()), TypeError)  # type: ignore[arg-type]
+
+def test_acb_theta_jets_basic() -> None:
+    if not _has_acb_theta():
+        return
+
+    from flint.types.acb_theta import acb_theta_jets
```
oscarbenjamin: I think that acb_theta and acb_theta_jets were added in different FLINT versions. The guard above in the test is not sufficient. The tests fail with FLINT 3.1 or 3.2:
```
Error importing flint.types.acb_theta: /opt/hostedtoolcache/Python/3.13.13/x64/lib/python3.13/site-packages/flint/types/
acb_theta.cpython-313-x86_64-linux-gnu.so: undefined symbol: acb_theta_jet
```
I guess there can be a `_has_acb_theta_jets` function as well for this.

At some point we can drop support for a lot of the older FLINT versions and then remove a lot of this compatibility code.

oscarbenjamin:
Could you add a release note with your names in the README?

oscarbenjamin:
The tests seem to be failing.

oscarbenjamin:
Looks good. Thanks!

Issue #394, opened 2026-05-22
Upload emscripten wheels to PyPI


oscarbenjamin:
@agriyakhetarpal now that [PEP 783](https://peps.python.org/pep-0783/) for emscripten packaging is accepted does that mean that we can upload the pyodide wheels straight to PyPI?

I think that implementing this would just be a case of removing the `rm` line here and then making a new release:
https://github.com/flintlib/python-flint/blob/d52114534b6896f97fd535ae4ae4afc11a7d4427/.github/workflows/buildwheel.yml#L479-L497

agriyakhetarpal:
Hi @oscarbenjamin, yes, we can! The pyemscripten_2025_0 (2025 ABI) wheels with Pyodide 0.29.4 are all fine to release to PyPI, but I would suggest waiting around a ~week or so to upload wheels for the 2026 ABI, as the Pyodide 314.0.0 stable release is not out yet – we only have alphas for it (see https://github.com/pyodide/pyodide/issues/6233).

agriyakhetarpal:
Pyodide 314.0.0 stable is out: https://blog.pyodide.org/posts/314-release, so you should be unblocked to release now. Note that cibuildwheel is being updated (https://github.com/pypa/cibuildwheel/pull/2906 was merged, but it needs a v4.1.0 release, which I'm coordinating on), but we haven't changed anything ABI-related between 314.0.0a2 and 314.0.0 final as such. So you can release with cibuildwheel right now if you'd like, or wait a bit, or you could use pyodide-build manually outside of cibuildwheel which will grab the 314.0.0 stable xbuildenv by default. Thanks!

agriyakhetarpal:
cibuildwheel v4.1 is out: https://github.com/pypa/cibuildwheel/releases/tag/v4.1.0

oscarbenjamin:
I guess new pyodide means that this should be updated as well:
https://github.com/flintlib/python-flint/blob/870524f89f217afb7f24f65989b696af50d06f12/.github/workflows/buildwheel.yml#L13-L18

oscarbenjamin:
I've opened a PR gh-401

oscarbenjamin:
I've merged gh-401 but a new release would be needed to actually upload the wheels to PyPI.

agriyakhetarpal:
Thank you! I noticed a few things there that are now redundant – let me open a follow-up PR for that.

agriyakhetarpal:
[...]

PR #403, opened 2026-06-15
Pyodide follow-ups

agriyakhetarpal:
xref #394, #401. I bumped to Node.js v24, as that's what we recommend in https://pyodide-build.readthedocs.io/en/latest/tutorials/installation.html#requirements. python-flint doesn't necessarily require it, but v24 is the first to support JSPI.

I have also removed the manual Emscripten toolchain setup, so that we always use the one that cibuildwheel installs via `pyodide xbuildenv install-emscripten` and we have the correct toolchain on PATH. Let's see if that works.

### DIFF
--- a/.github/workflows/buildwheel.yml
+++ b/.github/workflows/buildwheel.yml
@@ -211,7 +211,7 @@ jobs:
 
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: '22'
+          node-version: '24'
 
       - run: pip install pyodide-build
       - run: pyodide xbuildenv install "$PYODIDE_VERSION"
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Ah, right, @oscarbenjamin I remember that the cibuildwheel Emscripten installation happens a little later in the process and not at the time of before-all. I think we are good for you to release with #401 as such, but I should explore if there is a chance we can install Emscripten in cibuildwheel at an earlier stage.

However, there is still one way which we could take because we only build for one Pyodide version/ABI (314.0.0, hence the 2026 ABI), which would be that we move the WASM compilation of the libs to before-build, instead of before-all, as that occurs after Emscripten is installed. But before I do that, I think I should ask whether you plan to support only one ABI at a time? The 2027 ABI will arrive next year, and packaging tools will be updated accordingly, but I know other packages are being built for both the 2025 and 2026 ABIs to push to PyPI.

oscarbenjamin:
> But before I do that, I think I should ask whether you plan to support only one ABI at a time?

Exactly. I don't know what the plan would be in future but I imagine there would be reasons to support at least two pyodide versions simultaneously. Then for the build it is better if before-all builds the libs once to be reused in both wheels if possible.

oscarbenjamin:
> I imagine there would be reasons to support at least two pyodide versions simultaneously.

That being said I would definitely be happy to go with the lowest maintenance option right now and not consider anything else until the need arises in future. Currently we support zero pyodide versions so we can start by supporting one and then see where that lands us in future.

PR #413, opened 2026-07-07
Feature/qqbar direct backend

Madhu18S:


oscarbenjamin:
Something has gone badly wrong with the diff here.

src/flint/types/qqbar.pyx
```diff
@@ -0,0 +1,99 @@
...
+            sys.stdout.flush()
+        finally:
+            os.close(pipe_write)
+            os.dup2(saved_stdout, stdout_fd)
+            os.close(saved_stdout)
+
+        cdef bytes captured = os.read(pipe_read, 4096)
+        os.close(pipe_read)
+
+        return captured.decode('utf-8').strip()
```
oscarbenjamin: This is some weird AI stuff. There is no way that the os module should be involved here.

Madhu18S:
Hi Oscar,
I used the low-level file descriptor approach as an immediate workaround during development because FLINT's qqbar lacks a native string allocation function (like qqbar_get_str). I wanted a quick, working prototype to verify the arithmetic operations without the wrapper crashing on print, specifically because I wasn't sure if I could get it working for midterms.

Now that the logic is solid, I've refactored this method to use standard, cross-platform io.StringIO buffers to intercept the stream output from qqbar_printn. It removes the os module hacking while keeping the implementation lightweight, which is what took me a little bit more time as I was planning to push this along with the next step of comparisons and polynomial extraction. Since you are already reviewing this, I've updated it now.

